PR #4, opened 2022-09-27
Gepruts

DaanMaertens:
added a PeriodicPEPO type to be used in the calculation of expectation values.


maartenvd (approved):
I'll locally remove the .DS_Store

PR #15, opened 2024-03-04
Correct adjoint for truncated SVD

pbrehmer:
This PR implements the correct adjoint for a truncated SVD, as explained in [this](https://arxiv.org/abs/2311.11894) pre-print. To that end, we use KrylovKit's iterative `svdsolve` which is wrapped in the function `itersvd`, so that a custom rrule can defined locally.

I will also add a small test script that compares against the previous adjoint on some gauge-invariant loss function.


pbrehmer:
I think this is ready to merge from my side. The `itersvd` adjoint seems to be compatible with the `TensorKit.tsvd` adjoint for the loss function I tested it on. So this seems to works.

I scrapped the `test_rrule` bit for now since I thought we could add it later once we have unit tests running. Then I'll also add the rrule for `AbstractMatrix`.

pbrehmer:
I updated to the newest KrylovKit version that uses VectorInterface.jl to simplify the coupled linear problem that arises in the SVD adjoint. This allows us to just use a `Tuple`  and to get rid of all the reshapes, so this is much nicer now.

lkdvos:
https://github.com/Jutho/KrylovKit.jl/pull/83
Just wanted to link these PRs

lkdvos:
My take on this:

## Symmetric tensors + `KrylovKit.svdsolve`
In the case of an explicit input tensor, the mapping is straightforward, as this can be achieved simply block-by-block:
```julia
foreach blocksector(t) do c
   Us, Ss, Vs, info = svdsolve(block(t, c), num, which, alg)
   block(U, c)[:, 1:num] .= stack(Us)
    block(S, c)[diagind(Ss)] .= Ss
    block(v, c)[1:num, :] .= stack(Vs)'
end
```
The annoying thing is to determine how you want to specify `num`. As far as I can tell, the easiest way of achieving this is just through a `truncspace(V)`-like specification. This then immediately fixes the structure of `U, S, V`, so filling in the blocks is also straightforward (I am skipping over all details here).

## Symmetric tensors + `KrylovKit.svdsolve` + arbitrary function

In principle, it should also be possible to supply an implicit tensormap (e.g. as the contraction of a network instead of the dense tensormap). This still works, but now we need: 
1. the "row spaces" or "codomain" of the initial guess, or an explicit initial guess
2. the number of svdvalues per sector, which directly translates to inner spaces between U,S, V

The question here is whether or not we want to allow the initial guess to be an arbitrary tensormap, or restrict it to a tensor. In the latter case, it is quite straightforward to then build U and V, as this is just catdomain and catcodomain, but in the former there is an additional permutation involved. I guess this is still possible, but there are definitely some subtleties to consider

## Writing an interface
I guess the easiest is to figure out the implementation of the above first, and then check what input arguments we need. Then we just define a wrapper function in PEPSKit that dispatches to either KrylovKit or TensorKit, depending on if `alg isa SVD/SDD` or `alg isa GKL`.

---

In principle, I would be very happy to have this wrapper in a package extension of TensorKit, as a very similar thing exists for `eigsolve` and `expsolve`, but I would think here is a good place to have a first implementation, which we could consider migrating afterwards

pbrehmer:
I think your take is pretty much spot on. For the case of symmetric tensor + `KrylovKit.svdsolve` that's what I would have implemented. The case of an arbitrary function is not yet super relevant for PEPSKit I think (apparently it doesn't make too much of a performance difference, unless you go to really large dimensions) but of course it should be implemented eventually.

I'm currently writing such a SVD interface with a wrapper function. I think that makes the most sense because that would make user-specified custom SVD functions easy to implement (and that's what I frequently need at the moment). Regarding the input arguments, I would try to keep it quite general. I will push something next week and maybe try to finish the PR then.

lkdvos:
Great! I think the KrylovKit AD PR is getting its final touches, so by then this should be merged and AD might work out of the box. Let me know if I can do something!

pbrehmer:
I added some wrapper structs and a wrapper function for the SVD. I thought it would be a useful thing to have extra structs for the SVD algorithm since you can store the different algorithm parameters and still have a single `svdwrap` interface that will get called inside the CTMRG routine.

I'm not sure yet how to implement the adjoint for the `IterSVD` which uses `KrylovKit.GKL`. The problem with the adjoint inside KrylovKit is that it uses vectors of vectors under the hood instead of matrices to perform all operations. I wonder what the best way is to convert back and forth from `Matrix{ComplexF64}` to `Vector{Vector{ComplexF64}}`.

Also, both the `tsvd!` and `KrylovKit.svdsolve` adjoint lack the possibility to do Lorentzian broadening to avoid divergences in case of degenerate singular values. However, this will be necessary in some PEPS simulations. So here I'm also not sure what the best way is to proceed; either we implement the broadening directly in TensorKit and KrylovKit, or we need a custom adjoint inside PEPSKit.

Note that I also included the `OldSVD` with the previously used (wrong) adjoint. For now it is useful for testing purposes, I think, but we can still throw it out later.

I could use some help or input on how to reuse the `KrylovKit.svdsolve` adjoint and also how to realize the Lorentzian broadening. Of course we could also go for a different interface altogether, I'd happy about suggestions!

src/utility/svd.jl
```diff
@@ -0,0 +1,207 @@
...
+    for (c, b) in blocks(t)
+        x₀ = randn(eltype(b), size(b, 1))
+        Σ, lvecs, rvecs, info = svdsolve(b, x₀, alg.howmany, :LR, alg.alg)
+        if info.converged < alg.howmany  # Fall back to dense SVD if not properly converged
+            U, Σ, V = TensorKit.MatrixAlgebra.svd!(b, TensorKit.SVD())
+            Udata[c] = U
+            Vdata[c] = V
+        else
+            Udata[c] = stack(lvecs)
+            Vdata[c] = stack(rvecs)'
```
lkdvos, line 93: I think this implicitly converts the lazy adjoint to a regular matrix, I'd maybe like to make this a bit more explicit

src/utility/svd.jl
```diff
@@ -0,0 +1,207 @@
...
+    function svdwrap_fullsvd_pullback(Δ)
+        return tsvd!_pullback(Δ)..., NoTangent()
+    end
+    return tsvd_return, svdwrap_fullsvd_pullback
+end
+
+# Wrapper around Krylov Kit's GKL iterative SVD solver
+@kwdef struct IterSVD
+    alg::KrylovKit.GKL = KrylovKit.GKL(; tol=1e-14, krylovdim=25)
+    howmany::Int = 20
```
lkdvos, line 64: I think `howmany` does not carry enough information:
this cannot distinguish between the different blocks, and might give issues if we request too many values for small blocks.
We could use the same approach as `truncspace` however, where this gives a `gradedspace -> (c => howmany_c)` or alternatively just the list of pairs.

src/utility/svd.jl
```diff
@@ -0,0 +1,207 @@
...
+    lorentz_broad::Float64 = 0.0
+    alg_rrule::Union{GMRES,BiCGStab,Arnoldi} = GMRES(; tol=1e-14)
```
lkdvos, line 66: I would keep this `alg_rrule` separate, maybe re-using the `hook_pullback` interface, to decouple the implementation of the forward and backward rules

lkdvos (commented):
I left a couple comments in the code, but in general I quite like the idea.

Some more generic remarks:
I think, if possible, we should decouple the forwards and backwards implementations as much as possible. This means using something like `hook_pullback`, but also that we can move the Lorentzian broadening question into the configuration of the backwards algorithm instead of the forwards. (Am I correct in assuming that the broadening only makes sense when performing AD, and you would otherwise rather have the broadening turned off?) This would also mean that we automatically have the broadening only for the final iteration which is used in the fixed point differentiation, while the bulk of the CTMRG iterations just uses a full decomposition.

I dont like the term `svdwrap`, but I guess that can be changed however we like. I wouldnt mind reusing either `svdsolve` or `tsvd` without importing it, i.e. TensorKit.tsvd would not be the same as PEPSKit.tsvd, which would only matter within this package, and not change it outwards, and then afterwards we could merge the two?

lkdvos:
as a small heads-up, the latest version of tensorkit now is less strict with the `tsvd` types: https://github.com/Jutho/TensorKit.jl/pull/134#event-13366091444

pbrehmer:
Thanks for the heads-up, that is good timing! I will work on the PR tomorrow, so the TensorKit update comes in handy.

pbrehmer:
What I just realized: We can get rid of `eigsolve.jl` since that is now implemented in the new KrylovKit versions. I'll take care of that in this PR. (And of course we should adjust the compat entry for KrylovKit then.)

lkdvos:
It might conflict with the MPSKit compat entries, but I'll try and get a release out for that one asap

src/algorithms/ctmrg.jl
```diff
@@ -1,172 +1,381 @@
...
+            signs[WEST, _prev(r, end), c][-1 1] *
+            envfinal.corners[NORTHWEST, r, c][1; 2] *
+            conj(signs[NORTH, r, c][-2 2])
+    end
+    T1 = map(Iterators.product(axes(envfinal.edges)[2:3]...)) do (r, c)
+        @tensor Tfix[-1 -2 -3; -4] :=
+            signs[NORTH, r, c][-1 1] *
+            envfinal.edges[NORTH, r, c][1 -2 -3; 2] *
+            conj(signs[NORTH, r, _next(c, end)][-4 2])
+    end
```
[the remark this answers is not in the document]
pbrehmer: This is now already implemented in #49

src/utility/svd.jl
```diff
@@ -8,53 +8,106 @@ using TensorKit: SectorDict
...
+end
 
+# IterSVD adjoint for tsvd! using KrylovKit.svdsolve adjoint machinery for each block
 function ChainRulesCore.rrule(
-    ::typeof(TensorKit.tsvd),
+    ::typeof(TensorKit.tsvd!),
     t::AbstractTensorMap;
     trunc::TruncationScheme=notrunc(),
     p::Real=2,
     alg::IterSVD=IterSVD(),
```
lkdvos, line 69: you probably should not have the default arguments in the chainrules, they should be populated in the forwards pass, then dispatch to the full method, for which the chainrules are defined

pbrehmer: Yes good point, but I think this will not be possible using `TensorKit.tsvd` since it doesn't dispatch on the algorithm type (only `_tsvd!` does). I thought I could get away with not using any SVD wrapper but maybe I need to define one after all to be able to dispatch the adjoint on the `alg` type.

src/utility/svd.jl
pbrehmer: But in terms of usability, I think the SVD algorithm struct would be the best way to set the SVD adjoint parameters. So if I were to use `hook_pullback` in this case, I would need to store the `alg_rrule` and `lorentz_broad` settings in the `CTMRG` struct, which I also don't like that much.

src/utility/svd.jl
pbrehmer: I've implemented the `howmany` specification via either `truncspace` or `notrunc`, that should work for now.

src/utility/svd.jl
lkdvos: Ah I was already fearing this. In principle it would still work, we'd have to move the chainrules definitions in TensorKit from `tsvd!` to `_tsvd!`, and then we can still intercept... Maybe the most elegant for now is to just define `PEPSKit.tsvd!` to do what we want from `TensorKit.tsvd!`, and then migrate that upstream once we're happy with the interface?

pbrehmer: That's precisely what I did now, I guess that works for the moment!

pbrehmer:
I've added a compat entry for TensorKit and KrylovKit (which we need for this PR) but currently that's not satisfiable with MPSKit (as you already know I guess). So, once the next MPSKit release is done, we can adjust that. But of course that's not super urgent, so there's no point in stressing it :-)

src/algorithms/ctmrg.jl
```diff
@@ -1,172 +1,381 @@
...
+    maxiter::Int = Defaults.ctmrg_maxiter
+    miniter::Int = Defaults.ctmrg_miniter
+    verbosity::Int = 0
+    svdalg::S = FullSVD()
     trscheme::TruncationScheme = TensorKit.notrunc()
-    tol::Float64 = Defaults.tol
-    maxiter::Integer = Defaults.maxiter
-    miniter::Integer = 4
-    verbose::Integer = 0
     fixedspace::Bool = false
```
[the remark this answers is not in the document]
pbrehmer: Gave it a try and I think I like it. Especially in the future where we might want to have different projector schemes this will be useful.

pbrehmer:
Alright this should be relatively ready to go, I will check if the tests run through next week. I'm happy about any feedback @lkdvos when you have time.

As a last thing regarding the Lorentzian broadening: Since that can only be implemented directly in the adjoints inside of KrylovKit and TensorKit, we should think again about how we want to implement that. Eventually, we really need it for PEPSKit but I would be fine with merging the PR for now and adding the broadening later.

lkdvos:
I just finished the MPSKit release, so at the very least that problem should be fixed now. I'll allocate some time to review Monday (unless I want to procrastinate a bit more, in which case I might get to this sooner)

Project.toml
```diff
@@ -22,13 +22,13 @@ Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
...
-MPSKit = "0.10"
+MPSKit = "0.11"
 OptimKit = "0.3"
 Printf = "1"
 Statistics = "1"
-TensorKit = "0.12"
+TensorKit = "0.12.5"
 VectorInterface = "0.4"
 Zygote = "0.6"
 julia = "1.6"
```
lkdvos, line 34: This probably becomes 1.8 as well, as MPSKit requires this

src/algorithms/ctmrg.jl
```diff
@@ -1,24 +1,46 @@
...
+kind of truncation scheme can be used. If `fixedspace` is true, the truncation scheme is set to
+`truncspace(V)` where `V` is the environment bond space, adjusted to the corresponding
+environment direction/unit cell entry.
+"""
+@kwdef struct ProjectorAlg{S}
+    svd_alg::S = TensorKit.SVD()
+    trscheme::TruncationScheme = TensorKit.notrunc()
+    fixedspace::Bool = false
+    verbosity::Int = 0
+end
```
lkdvos, line 17:
```suggestion
@kwdef struct ProjectorAlg{S,T}
    svd_alg::S = TensorKit.SVD()
    trscheme::T = TensorKit.notrunc()
    verbosity::Int = 0
end
struct FixedSpaceTruncation <: TensorKit.TruncationScheme end
```
How about something like this, where we absorb the "fixedspace" option into the `trscheme`? In any case, I don't think it makes sense to have both `fixedspace` and a non-trivial `trscheme`.
As a side note, I would assume that the default option of `fixedspace=true` makes more sense than `notrunc`, considering the fact that `notrunc` would necessarily run out of memory at some point.

src/algorithms/ctmrg.jl
```diff
@@ -1,24 +1,46 @@
...
+`ProjectorAlg` struct.
 """
 @kwdef struct CTMRG
-    trscheme::TruncationScheme = TensorKit.notrunc()
     tol::Float64 = Defaults.ctmrg_tol
     maxiter::Int = Defaults.ctmrg_maxiter
     miniter::Int = Defaults.ctmrg_miniter
     verbosity::Int = 0
-    fixedspace::Bool = false
+    projector_alg::ProjectorAlg = ProjectorAlg()
```
lkdvos, line 43: It would be cool if we could hide this from an end-user. Ie, get a `CTMRG(; kwargs...)` function that takes in some reasonable kwargs (verbosity, trscheme, svdalg...?), and then automatically constructs the projector algorithm.

src/algorithms/ctmrg.jl
```diff
@@ -1,24 +1,46 @@
...
+    trscheme::TruncationScheme = TensorKit.notrunc()
+    fixedspace::Bool = false
+    verbosity::Int = 0
+end
+
+function fix_spaces(alg::ProjectorAlg)
+    return ProjectorAlg(;
+        svd_alg=alg.svd_alg, trscheme=alg.trscheme, fixedspace=true, verbosity=alg.verbosity
+    )
+end
```
lkdvos, line 23: I think we already have Accessors.jl in our dependencies. This then should be equivalent to 
```julia
alg_fixed = @set alg.fixedspace = true
```

src/algorithms/ctmrg.jl
```diff
@@ -87,7 +110,11 @@ function MPSKit.leading_boundary(envinit, state, alg::CTMRG)
 
     # Do one final iteration that does not change the spaces
     alg_fixed = CTMRG(;
-        alg.trscheme, alg.tol, alg.maxiter, alg.miniter, alg.verbosity, fixedspace=true
+        tol=alg.tol,
+        maxiter=alg.maxiter,
+        miniter=alg.miniter,
+        verbosity=alg.verbosity,
+        projector_alg=fix_spaces(alg.projector_alg),
```
lkdvos, line 117: Similar remark about accessors.jl

src/utility/svd.jl
lkdvos: Ah I see, that's a very good point.

How about a wrapper then?
```julia
struct SVDrrule{F,T,R,E<:Real}
    svd_alg::F
    truncation_alg::T
    rrule_alg::R
    lorentz_broadening::E
end
```

In this case, we still at least semantically separate the rrule from the forward pass (which I really like), while also grouping them together, which I also quite like. It also looks to me to be quite flexible.

I think we should probably investigate (not just here, but throughout PEPSKit), if we want to include the truncation algorithm here or as a separate field. My guess would be that it's actually not a bad idea to keep them separate, considering we presumably might want to have varying `truncspace` settings throughout the unit cell, without changing the SVD algorithm.

src/utility/svd.jl
```diff
@@ -0,0 +1,240 @@
...
+        x₀ = randn(eltype(b), size(b, 1))
+        howmany = trunc isa NoTruncation ? minimum(size(b)) : blockdim(trunc.space, c)
+        S, lvecs, rvecs, info = KrylovKit.svdsolve(b, x₀, howmany, :LR, alg.alg)
+        if info.converged < howmany  # Fall back to dense SVD if not properly converged
+            U, S, V = TensorKit.MatrixAlgebra.svd!(b, TensorKit.SVD())
+            Udata[c] = U
+            Vdata[c] = V
+        else  # Slice in case more values were converged than requested
+            Udata[c] = stack(lvecs[1:howmany])
+            Vdata[c] = stack(rvecs[1:howmany])'
```
lkdvos, line 68:
```suggestion
            Udata[c] = stack(view(lvecs, 1:howmany))
            Vdata[c] = stack(adjoint, view(rvecs, 1:howmany); dims=1)
```
I think using the views removes one allocation. The implementation for Vdata is untested, and might remove another allocation, but you should probably check this.

src/utility/svd.jl
```diff
@@ -0,0 +1,240 @@
...
+    A = storagetype(t)
+    Udata = SectorDict{I,A}()
+    Vdata = SectorDict{I,A}()
+    dims = SectorDict{I,Int}()
+    local Sdata
+    for (c, b) in blocks(t)
+        x₀ = randn(eltype(b), size(b, 1))
+        howmany = trunc isa NoTruncation ? minimum(size(b)) : blockdim(trunc.space, c)
+        S, lvecs, rvecs, info = KrylovKit.svdsolve(b, x₀, howmany, :LR, alg.alg)
+        if info.converged < howmany  # Fall back to dense SVD if not properly converged
```
lkdvos, line 62: We could also consider falling back to dense SVD if `howmany / minimum(size(b)) ~ 1`, or keep this as a separate hybrid algorithm that automatically makes this choice based on some factor. I think I like having that as a separate algorithm though, because now you need to hard-code the fallback algorithm here.

src/utility/svd.jl
```diff
@@ -0,0 +1,240 @@
...
+Wrapper around `TensorKit.tsvd` which dispatches on the `alg` argument.
+This is needed since a custom adjoint for `PEPSKit.tsvd` may be defined,
+depending on the algorithm. E.g., for `IterSVD` the adjoint for a truncated
+SVD from `KrylovKit.svdsolve` is used.
+"""
+function PEPSKit.tsvd(
+    t::AbstractTensorMap, alg; trunc::TruncationScheme=notrunc(), p::Real=2
+)
+    return TensorKit.tsvd(t; alg, trunc, p)
+end
```
lkdvos, line 23: This should probably be `tsvd!`. In TensorKit as well, we abuse the fact that `svd!` is not really an in-place function, rather it is a function that destroys its input by using it as workspace. Because of that reason, we assume that if we call `svd!(A)`, afterwards accessing `A` is undefined behaviour, and no guarantees are made on what the contents of `A` will be. Thus, we hack into this by making `svd!` act like `svd` whenever it is being handled by AD.
Effectively, this means that throughout the CTMRG we don't need to use copies, while only the iteration that is being differentiated gets an additional copy.

src/utility/svd.jl
pbrehmer: The views are definitely removing one allocation . The problem with the `stack(adjoint, ...)` construction is that a singleton dimension is created and you have to get rid of it, which again creates allocations. Consider for example

```julia
julia> v = [[1, 2, 3], [4, 5, 6]]
2-element Vector{Vector{Int64}}:
 [1, 2, 3]
 [4, 5, 6]

julia> @time stack(adjoint, v; dims=2)
  0.000009 seconds (1 allocation: 112 bytes)
1×2×3 Array{Int64, 3}:
[:, :, 1] =
 1  4

[:, :, 2] =
 2  5

[:, :, 3] =
 3  6

julia> @time dropdims(stack(adjoint, v; dims=2); dims=1)
  0.000010 seconds (4 allocations: 240 bytes)
2×3 Matrix{Int64}:
 1  2  3
 4  5  6

julia> @time @view stack(adjoint, v; dims=2)[1, :, :]
  0.000013 seconds (3 allocations: 192 bytes)
2×3 view(::Array{Int64, 3}, 1, :, :) with eltype Int64:
 1  2  3
 4  5  6

julia> @time stack(v)'
  0.000007 seconds (2 allocations: 128 bytes)
2×3 adjoint(::Matrix{Int64}) with eltype Int64:
 1  2  3
 4  5  6
```

I don't know whether there is an allocation-free way to get rid of the singleton dimension, so for now I will just use the outer adjoint `stack(view(rvecs, 1:howmany))'`.

src/utility/svd.jl
pbrehmer: Thanks for the hint, that is an important one. Especially with big bond dimensions this should make difference for the CTMRG performance.

src/utility/svd.jl
lkdvos: `stack(conj, view(rvecs, 1:howmany))` might do the trick

lkdvos: also keep in mind that the last benchmark is not entirely fair, `adjoint(::Matrix{Int64})` is just a lazy adjoint implementation, so in principle you'd have to use `copy(stack(v)')` to instantiate the actual adjoint. This is also what is happening implicitly upon assignment: `Vdata[c] = V'` first creates the lazy wrapper, and then the `setindex!` converts it to the `valtype` of `Vdata`, which needs to instantiate it. 
The cleaner approach here is probably to construct `V'` and then return an `AdjointTensorMap`, but if this does not show up in a profiler we should probably not worry too much about it.
I would be happier to include an explicit `copy` statement though, instead of the implicit conversion during the `setindex!`, just to remind ourselves that this is happening

src/algorithms/ctmrg.jl
pbrehmer: I didn't know about Accessors.jl yet, that's super helpful. In that case, the `fix_spaces` function is not really needed anymore!

src/algorithms/ctmrg.jl
pbrehmer: Yes, totally agree, that's much more useful than having to initialize a `ProjectorAlg` every time. I'll implement such a convenience constructor.

src/utility/svd.jl
pbrehmer: Yes that's a fair point, I ignored the lazy behavior. I think for now `stack(conj, view(rvecs, 1:howmany))` is a good working solution. In that case an explicit `copy` statement is not needed, right? (i.e. there should be no hidden `copy` happening)

src/utility/svd.jl
pbrehmer: That's a great idea! I have now implemented it as an option for `IterSVD`. I think I prefer having it as an option in `IterSVD` rather than having a separate algorithm because we would have quite a bit of code duplication. Having the block-wise dense fallback is anyway a good default thing to do, I would think.

src/utility/svd.jl
```diff
@@ -59,16 +60,24 @@ function TensorKit._compute_svddata!(
...
-            Vdata[c] = stack(conj, view(rvecs, 1:howmany); dims=1)
-            S = @view S[1:howmany]
+            Udata[c] = @view U[:, 1:howmany]
+            Vdata[c] = @view V[1:howmany, :]
+        else
+            S, lvecs, rvecs, info = KrylovKit.svdsolve(b, x₀, howmany, :LR, alg.alg)
+            if info.converged < howmany  # Fall back to dense SVD if not properly converged
+                U, S, V = TensorKit.MatrixAlgebra.svd!(b, TensorKit.SVD())
+                Udata[c] = @view U[:, 1:howmany]
+                Vdata[c] = @view V[1:howmany, :]
```
lkdvos, line 73: Here, the view actually does not make a lot of sense, because the `setindex!` implicitly converts it back to a dense array :wink:

pbrehmer: Ha, you're right, I need to work on my Julia allocation skills :P

src/utility/svd.jl
pbrehmer: I have now added such a wrapper. It leads to the cool thing that you can easily choose the rrule with the corresponding algorithm struct and this decouples the forward SVD algorithm and its reverse rule. I decided to hide the `lorentz_broadening` away inside the `rrule_alg` since in general Lorentzian broadening might not be needed in every SVD implementation.

The only thing I'm unsure about is the naming; maybe something like `SVDWithRrule`, `SVDWithAdjoint`, `DiffSVD` or `DifferentiableSVD` would be more descriptive?

src/algorithms/ctmrg.jl
```diff
@@ -1,172 +1,381 @@
...
+"""
+function gauge_fix(envprev::CTMRGEnv{C,T}, envfinal::CTMRGEnv{C,T}) where {C,T}
+    # Check if spaces in envprev and envfinal are the same
+    same_spaces = map(Iterators.product(axes(envfinal.edges)...)) do (dir, r, c)
+        space(envfinal.edges[dir, r, c]) == space(envprev.edges[dir, r, c]) &&
+            space(envfinal.corners[dir, r, c]) == space(envprev.corners[dir, r, c])
+    end
+    @assert all(same_spaces) "Spaces of envprev and envfinal are not the same"
+
+    # Try the "general" algorithm from https://arxiv.org/abs/2311.11894
```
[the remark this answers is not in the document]
pbrehmer: I think in general that's a good idea but perhaps we should tackle that in a different PR. There are more sophisticated gauge fixing techniques (and also much simpler ones), so it might take some extra time to get that working. I'd suggest we leave it for now and come back later, when we need it ;-)

pbrehmer:
Will add a `IterSVD` test for symmetric tensors tomorrow and try to fix the other tests. Then the PR should be finally done (hopefully).

pbrehmer:
Alright, I think I would be fine with merging the PR like this (once the tests work). Maybe we could first merge #49 and get the tests to run there, and then merge master into this branch (such that the tests still hopefully work).

src/utility/svd.jl
```diff
@@ -0,0 +1,303 @@
...
+    return Udata, Sdata, Vdata, dims
+end
+
+"""
+    struct CompleteSVDAdjoint(; lorentz_broadening = 0.0)
+
+Wrapper around the complete `TensorKit.tsvd!` rrule which requires computing the full SVD.
+"""
+@kwdef struct CompleteSVDAdjoint
+    lorentz_broadening::Float64 = 0.0
```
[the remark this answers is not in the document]
pbrehmer: But that is already contained in the `SVDrrule` wrapper, right? Then it shouldn't be needed inside the adjoint structs.

test/ctmrg/svd_wrapper.jl
```diff
@@ -0,0 +1,88 @@
...
+    rrule_alg=SparseSVDAdjoint(; alg=GMRES(; tol=1e-13)),
+)
+
+@testset "Non-truncacted SVD" begin
+    l_fullsvd, g_fullsvd = withgradient(A -> lossfun(A, full_alg, R), r)
+    l_oldsvd, g_oldsvd = withgradient(A -> lossfun(A, old_alg, R), r)
+    l_itersvd, g_itersvd = withgradient(A -> lossfun(A, iter_alg, R), r)
+
+    @test l_oldsvd ≈ l_itersvd ≈ l_fullsvd
+    @test norm(g_fullsvd[1] - g_oldsvd[1]) / norm(g_fullsvd[1]) < rtol
```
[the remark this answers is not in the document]
pbrehmer: That's indeed much nicer!

src/utility/svd.jl
```diff
@@ -0,0 +1,303 @@
...
+    end
+    return Udata, Sdata, Vdata, dims
+end
+
+"""
+    struct CompleteSVDAdjoint(; lorentz_broadening = 0.0)
+
+Wrapper around the complete `TensorKit.tsvd!` rrule which requires computing the full SVD.
+"""
+@kwdef struct CompleteSVDAdjoint
```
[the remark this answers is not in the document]
pbrehmer: Yes good idea, I also like the contrast to `SparseSVDAdjoint`.

pbrehmer:
@lkdvos I'm not sure how to make the PR compatible with Julia 1.8 since I have to use `Base.get_extension` for the sparse SVD adjoint in order to access functions from the KrylovKit AD package extension. AFAIK package extensions were only added after v1.8 - so do you think there is a way around it somehow, without losing backwards compatibility?

src/utility/svd.jl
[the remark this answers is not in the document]
lkdvos: Aha, I see, I misunderstood, somehow I thought this was the SVDrrule wrapper already. It seems like now, there are quite a lot of these structs, can we converge on something like:
```julia
struct SVDAdjoint{A,B,E}
    alg_fwd::A
    alg_rrule::B
    broadening::E
end
```
where `alg_rrule=nothing` for `alg_fwd= SVD()` or `SDD()`, and otherwise this is just the algorithm?

This probably still requires some dummy struct for dispatching with the "old/wrong" rrule, but otherwise it might be just a little too convoluted

lkdvos:
> @lkdvos I'm not sure how to make the PR compatible with Julia 1.8 since I have to use `Base.get_extension` for the sparse SVD adjoint in order to access functions from the KrylovKit AD package extension. AFAIK package extensions were only added after v1.8 - so do you think there is a way around it somehow, without losing backwards compatibility? Other than that the tests should run!

I think in 1.8, these extensions are just loaded as regular modules, so you can just import them accordingly. For reference, I did a similar thing [here](https://github.com/Jutho/TensorKit.jl/blob/6265c3525e4ce0b27b5471d05d37100f6149c54b/ext/TensorKitChainRulesCoreExt/TensorKitChainRulesCoreExt.jl#L14-L18)

src/utility/svd.jl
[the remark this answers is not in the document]
pbrehmer: Yes I can see that now the struct situation is too convoluted, so I will try to cut down on that by using `SVDAdjoint`. Let me see, if I can clean it up...

pbrehmer:
> I think in 1.8, these extensions are just loaded as regular modules, so you can just import them accordingly

Gave it a try and I'm not so sure, why I can't access the extensions from the main module. I don't quite understand why the TensorOperations example works. I might need some help here

lkdvos:
I'm super confused myself now. I have no problem getting it to work from the REPL on 1.8, and apparently it's also fine from a package extension, but somehow the construction in the main package of a module seems to not work.
I am doing some further digging, but it seems like this is really a limitation of how package extensions and "Requires" work, and in particular from what I am gathering this code is only included after the main module is loaded, so there really is no way to access it.

Technically, that construction is also just very illegal, Extensions should not define new functions, so accessing them from a different package is just going against all guidelines. I have an idea on how to avoid the entire problem, but the simplest solution is to just restrict to julia 1.9 :grin:

pbrehmer:
Oh well, what a mess. Thanks for digging deep :-) I would also be very okay with restricting to v1.9.

pbrehmer:
Alright that would be possible, but it seems like quite a big hassle for supporting a non-LTS version... I'd still go with restricting to v1.9 :)

leburgel:
> So, in principle, we could support 1.8 by computing the rrule block-per-block in the forwards pass using the normal chainrules interface, and then in the pullback applying these rules block per block. I don't think this makes any difference performance-wise, but I am definitely a little lazy right now. @leburgel , do you have any strong feelings against restricting to 1.9?

I have no objections whatsoever to restricting to 1.9, go for it.

lkdvos (approved):
If the tests are green, good to go for me

pbrehmer:
Just noticed a duplicate line, so I removed it. If the tests run, this is also good to go for me!

lkdvos (approved):

Issue #18, opened 2024-03-21
Gauge fixing for nontrivial symmetries and unit cells


leburgel:
I was trying out the new updates on a model with $\mathbb{Z}_2$ symmetry and a $2 \times 2$ unit cell as a test, and I ran into some issues with the gauge fixing.

Unit cells should be fine in principle, but in the current implementation this can give some space mismatches when the edge tensor spaces aren't the same for every site, I think due to this line
https://github.com/quantumghent/PEPSKit.jl/blob/a24ac36321dfbccb0e1e12b3eb9378411f0a4ab5/src/algorithms/ctmrg.jl#L95
not taking into account the site dependence, which is easily fixed.

Symmetries are bit more tricky, since there's two steps where the block structure isn't taken into account,
https://github.com/quantumghent/PEPSKit.jl/blob/a24ac36321dfbccb0e1e12b3eb9378411f0a4ab5/src/algorithms/ctmrg.jl#L82-L83
and
https://github.com/quantumghent/PEPSKit.jl/blob/a24ac36321dfbccb0e1e12b3eb9378411f0a4ab5/src/algorithms/ctmrg.jl#L93-L95
For the first one, projecting out one specific configuration of physical indices isn't very natural in the presence of symmetries. We could just do it, but I'm not sure if this is then still what we want. The second one seems more involved, mainly because I'm not sure what's happening exactly. Just trying the closest block-diagonal analogue to what's there breaks the $\mathbb{Z}_2$ version of the `test_gauge_fixing.jl` test, so I didn't manage to patch it without understanding it.

Probably the case of symmetries just requires the use of a different gauge-fixing procedures, but just wanted to point this out in case there's a quick fix after all. Any thoughts @pbrehmer, @lkdvos?

leburgel:
Ok, after some more messing around I gave up on not thinking and actually implemented the "general" gauge fixing algorithm from [the paper](https://arxiv.org/abs/2311.11894). This works for $\mathbb{Z}_2$ and unit cells, but it breaks the actual AD optimization example since `KrylovKit.eigsolve` is not differentiable (yet). I think Lukas has an rrule for that somewhere, so I'll see if I can get the whole thing to work after I get my hands on that.

pbrehmer:
[...]

PR #19, opened 2024-03-21
Update gauge fixing to work with symmetries

leburgel:
Placeholder PR for an update to the gauge fixing algorithm to one that works for symmetric `TensorMap`s. This implementation fixes the issues #18 , but it breaks the AD since it uses `KrylokvKit.eigsolve` which is not differentiable. Hopefully this works once we add the corresponding rrule.

We should probably also try the [other approaches](https://arxiv.org/abs/2311.11894), but this seemed the simplest one that worked so just adding it here for reference and discussion.

### DIFF
--- a/src/algorithms/ctmrg.jl
+++ b/src/algorithms/ctmrg.jl
@@ -70,7 +70,7 @@ function MPSKit.leading_boundary(state, alg::CTMRG, envinit=CTMRGEnv(state))
 
     env′, = ctmrg_iter(state, env, alg)
     envfix = gauge_fix(env, env′)
-    check_elementwise_convergence(env, envfix; atol=alg.tol) ||
+    check_elementwise_convergence(env, envfix; atol=alg.tol^(3 / 4)) ||
         @warn "CTMRG did not converge elementwise."
     return envfix
 end
@@ -97,13 +97,19 @@ function gauge_fix(envprev::CTMRGEnv{C,T}, envfinal::CTMRGEnv{C,T}) where {C,T}
 
     # Try the "general" algorithm from https://arxiv.org/abs/2311.11894
     signs = map(Iterators.product(axes(envfinal.edges)...)) do (dir, r, c)
-        # Gather edge tensors and pretend they're InfinitMPSs
-        if isodd(dir)
+        # Gather edge tensors and pretend they're InfiniteMPSs
+        if dir == NORTH
             Tsprev = circshift(envprev.edges[dir, r, :], 1 - c)
             Tsfinal = circshift(envfinal.edges[dir, r, :], 1 - c)
-        else
+        elseif dir == EAST
             Tsprev = circshift(envprev.edges[dir, :, c], 1 - r)
             Tsfinal = circshift(envfinal.edges[dir, :, c], 1 - r)
+        elseif dir == SOUTH
+            Tsprev = circshift(reverse(envprev.edges[dir, r, :]), c)
+            Tsfinal = circshift(reverse(envfinal.edges[dir, r, :]), c)
+        elseif dir == WEST
+            Tsprev = circshift(reverse(envprev.edges[dir, :, c]), r)
+            Tsfinal = circshift(reverse(envfinal.edges[dir, :, c]), r)
         end
 
         # Random MPS of same bond dimension
@@ -117,13 +123,16 @@ function gauge_fix(envprev::CTMRGEnv{C,T}, envfinal::CTMRGEnv{C,T}) where {C,T}
             scalartype(T),
             MPSKit._lastspace(Tsfinal[end])' ← MPSKit._lastspace(M[end])',
         )
+
         rhoprev = eigsolve(TransferMatrix(Tsprev, M), rhoinit, 1, :LM)[2][1]
         rhofinal = eigsolve(TransferMatrix(Tsfinal, M), rhoinit, 1, :LM)[2][1]
         # rhoprev = @showgrad rhoprev
+
         # Decompose and multiply
         Qprev, = leftorth(rhoprev, ((1,), (2,)))
         Qfinal, = leftorth(rhofinal, ((1,), (2,)))
         σ = @checkgrad Qprev * Qfinal'
+        # σ = Qprev * Qfinal'
 
         return σ
     end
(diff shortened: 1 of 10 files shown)


pbrehmer:
I can try to code up the truncated eigensolve adjoint similar to the SVD truncated adjoint (also detailed [in here](https://arxiv.org/abs/2311.11894)), which would add a term to the conventional eigensolve adjoint that accounts for the truncated ranks. This could just be added to PR #15. (An SVD would also work instead of an eigendecomposition here, if we decide that implementing the `KrylovKit.eigsolve` adjoint is too much for now.)

lkdvos:
https://github.com/Jutho/KrylovKit.jl/tree/eigsolve_ad/src/adrules

This has been stuck there for quite a long time, but maybe I'll find the time to actually properly implement and test these things such that this can be merged. I am pretty sure that these rules work, and they should work for generic functions so it might be worth a try

leburgel:
I copied the `eigsolve` rrule from Jutho/KrylovKit.jl#56 with some minor tweaks, namely filtering the `ZeroTangent`s here
https://github.com/quantumghent/PEPSKit.jl/blob/f7cf90456269a6085422a5e9d7a7ff2b46b37094/src/utility/eigsolve.jl#L177
and specifying the `add!!` here
https://github.com/quantumghent/PEPSKit.jl/blob/f7cf90456269a6085422a5e9d7a7ff2b46b37094/src/utility/eigsolve.jl#L248

This doesn't work yet, specifically the linear problem rrule sometimes doesn't converge because things inside the anonymous function
https://github.com/quantumghent/PEPSKit.jl/blob/f7cf90456269a6085422a5e9d7a7ff2b46b37094/src/utility/eigsolve.jl#L230-L236
can become `NaN` already for some reason. This is seen from just from running `test_gradients.jl` in the examples folder.

lkdvos:
Some questions I have:
There is a check_elementwise_convergence which seems to check if the gaugefixing didn't do anything, which always triggers for me. Is this expected/necessary?

pbrehmer:
While I haven't made sense of the `@checkgrad` thing, I finally managed to repair the gauge-fixing for arbitrary unit cells. In particular for unit cells larger than $(2, 2)$ the tensor now also converge element-wise.

lkdvos:
I'm still confused about the gradient, I'm quite sure that the @checkgrad implementation I have just discarded the gradient, you can test this by returning the gradient after printing, which breaks it again. The contribution might be small, making the Heisenberg example still converge, but it still feels wrong to just discard it.

pbrehmer:
I fully agree that just discarding the gradient is a bad idea, we should be able to figure out how to compute the adjoint properly. Still, I wasn't able to make any progress on the `eigsolve` adjoint: I can't figure out what the actual origin of the NaNs is. Usually, the first few times the rrule for `eigsolve` is called, the linear problem converges and `Δvecs` is still finite. But then `Δvecs` will only contain NaNs. I'm not sure, however, if this is due to the linear problem not converging or if the linear problem doesn't converge because it gets NaNs as an input (through `b` which depends on `Δvecs`).

Another thought I had is that the adjoint of an eigendecomposition contains possibly diverging terms with $F_{ij} = (\lambda_j - \lambda_i)^{-1}$. This would only be a problem if the eigenvalues are indeed degenerate or very small or very close, right? And I don't think that should the case generally, but maybe there is still something going on with that?

lkdvos:
When I was playing around with it, I also felt like it wasn't actually the rrule of eigsolve that caused the problem, and actually the NaN appeared sooner. But it does feel like sometimes there is just some incredibly small values, which then seem to cause this behaviour

pbrehmer:
Okay, so I think the problem must be somewhere in the differentiation of `leftorth`. I don't know if there is a problem with the rrule or something else going on, but replacing `leftorth` with `tsvd` (and constructing a unitary matrix with $Q=UV^\dagger$) seems to resolve the AD problems of `gauge_fix`. Of course it would be nicer to do a QR decomposition instead of an SVD, but at least this works as a quick fix.

src/algorithms/ctmrg.jl
```diff
@@ -70,29 +70,73 @@ function MPSKit.leading_boundary(state, alg::CTMRG, envinit=CTMRGEnv(state))
...
+    x = gensym(:x)
+    return esc(:(
+        Zygote.hook($expr) do $x
+            if !isfinite(norm($x))
+                @error $x
+            end
+        end
+    ))
+end
+
```
lkdvos, line 88: Can be removed

src/algorithms/ctmrg.jl
```diff
@@ -70,29 +70,73 @@ function MPSKit.leading_boundary(state, alg::CTMRG, envinit=CTMRGEnv(state))
...
-    # Compute gauge tensors by comparing signs
-    # First fix physical indices to (1, 1)
-    Tfixprev = map(x -> convert(Array, x)[:, 1, 1, :], envprev.edges)
-    Tfixfinal = map(x -> convert(Array, x)[:, 1, 1, :], envfinal.edges)
+    # Check if spaces in envprev and envfinal are the same
+    same_spaces = map(Iterators.product(axes(envfinal.edges)...)) do (dir, r, c)
+        space(envfinal.edges[dir, r, c]) == space(envprev.edges[dir, r, c]) &&
+            space(envfinal.corners[dir, r, c]) == space(envprev.corners[dir, r, c])
+    end
+    @assert all(same_spaces) "Spaces of envprev and envfinal are not the same"
```
lkdvos, line 96: I think we can assert that spaces don't change by doing a final step with a truncation using `truncspace`. This is probably a good thing to do anyways.

src/environments/ctmrgenv.jl
```diff
@@ -93,3 +93,66 @@ function LinearAlgebra.dot(e₁::CTMRGEnv, e₂::CTMRGEnv)
 end
 
 LinearAlgebra.norm(e::CTMRGEnv) = norm(e.corners) + norm(e.edges)
+
```
lkdvos, line 96: Should probably implement these in terms of VectorInterface functions

src/utility/rotations.jl
```diff
@@ -10,22 +10,3 @@ const SOUTHWEST = 4
 
 # Rotate tensor to any direction by successive application of Base.rotl90
 rotate_north(t, dir) = mod1(dir, 4) == NORTH ? t : rotate_north(rotl90(t), dir - 1)
```
lkdvos, line 12: Maybe also define rotl180 and rotr90, to avoid double and triple reallocations

lkdvos:
Just put these things here to not forget them, will try and address them myself tomorrow if I find some time

lkdvos:
I cleaned up the VectorInterface part, and did some minor tweaks. For me this is ready to go.

lkdvos (approved):

PR #23, opened 2024-05-01
Update optimization to use rrule system

lkdvos:
This is a major refactor of the optimization code to have a uniform interface.

In particular, I implemented all of the `ctmrg_gradient` functions through AD by using @pbrehmer 's implementation of the gradient and transforming them into rrules via @leburgel 's adaptation.

Then, in order to be able to dynamically select which rrule to use, I added the function `hook_pullback`, which functions as a convenient interface to change the rrule algorithm through an optional keyword that can later be dispatched on in the rrule.

Finally, I added some rrules for constructors/getters for InfinitePEPS and CTMRGEnv, which prevents our costfunction from yielding weird tangent types. This liquidates lots of the checks to see if the tangents are the correct type and gives a bit of a cleaner implementation. I have not played around with how robust this is, and if custom costfunctions would easily break this, but I did add a type check in the implementation of the rrules to at least detect when this is happening.


src/algorithms/peps_opt.jl
```diff
@@ -45,12 +45,24 @@ function fixedpoint(
...
+        E, g = withgradient(peps) do ψ
+            envs´ = hook_pullback(
+                leading_boundary,
+                ψ,
+                alg.boundary_alg,
+                envs;
+                alg_rrule=alg.gradient_alg,
+            )
+            ignore_derivatives() do
+                alg.reuse_env && (envs.edges .= envs´.edges; envs.corners .= envs´.corners)
```
lkdvos, line 59:
```suggestion
                alg.reuse_env && update!(envs, envs´)
```
@leburgel , can you refactor this into a separate function?

pbrehmer (approved):
The refactor looks really great, overall it's a big improvement. Thanks!

The gradient test was broken since it wasn't yet adapted to the `hook_pullback` wrapper, so I quickly repaired it. Also, the `ManualIter` fixed-point differentiation had a weird convergence criterion. I changed that to make it comparable to `GeometricSum` (these two approaches are kind of equivalent anyways).

pbrehmer:
Yes that's a good idea, I would think it is easier to merge that way.

lkdvos:
I'll rebase tomorrow and get this ready to merge

pbrehmer (approved):

PR #24, opened 2024-05-14
Add tests to CI

lkdvos:
This moves some of the tests that were in the examples folder to a proper testing framework


lkdvos:
This seems like its still not very stable, and we should probably do something about the overly verbose printing... I might be able to find some more time this afternoon

pbrehmer:
I will also take a look in a bit, maybe I manage to stabilize the tests.

pbrehmer:
I like how MPSKit prints out the first and last VUMPS iterations when you set `verbosity=1`. I think this would be useful to have also in PEPSKit and it would give some useful printing during the tests. Let me quickly add that, while I'm at it!

lkdvos:
I definitely agree, sadly the machinery for that is still in the untagged master of mpskit, so its not so easy to just import it :p

pbrehmer (dismissed):

pbrehmer:
Maybe, once it's tagged, we can use that machinery to make the output printing and logging a bit more nice :)

pbrehmer:
Well, the tests ran through on v1.9.3, let's see if I can also get them to run on v1.6.

pbrehmer:
@lkdvos Do you think we could get rid of the Julia 1 tests? I'm not sure if it makes sense to be backwards compatible to v1.0 if PEPSKit anyway still is in a very experimental stage. Also it seems to be difficult to get the tests running consistently on both v1.6 and v1.0. Maybe we could instead test on the latest stable release.

lkdvos:
I think the Julia 1 tests expand to the latest stable Julia, so 1.10 in this case...

pbrehmer:
It's really annoying that some of these runs are so sensitive (at low bond/environment dimension) to the random initial state. I guess it takes a bit of trying around with the random seed to find one that works for both versions. I can try around tomorrow if you don't have the time/motivation now (which would be perfectly understandable :D)

lkdvos:
So at this point, I'm basically ready to give up...
Probably it is better to just look for a reasonable physical state somewhere, such as some state which is at the very least approximately close to a groundstate?
I'm fine with merging like this, and maybe we can add an ED thing on a 3x3 lattice later and then just decompose that state and use the center peps tensor?

pbrehmer:
Thanks for the effort, it really is annoying! Having an ED initial PEPS also sounds like a good idea. I will give it a final go tomorrow but then I'd also be fine with merging.

lkdvos:
Small FYI, I tried playing around with the Ising PEPS, also does not seem to work (doesnt even want to converge with CTMRG). I think that one might also just be a bit pathological anyways, or I might have implemented it wrong, but I am kind of running out of ideas. I would think the best approach is to just save a reasonable state on disk somewhere, and use that

pbrehmer:
> I would think the best approach is to just save a reasonable state on disk somewhere, and use that

Yes, I agree that this is the most sensible approach in the long run. Maybe we can code up a small state/environment saving and loading interface soon. I guess that's useful to have anyways.

pbrehmer (approved):

pbrehmer:
Hope that this runs for now... I think this should be ready to merge!

PR #28, opened 2024-05-16
Add docstrings and example for boundary MPS functionality

leburgel:
Also filtered some internals from the library docs.


test/boundarymps/vumps.jl
lkdvos, line 1: maybe we could wrap these in some `@testset` environments, just to increase the readability?
Otherwise looks good to me!

leburgel: Something like this you mean? (I didn't want to merge this first two in a loop since I want to explicitly test the non-multiline syntax as well.)

lkdvos (approved):

PR #38, opened 2024-05-30
Actually fix the thing I said I fixed before...

lkdvos:
:(


leburgel (approved):
I also didn't notice, was not paying close attention. I guess it's pretty hard to add a test for this kind of edge case?

lkdvos:
The only thing I could come up with is to start from something symmetric, and not impose the symmetry, but even then, I would guess that the PEPS will just break the symmetry anyways, which leads to a super contrived setup to test this stupid warning message... maybe its fine just like this, Gleb tested it so now at least it runs...

leburgel:
I did explicitly try out the line which gets the singular value dict this time around, so it should be good to go now.

PR #46, opened 2024-06-19
Rewrite local expectation values as a single contraction

leburgel:
Rewrites the local expectation values as a single contraction and switches on the optimal contraction ordering in `@tensor`. This should improve performance, since I'm a bit skeptical the approach of first contracting the full environment and then multiplying in the operator afterward is the most efficient. This currently bypasses the `operator_env` function entirely, although I left it in just in case it is still needed elsewhere.

In particular, this should fix the `OnSite` expectation value, which was wrong before I think (it seems it assumed a transposed environment?). Also, I chose to transpose the `NearestNeighbor` environment density matrix to be consistent with the usual interpretation.

### DIFF
--- a/src/operators/localoperator.jl
+++ b/src/operators/localoperator.jl
@@ -77,12 +77,18 @@ Evaluate the expectation value of any `NLocalOperator` on each unit-cell entry
 of `peps` and `env`.
 """ MPSKit.expectation_value(::InfinitePEPS, ::Any, ::NLocalOperator)
 
+# Optimal contraction order is obtained by manually trying out some space sizes and using costcheck = warn
+# in principle, we would like to use opt = true, but this does not give optimal results without also supplying costs
+# However, due to a bug in tensoroperations this is currently not possible with integer labels.
+# Thus, this is a workaround until the bug is fixed. (otherwise we'd need to rewrite all the labels to symbols...)
+
 # 1-site operator expectation values on unit cell
 function MPSKit.expectation_value(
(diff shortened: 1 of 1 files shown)


lkdvos:
Nice! I'll try and add some more tests to cover the onsite case later today, so let's wait just a second before merging

src/operators/localoperator.jl
```diff
@@ -77,21 +77,140 @@ Evaluate the expectation value of any `NLocalOperator` on each unit-cell entry
...
+            10,
+            14,
+            19,
+            15,
+            13,
+            31,
+            32,
+            29,
+            30,
+        ) begin # physical spaces
```
lkdvos, line 153: It would be nice if this got formatted better, but I am not sure how to achieve this.

lkdvos (approved):
Looks like a good addition to me. I added the "optimal" contraction order, by taking an order which seems to be optimal at reasonable values of the physical, bond, and boundary dimensions, but there are a couple TensorOperations fixes that need to happen before I can properly make this automatically compute an optimal order.

In general, good to go for me!

src/operators/localoperator.jl
leburgel: I tried switching off the formatting just for that method. Looks a lot better, but very manual and not robust to changes so not sure if it's a great idea. I can revert if that's better.

lkdvos: I definitely prefer it like this :)

lkdvos (approved):

PR #47, opened 2024-06-20
Add support for fermionic systems

qmortier:
fixes for fermionic CTMRG
tests for (parts of) AD energy gradients
anisotropic Hamiltonians
p-wave superconductor example
ugly but effective fix for fermionic CTMRG gauge fix


src/algorithms/ctmrg.jl
```diff
@@ -96,11 +91,21 @@ function MPSKit.leading_boundary(envinit, state, alg::CTMRG)
     )
     env′, = ctmrg_iter(state, env, alg_fixed)
     envfix = gauge_fix(env, env′)
-    check_elementwise_convergence(env, envfix; atol=alg.tol^(3 / 4)) ||
+    check_elementwise_convergence(env, envfix; atol=alg.tol^(1 / 2)) ||
```
lkdvos, line 94: do we really need this to be 1 / 2? is this an artifact from testing?

src/algorithms/ctmrg.jl
```diff
@@ -96,11 +91,21 @@ function MPSKit.leading_boundary(envinit, state, alg::CTMRG)
...
 
+#very ugly fix to avoid Zygoting @plansor during fermionic gauge fix
+@generated function MPSKit.transfer_right(v::AbstractTensorMap{S,1,N₁}, A::GenericMPSTensor{S,N₂},
+    Ā::GenericMPSTensor{S,N₂}) where {S,N₁,N₂}
+        t_out = MPSKit.tensorexpr(:v, -1, -(2:(N₁ + 1)))
+        t_top = MPSKit.tensorexpr(:A, (-1, reverse(3:(N₂ + 1))...), 1)
+        t_bot = MPSKit.tensorexpr(:Ā, (-(N₁ + 1), reverse(3:(N₂ + 1))...), 2)
+        t_in = MPSKit.tensorexpr(:v, 1, (-(2:N₁)..., 2))
+        return :(return @tensor $t_out := $t_top * conj($t_bot) * $t_in)
+end
```
lkdvos: This could be made less ugly by explicitly checking that the braiding is symmetric, otherwise this definition also breaks anyonic mps code. Alternatively, (and way nicer) is to implement the reverse rule for this operation (which should be `transfer_left`)

src/operators/localoperator.jl
```diff
@@ -24,6 +24,18 @@ struct NLocalOperator{I<:AbstractInteraction}
...
+    struct NLocalOperator{I<:AbstractInteraction}
+    
+Operator in form of a `AbstractTensorMap` which is parametrized by an interaction type.
+Mostly, this is used to define Hamiltonian terms and observables.
+"""
+struct AnisotropicNNOperator
+    h0::NLocalOperator{OnSite}
+    hx::NLocalOperator{NearestNeighbor}
+    hy::NLocalOperator{NearestNeighbor}
+end
```
lkdvos, line 37: this implementation is type-unstable, and the docstring is incorrect

src/utility/symmetrization.jl
```diff
@@ -61,7 +61,7 @@ end
 
 # rotation invariance
 
-Base.rotr90(t::PEPSTensor) = permute(t, ((1,), (5, 2, 3, 4)))
+#Base.rotr90(t::PEPSTensor) = permute(t, ((1,), (5, 2, 3, 4)))
 Base.rot180(t::PEPSTensor) = permute(t, ((1,), (4, 5, 2, 3)))
```
lkdvos, line 65: Can we move all of these definitions together? `src/states/abstractpeps.jl` also has some of these

test/ctmrg/gradients.jl
```diff
@@ -2,12 +2,92 @@ using Test
...
+                        complex.(x[(ctr + 1):(ctr + n)], x[(ctr + n + 1):(ctr + 2n)]) ./
+                        sqrt(dim(c)))
+            end
+            ctr += T <: Real ? n : 2n
+        end
+        return t′
+    end
+
+    return vec, from_vec
+end
```
lkdvos: Have to recheck this with the current code from TensorKit, as this is copied from an old version. Ideally we move this to a package extension for ChainRulesTestUtils in TensorKit itself

test/ctmrg/gradparts.jl
```diff
@@ -0,0 +1,147 @@
...
+using Random
+using PEPSKit
+using TensorKit
+using PEPSKit: NORTH, SOUTH, WEST, EAST, NORTHWEST, NORTHEAST, SOUTHEAST, SOUTHWEST, rotate_north, left_move, ctmrg_iter
+using Zygote
+using ChainRulesCore
+using ChainRulesTestUtils
+using FiniteDifferences
+
+## Test utility
```
lkdvos, line 22: Please refactor to separate file, so it can be included from multiple places instead of being copied everywhere...

test/heisenberg.jl
```diff
@@ -24,14 +24,13 @@ end
...
+ctm_alg = CTMRG(; trscheme=truncdim(χenv), tol=1e-10, miniter=4, maxiter=100, verbosity=1)
 opt_alg = PEPSOptimize(;
     boundary_alg=ctm_alg,
-    optimizer=LBFGS(4; maxiter=100, gradtol=1e-3, verbosity),
-    gradient_alg=GMRES(; tol=1e-6, maxiter=100),
+    optimizer=LBFGS(4; maxiter=100, gradtol=1e-3, verbosity=2),
+    gradient_alg= GMRES(; tol=1e-6, maxiter=100),
     reuse_env=true,
-    verbosity,
+    verbosity=2,
```
lkdvos: I think all of these changes can be safely reverted?

test/runtests.jl
```diff
@@ -19,6 +20,7 @@ const GROUP = get(ENV, "GROUP", "All")
     if GROUP == "All" || GROUP == "EXAMPLES"
         @time @safetestset "Heisenberg model" begin
             include("heisenberg.jl")
+            include("pwave.jl")
```
lkdvos, line 23: Separate testset, in the same style of everything above

src/algorithms/ctmrg.jl
```diff
@@ -409,8 +418,10 @@ function build_projectors(
     U::AbstractTensorMap{E,3,1}, S, V::AbstractTensorMap{E,1,3}, Q_sw, Q_nw
 ) where {E<:ElementarySpace}
     isqS = sdiag_inv_sqrt(S)
-    @tensor Pl[-1 -2 -3; -4] := Q_nw[-1 -2 -3; 1 2 3] * conj(V[4; 1 2 3]) * isqS[4; -4]
-    @tensor Pr[-1; -2 -3 -4] := isqS[-1; 1] * conj(U[2 3 4; 1]) * Q_sw[2 3 4; -2 -3 -4]
+    Pl = Q_nw * V' * isqS
+    Pr = isqS * U' * Q_sw
+    #@tensor Pl[-1 -2 -3; -4] := Q_nw[-1 -2 -3; 1 2 3] * conj(V[4; 1 2 3]) * isqS[4; -4]
+    #@tensor Pr[-1; -2 -3 -4] := isqS[-1; 1] * conj(U[2 3 4; 1]) * Q_sw[2 3 4; -2 -3 -4]
```
lkdvos, line 424: remove comments

test/heisenberg.jl
```diff
@@ -24,14 +24,13 @@ end
...
-ctm_alg = CTMRG(; trscheme=truncdim(χenv), tol=1e-10, miniter=4, maxiter=100, verbosity)
+ctm_alg = CTMRG(; trscheme=truncdim(χenv), tol=1e-10, miniter=4, maxiter=100, verbosity=1)
 opt_alg = PEPSOptimize(;
     boundary_alg=ctm_alg,
-    optimizer=LBFGS(4; maxiter=100, gradtol=1e-3, verbosity),
+    optimizer=LBFGS(4; maxiter=100, gradtol=1e-3, verbosity=2),
     gradient_alg=GMRES(; tol=1e-6, maxiter=100),
     reuse_env=true,
-    verbosity,
+    verbosity=2,
```
lkdvos, line 33: these changes can be reverted I think?

lkdvos (changes requested):
Some clean-up left, mostly left the comments to remember to do them. Will look into it when I have some time

test/pwave.jl
```diff
@@ -0,0 +1,58 @@
...
+H = square_lattice_pwave()
+χbond = 2
+χenv = 16
+ctm_alg = CTMRG(;
+    trscheme=truncdim(χenv), tol=1e-10, miniter=4, maxiter=200, fixedspace=true, verbosity=1
+)
+opt_alg = PEPSOptimize(;
+    boundary_alg=ctm_alg,
+    optimizer=LBFGS(4; maxiter=100, gradtol=1e-3, verbosity=2),
+    gradient_alg=ManualIter(; tol),
```
leburgel, line 42: The `tol` here is not defined, causing the test to fail. Also, is there a particular reason to use `ManualIter` as the fixed-point gradient algorithm? I think `GMRES` is expected to be the 'best' approach in general, so maybe we should test using this instead?

lkdvos: I think GMRES is definitely ok

src/algorithms/ctmrg.jl
qmortier: the gaugefix tests use 1e-10 for the ctm and 1e-4 for the elementwise convergence so fairly close to 1/2. I never encountered problems with 1/2 but don't know if higher powers would also work

src/algorithms/ctmrg.jl
qmortier: In general this is also not correct for fermions and one would need P's in there when doing something MPS optimization or whatever oriented and using @tensor. However, here you don't really need them as you can absorb everything in fixed-points and random bottom layer MPS. 

How would you go about the transfer_left rrule option?

test/pwave.jl
qmortier: for the pwave GMRES takes significantly longer than the other options, so I choose another one

test/heisenberg.jl
qmortier: mostly did this to have the CTM on verbosity 1 and LBFGS on 2 which gives the most informative output imo, dont know if you want this in a test tho

src/algorithms/ctmrg.jl
lkdvos: No, I mean this implementation breaks the non-ad pass of MPSKit. If you now have both MPSKit and PEPSKit loaded, you will no longer be able to do Fibonacci anyons.

Also, I think I was too quick with thinking it is just the other transfer, as you also need the derivatives to the transfer matrix

qmortier: yes I know, but also even for fermions the forward pass would be broken now. However this broken fermionic is still ok here due to the specific nature of what we're trying to do here (find these phases). In short: I fully agree that it is a very messy fix :p

test/pwave.jl
lkdvos: Do we have any idea why? In principle the manual iteration should really be doing the exact same as GMRES, but GMRES has some additional steps that should make it such that it has to perform less function evaluations. Thus, I would guess that it's just some settings that have to be tweaked. Maybe the `eager=true`, or just lower the krylov subspace dimension? In any case, I would rather solve this problem than just special case this to `ManualIter` if possible :)

src/algorithms/ctmrg.jl
lkdvos: Ok, in any case, I changed the implementation to avoid using MPSKit, as I don't think fixing one thing while breaking another is really a fix :p

src/algorithms/ctmrg.jl
```diff
@@ -145,12 +140,33 @@ function gauge_fix(envprev::CTMRGEnv{C,T}, envfinal::CTMRGEnv{C,T}) where {C,T}
             scalartype(T),
             MPSKit._lastspace(Tsfinal[end])' ← MPSKit._lastspace(M[end])',
         )
-        ρprev = eigsolve(TransferMatrix(Tsprev, M), ρinit, 1, :LM)[2][1]
-        ρfinal = eigsolve(TransferMatrix(Tsfinal, M), ρinit, 1, :LM)[2][1]
+
+        # this is a bit of a hack to get the fixed point of the mixed transfer matrix
+        # because MPSKit is not compatible with AD
+        ρ_prev = let Tsprev = Tsprev, M = M
```
leburgel, line 146: Probably better to split the contents of the `let` block off into a separate function and then call it twice instead of having basically the same block twice?

src/operators/localoperator.jl
```diff
@@ -178,3 +190,17 @@ function costfun(peps::InfinitePEPS, env, op::NLocalOperator{NearestNeighbor})
...
 end
+
+"""
+    costfun(peps::InfinitePEPS, env, op::AnisotropicNNOperator)
+    
+Compute the expectation value of an on-site and an anisotropic nearest-neighbor operator.
+This is used to evaluate and differentiate the energy in ground-state PEPS optimizations.
+"""
+function costfun(peps::InfinitePEPS, env, op::AnisotropicNNOperator)
+    oos = expectation_value(peps, env, op.h0)[1]
```
leburgel, line 201: Why is this one not a `sum`? Seems like it should be.

lkdvos: I agree

lkdvos (approved):

PR #48, opened 2024-07-02
Implement more consistent unit cell index convention

pbrehmer:
This PR will implement a more consistent (or at least human-understandable) index convention for labeling PEPS and environment tensors in a unit cell. To that end, we will follow the convention of [Bruognolo et al.](https://scipost.org/submissions/2006.08289v3/):

image

However, I won't use `y` and `x` to label the rows and columns. Instead I find it more helpful to use `r` and `c`, respectively, since the indexing convention is that of a matrix and not that of Cartesian coordinates. Especially when dereferencing arrays, e.g. `env.corners[NORTHWEST, y, x]`, feels annoying to me.

To test for correctness, I will add a small PEPS optimization example for the Heisenberg Hamiltonian with a 2x2 unit cell ansatz. Since I'm anyway relabeling many indices, I will replace the NCON numbers with symbols that are actually descriptive, e.g. `Dnorth`, `chiwest`, and so on.

### DIFF
--- a/src/algorithms/ctmrg.jl
+++ b/src/algorithms/ctmrg.jl
@@ -379,7 +379,7 @@ function left_move(state, env::CTMRGEnv{C,T}, alg::CTMRG) where {C,T}
     return CTMRGEnv(corners, edges), copy(Pleft), copy(Pright), ϵ
 end
 
-# Compute enlarged NW corner
+# Compute enlarged corners
 function northwest_corner(E4, C1, E1, peps_above, peps_below=peps_above)
     @tensor corner[-1 -2 -3; -4 -5 -6] :=
         E4[-1 1 2; 3] *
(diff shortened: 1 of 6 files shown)


lkdvos:
As a small suggestion, it's probably best to test this also by just computing a random expectation value of a 3x3 unit cell with varying spaces on each of the bonds. The 3x3 is necessary to be able to distinguish between +- 1 mistakes, while the different spaces ensure that the contraction will just error if wrong tensors are matched up.

pbrehmer:
Thanks for the suggestion. I anyway wanted to code up the bond matching test at some point but was always a bit too lazy. Now would be a good occasion to finally implement that.

pbrehmer:
I will wait for PR #49 to be merged into master before continuing because I would have to change some parts anyway with the new local operator system.

pbrehmer:
Alright, in that case I will just push the rest into your branch. Should be done tomorrow!

pbrehmer:
All the things are now implemented in #49, so I will close this PR :)

lkdvos:
@pbrehmer Can I delete this branch?

pbrehmer:
Yes sorry, I'll do it!

PR #49, opened 2024-07-02
PEPSHamiltonian Refactor

lkdvos:
This PR refactors the operator construction. It consists of a couple parts that I think will make it nice to work with generic operators:

## Generic contractions
I bit the bullet and just wrote some generic contraction routines for arbitrary interactions. These are captured in the functions `contract_localoperator` and `contract_localnorm`, which take `N` coordinates and a `AbstractTensorMap{S,N,N}`, and generates code for contracting them. It's using TensorOperations' functionality to optimize the contraction order, and in principle could work for arbitrary ranges, but probably at some point either the contraction order optimizer will become prohibitively expensive, or the contraction will.
Highlights here are the fact that this means we can also do J1J2 models, or even plaquette operators.

## PEPSHamiltonian
Because of the way the generic contractions work, all we really need to define a hamiltonian term is a set of indices and an operator. Thus, I wrote a simple wrapper `PEPSHamiltonian`, which just represents a tuple of these terms. I think it will be convenient in the future to have a field `lattice`, which is just an array of the physical spaces upon which it acts. This struct will silently assume to act as the identity on all sites that are not present, and as such it can be useful to know what spaces these are (for example if we would like to initialize a peps for such an operator)

There's definitely a lot of things that can still be added, from convenience methods such as `+(PEPSHamiltonian, PEPSHamiltonian)` etc, to convenience constructors such as `next_nearest_neighbour`, and I am missing a bit of documentation and docstrings still, but I already procrastinated a bit too much so I was hoping to give this as an initial launch.
@pbrehmer , it might be nice to couple this with the new re-ordering of the coordinates for the environments. I currently wrote this with the old convention, just to make everything work and because I wasn't entirely sure what we settled on, but it should be relatively straightforward to change these offsets. Do you want to look into this?


pbrehmer:
Thanks for the rewrite, it looks quite amazing at first glance!

> I currently wrote this with the old convention, just to make everything work and because I wasn't entirely sure what we settled on, but it should be relatively straightforward to change these offsets.

I can definitely adjust all the indices to the new convention, that shouldn't be too much work. I'll take a look tomorrow.

src/operators/models.jl
```diff
@@ -11,11 +11,12 @@ function square_lattice_heisenberg(
     ::Type{T}=ComplexF64; Jx=-1, Jy=1, Jz=-1
 ) where {T<:Number}
     physical_space = ComplexSpace(2)
+    lattice = fill(physical_space, 1, 1)
     σx = TensorMap(T[0 1; 1 0], physical_space, physical_space)
     σy = TensorMap(T[0 im; -im 0], physical_space, physical_space)
     σz = TensorMap(T[1 0; 0 -1], physical_space, physical_space)
     H = (Jx * σx ⊗ σx) + (Jy * σy ⊗ σy) + (Jz * σz ⊗ σz)
-    return NLocalOperator{NearestNeighbor}(H / 4)
+    return nearest_neighbour_hamiltonian(lattice, H)
```
leburgel, line 19:
```suggestion
    return nearest_neighbour_hamiltonian(lattice, H / 4)
```
Accidentally dropped the division, so the Heisenberg test is failing.

lkdvos: I'm super confused, I noticed this this morning, pushed a fix, which shows up [here](https://github.com/QuantumKitHub/PEPSKit.jl/tree/ld-expval), but not in the PR

lkdvos: ok, I forced the matter by changing the base branch a bit, seems ok now

src/operators/localoperator.jl
```diff
@@ -1,219 +1,500 @@
...
+    lattice::Matrix{S}
+    terms::T
+end
+function PEPSHamiltonian(lattice::Matrix{S}, terms::Pair...) where {S}
+    lattice′ = PeriodicArray(lattice)
+    for (inds, operator) in terms
+        @assert operator isa AbstractTensorMap
+        @assert numout(operator) == numin(operator) == length(inds)
+        for i in 1:length(inds)
+            @assert space(operator, i) == lattice′[i]
```
leburgel, line 438:
```suggestion
            @assert space(operator, i) == lattice′[inds[i]]
```

src/algorithms/ctmrg.jl
```diff
@@ -140,18 +140,14 @@ function gauge_fix(envprev::CTMRGEnv{C,T}, envfinal::CTMRGEnv{C,T}) where {C,T}
...
+        ρfinal = transfermatrix_fixedpoint(Tsfinal, M, ρinit)
 
         # Decompose and multiply
-        Up, _, Vp = tsvd!(ρ_prev)
-        Uf, _, Vf = tsvd!(ρ_final)
-        Qprev = Up * Vp
-        Qfinal = Uf * Vf
-        σ = Qprev * Qfinal'
+        Qprev, = leftorth(ρprev)
+        Qfinal, = leftorth(ρfinal)
```
lkdvos, line 148: we should probably restrict the compat for TensorKit to be above the version where the rrules for this got fixed, I think this is at least 0.12.4

pbrehmer: Definitely a good idea, just added that

pbrehmer:
I have added a small transverse field Ising test, but don't have reference data for the energy and magnetization yet (will try to get something). And I was thinking about two more things:

- Should we maybe add a J1J2 example or test? I'm not quite sure yet if the local operator system automatically does an efficient construction for the next-neighbor interactions (this L-shaped contraction between next-neighbor PEPS).
- I don't know if `PEPSHamiltonian` is a good name for that struct since it should also be used for more general operators, e.g. the (staggered) magnetization. Maybe we should go for something like `PEPSObservable` (also not perfect since the operator doesn't have to be an observable) or `PEPSOperator` (which is maybe too close to PEPO)?

lkdvos:
I'd be very happy with the J1J2 example as well. In principle, the contraction should be generic, so at least it should work. I'm not entirely sure though about the efficiency. I didn't check what people usually do, the contraction routine will just take the rectangular area around the active sites and contract the entire thing, which obviously becomes expensive whenever the diagonal distance increases.

I have no strong opinions on the name, I chose PEPSHamiltonian in analogy to the MPOHamiltonian of MPSKit, but I do agree with your arguments. Pepsoperator is probably to close to PEPO, but maybe even LocalOperator might be fine? I don't mind abusing the term Hamiltonian to denote "a sum of (quasi) local terms" either so anything goes really 😅

pbrehmer:
I think overall I'd prefer `LocalOperator`, so I can change that in the next commit :-)

Most of the contraction indices should have better labeling now (symbols instead of integers) but there are a few files where I wasn't sure about the labeling. Could someone (@leburgel @lkdvos) maybe take a look at some point? The files I'm thinking of are `derivatives.jl`, `transferpepo.jl` and `transferpeps.jl`. It's not super important I guess so we can also leave it for now.

Another thing I wasn't sure about: Is there a smart way to go about optimizing the contraction order? The way I would do it now is to use the `opt = (chi => chi^2, D => chi, ...)` argument. But that's quite annoying for big contractions since the list of indices gets really long and I basically provide the same information many times. @lkdvos Do you think it's somehow possible to generically fix the index cost in all tensor contractions which have indices that start with `chi` or with `D` to a certain value?

lkdvos:
I think this should do the trick :)

pbrehmer:
So I now added `@autoopt` to all CTMRG contractions but I now get weird errors, where I'm not sure why and where the macros causes that. For example, if I try to run the `heisenberg.jl` example, the `leading_boundary` call will lead to the following:

```julia
ERROR: UndefVarError: `edge_W` not defined
Stacktrace:
 [1] southwest_corner(edge_S::TrivialTensorMap{…}, corner_SW::TrivialTensorMap{…}, edge_W::TrivialTensorMap{…}, peps_above::TrivialTensorMap{…}, peps_below::TrivialTensorMap{…})
   @ PEPSKit ~/repos/PEPSKit.jl/src/algorithms/ctmrg.jl:408
 [2] southwest_corner
   @ ~/repos/PEPSKit.jl/src/algorithms/ctmrg.jl:408 [inlined]
 [3] left_move(state::InfinitePEPS{TrivialTensorMap{…}}, env::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}}, alg::CTMRG)
   @ PEPSKit ~/repos/PEPSKit.jl/src/algorithms/ctmrg.jl:320
 [4] ctmrg_iter(state::InfinitePEPS{…}, env::CTMRGEnv{…}, alg::CTMRG)
   @ PEPSKit ~/repos/PEPSKit.jl/src/algorithms/ctmrg.jl:290
 [5] leading_boundary(envinit::CTMRGEnv{…}, state::InfinitePEPS{…}, alg::CTMRG)
   @ PEPSKit ~/repos/PEPSKit.jl/src/algorithms/ctmrg.jl:41
 [6] top-level scope
   @ ~/repos/PEPSKit.jl/test/heisenberg.jl:24
Some type information was truncated. Use `show(err)` to see complete types.
```

@lkdvos Do you have an idea what might be happening?

lkdvos:
Macros are hard...

lkdvos:
What's the status on this PR right now? Are there still things we need to do, or is it just a matter of stabilizing the tests a bit?

pbrehmer:
Yes, I think when the tests run through, this should be mergeable (at least from my side).

Regarding the tests: on the new version of MPSKit (on the other PR), the VUMPS "(2, 2) PEPS" was spitting out a lot of imaginary warnings and the results came out a bit too wrong (had to increase the tolerance quite a bit). So maybe that needs some stabilizing.

lkdvos:
I'll fix that tomorrow. The new version of Krylovkit allows us to use lanczos for hermitian systems because it warns instead of errors, so the default there got switched to lanczos. PEPS transfer matrices are just typically not hermitian, hence the problems. It's just a matter of switching back to Arnoldi

lkdvos:
Ok, I'm hoping this should now be fixed, so once the tests turn green this is good to go for me.
The warnings with the mpskit loggers is something I'll fix on the MPSKit side of things, so this should be automatically inherited after I tag a new release there

PR #51, opened 2024-07-09
Logging utilities

lkdvos:


leburgel:
Looks good to me, but we should probably increase the verbosity in the examples and tests to at least `verbosity=2`, since level 1 got less verbose by default. None of the tests actually output log anything CTMRG-related here as far as I can tell. Also, it would be good to document what the different verbosity levels actually do (probably in the `CTMRG` docstring, which is now out of date with the possible levels).

pbrehmer:
I agree, that probably the default verbosity should be changed and that a quick explanation of levels 0 to 3 would be useful to have in the `CTMRG` docstring. Other than that, I find the new logging machinery really neat (and also makes it more elegant to log on a HPC environment). Thanks!

src/algorithms/peps_opt.jl
```diff
@@ -49,7 +51,7 @@ Different levels of output verbosity can be activated using `verbosity` (0, 1 or
     )
     reuse_env::Bool = true  # Reuse environment of previous optimization as initial guess for next
     gradient_alg::G = GeomSum() # Algorithm to solve gradient linear problem
-    verbosity::Int = 0
+    verbosity::Int = 1 # currently unused?
```
pbrehmer, line 54: Yes, this is currently unused and I think that by now it actually is redundant. All of the printing information happens either in the `optimizer` or the `gradient_alg` - both of which have their own verbosity. So how about we get rid of it?

pbrehmer (commented):
Except for the verbosity in `PEPSOptimize` I find this really nice and useful. Thanks!

pbrehmer (approved):

PR #52, opened 2024-07-09
Fix unit cell indexing in `CTMRGEnv` constructor

leburgel:
The `CTMRGEnv(peps::InfinitePEPS{P}; Venv=oneunit(spacetype(P))) where {P}` constructor was not adapted to the new unit cell convention. Noticed this right before #49 got merged, but was too late to stop the automerge.

I added this as a separate PR since there anyway might be some more things related to unit cells we want to add. One thing that comes to mind is to extend this constructor to support non-uniform virtual environment spaces. Also, we should probably add a more extensive test that puts random duals as well as random dimensions on all virtual spaces (PEPS and environments) and checks that everything matches up everywhere.

Alternatively, I could quickly tag the fix onto #15 and then add the more general constructors and tests later, if that's better @lkdvos @pbrehmer ?

### DIFF
--- a/src/environments/ctmrgenv.jl
+++ b/src/environments/ctmrgenv.jl
@@ -22,21 +22,30 @@ function CTMRGEnv(peps::InfinitePEPS{P}; Venv=oneunit(spacetype(P))) where {P}
     corners = Array{C_type}(undef, 4, size(peps)...)
     edges = Array{T_type}(undef, 4, size(peps)...)
 
-    for dir in 1:4, i in 1:size(peps, 1), j in 1:size(peps, 2)
-        @diffset corners[dir, i, j] = TensorMap(randn, scalartype(P), Venv, Venv)
-        @diffset edges[dir, i, j] = TensorMap(
-            randn,
-            scalartype(P),
-            Venv * space(peps[i, j], dir + 1)' * space(peps[i, j], dir + 1),
-            Venv,
+    for (r, c) in Iterators.product(axes(peps)...)
+        for dir in 1:4
+            corners[dir, r, c] = TensorMap(randn, scalartype(P), Venv, Venv)
+        end
+        edges[NORTH, _prev(r, end), c] = TensorMap(
+            randn, scalartype(P), Venv * space(peps[r, c], 2)' * space(peps[r, c], 2), Venv
+        )
+        edges[EAST, r, _next(c, end)] = TensorMap(
+            randn, scalartype(P), Venv * space(peps[r, c], 3)' * space(peps[r, c], 3), Venv
+        )
+        edges[SOUTH, _next(r, end), c] = TensorMap(
+            randn, scalartype(P), Venv * space(peps[r, c], 4)' * space(peps[r, c], 4), Venv
+        )
+        edges[WEST, r, _prev(c, end)] = TensorMap(
+            randn, scalartype(P), Venv * space(peps[r, c], 5)' * space(peps[r, c], 5), Venv
         )
     end
 
-    @diffset corners[:, :, :] ./= norm.(corners[:, :, :])
-    @diffset edges[:, :, :] ./= norm.(edges[:, :, :])
+    corners[:, :, :] ./= norm.(corners[:, :, :])
+    edges[:, :, :] ./= norm.(edges[:, :, :])
 
     return CTMRGEnv(corners, edges)
 end
+@non_differentiable CTMRGEnv(peps::InfinitePEPS)
 
 # Custom adjoint for CTMRGEnv constructor, needed for fixed-point differentiation
 function ChainRulesCore.rrule(::Type{CTMRGEnv}, corners, edges)


pbrehmer:
Good that you noticed, thanks! I'd say that we could still tag it onto #15 since it is an important fix. I totally agree that we need more unit cell tests and constructors, but maybe we can add them later since that takes a bit of time. (Time that I probably don't have at the moment ;-))

lkdvos:
I would actually argue the opposite. In principle, the tests run, and the released version is not affected by this problem. So only the master is currently in a slightly broken state, which is not that big of a problem as no-one should use that version anyways.

It's nice to have changes separated by commits, large PRs with multiple different changes are very hard to debug afterwards, and pinpointing where something went wrong is a bit of a nightmare when all we are left with is the big squashed commit...

So maybe, let's ignore the issue for PR #15 , and just merge this afterwards. It would then be nice to have this PR include:
- [ ] a CTMRGEnv constructor with variable spaces
- [ ] a 3x3 test that uses this and covers the error discovered here

I would suggest to re-use the same structure as the `InfinitePEPS` constructor:
```julia
CTMRG([f, T], peps::InfinitePEPS, chi_vertical::S, chi_horizontal::S=chi_vertical; unitcell::Tuple{Int,Int}=(1,1)) where S<:Union{Int,ElementarySpace}
CTMRG([f, T], D_horizontal::S, D_vertical::S, chi_vertical::S, chi_horizontal::S=chi_vertical; unitcell::Tuple{Int,Int}=(1,1)) where S<:Union{Int,ElementarySpace}
CTMRG([f, T], peps::InfinitePEPS, chis_vertical::A, chis_horizontal::A=chis_vertical) where {A<:AbstractMatrix{<:Union{Int,ElementarySpace}}}
CTMRG([f, T], Ds_horizontal::A, Ds_vertical::A, chis_vertical::A, chis_horizontal::A=chis_vertical) where {A<:AbstractMatrix{<:Union{Int,ElementarySpace}}}
```
In an ideal world, it would be cool if we could also allow the peps bond dimensions to be different for the bra and the ket, which enables things like VOMPS -- approximating a peps by a different one with a different bonddimension, or some time-evolution algorithms as well.

src/environments/ctmrgenv.jl
```diff
@@ -8,44 +8,272 @@ struct CTMRGEnv{C,T}
     edges::Array{T,3}
 end
 
+_spacetype(::Int) = ComplexSpace
+_spacetype(::S) where {S<:ElementarySpace} = S
+
+_to_space(χ::Int) = ℂ^χ
+_to_space(χ::ElementarySpace) = χ
```
lkdvos, line 15: I like this approach, is there any chance you can also do something like this for the `InfinitePEPS` constructors?

leburgel: I don't think this would help with the `InfinitePEPS` constructors themselves, since these never have to make a distinction but just delegate to the `PEPSTensor` constructor. For the latter, I think it's a bit tricky to squash the two types of constructors since it's impossible to specify a dual space when passing only integer dimensions.

Specifically, it seems to me this
https://github.com/QuantumKitHub/PEPSKit.jl/blob/5cd0304a0475ff880d7ceb0f534f1fc90bf9b842/src/states/abstractpeps.jl#L39
choice of automatically dualizing the south and west spaces makes it hard to merge the two cases .

lkdvos: Ok, I didn't really look into it in detail, I just had a vague memory of a similar pattern.

docs/src/index.md
```diff
@@ -37,8 +37,13 @@ opt_alg = PEPSOptimize(;
 
 # ground state search
 state = InfinitePEPS(2, D)
+<<<<<<< HEAD
+ctm = leading_boundary(CTMRGEnv(state, ComplexSpace(chi)), state, ctm_alg)
+result = fixedpoint(state, Heisenberg_hamiltonian, opt_alg, ctm)
+=======
 ctm = leading_boundary(CTMRGEnv(state; Venv=ComplexSpace(chi)), state, ctm_alg)
 result = fixedpoint(state, H, opt_alg, ctm)
+>>>>>>> master
```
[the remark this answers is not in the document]
leburgel: Good catch, fixed now.

src/algorithms/peps_opt.jl
```diff
@@ -60,7 +60,7 @@ in `alg`. The initial environment `env₀` serves as an initial guess for the fi
 By default, a random initial environment is used.
 """
 function fixedpoint(
-    ψ₀::InfinitePEPS{T}, H, alg::PEPSOptimize, env₀::CTMRGEnv=CTMRGEnv(ψ₀; Venv=field(T)^20)
+    ψ₀::InfinitePEPS{T}, H, alg::PEPSOptimize, env₀::CTMRGEnv=CTMRGEnv(ψ₀, field(T)^20)
```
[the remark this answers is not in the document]
leburgel: Yes, I only updated the constructor syntax but also thought this was not particularly robust as a default. Using a `oneunit` would be more generic, but that leads to all kinds of other trouble if fixed space truncation is the default.

A default virtual space in itself does not really make sense to me anyway. If the 'barrier' is in users constructing an explicit `CTMRGEnv` beforehand, we could add a separate method that takes a virtual space (or even integer) as the last argument, which first constructs the environment using that and then calls this one.

leburgel:
Okay, I think this ticks all the boxes now. Feels a bit clunky, but I don't immediately see how to make things more compact. So good to go from my end unless there are more comments.

src/environments/ctmrgenv.jl
```diff
@@ -2,42 +2,333 @@
     struct CTMRGEnv{C,T}
 
 Corner transfer-matrix environment containing unit-cell arrays of corner and edge tensors.
+The last two indices of the arrays correspond to the row and column indices of the unit
+cell, whereas the first index corresponds to the direction of the corner or edge tensor. The
+directions are labeled in clockwise direction, starting from the north-west corner and north
+edge respectively.
+
+Given arrays of corners `c` and edges `t`, they connect to the PEPS tensors at site `(r, c)`
+isn the unit cell as:
```
lkdvos, line 11:
```suggestion
in the unit cell as:
```

lkdvos (dismissed):
I quickly glanced over this, looks good to me. Maybe @pbrehmer want's to have another look, but otherwise we can merge this.

pbrehmer:
Looks lovely, I think it can be merged as it is!

PR #53, opened 2024-07-09
Implement simultaneous CTMRG scheme and CTMRG gradients without gauge-fix differentiation

pbrehmer:
Here I will implement the CTMRG scheme where all projector are computed simultaneously and applied from both sides to the corners at the same time. In combination with a gauge-fixed SVD, it makes it possible to differentiate a CTMRG iteration without also differentiating the `gauge_fix` function itself. This speeds up and stabilizes the fixed-point gradient computation.

Things still to do:
- [x] add a flag for selecting the way how the CTMRG iteration is differentiated (either with or without differentiating `gauge_fix`)
- [x] implement new index convention & use `@autoopt`
- [x] implement a good way of passing a unit cell of U, S and Vs (probably with some `SVDAdjoint` algorithm)
- [x] add multi-threading (since all enlarged corners and projectors can be computed simultaneously now)
- [x] add new tests
- [x] add safeguards for illegal setting configurations


pbrehmer:
This is almost there, but I still have to fix a few things:
- the `gaugefix.jl` test is quite unstable with the new CTMRG scheme, and I don't know why yet
- if one uses the `FixedSVD` wrapper where the fixed U, S and V were computed from a `IterSVD`, this doesn't lead to element-wise convergence, which I find very confusing; this needs to be fixed
- it's not very clear how to use the KrylovKit adjoint when the forward algorithm hasn't been `GKL`; this is the case if we use `FixedSVD` where then a dummy forward algorithm needs to be created

Other than that, most of the new features work well I think. The naming of the new options is not yet perfect, so I'd be happy about alternative suggestions. Maybe you can take a look @lkdvos when you have the time? :-)

src/algorithms/ctmrg.jl
```diff
@@ -117,200 +126,27 @@ function MPSKit.leading_boundary(envinit, state, alg::CTMRG)
                 @warn(
                     "CTMRG reached maximal number of iterations at (Δnorm=$Δnorm, ΔCS=$ΔCS, ΔTS=$ΔTS)"
                 )
-            return conv_condition, normnew, CSnew, TSnew, ϵ
+            flush(stdout)  # Flush output to enable live printing on HPC
+            flush(stderr)  # Same for @info, @warn, ...
```
lkdvos, line 130: I think this is something we can handle with a logger as well. I don't have a lot of experience with working with HPC, but I would guess that by implementing a non-default logger and using that one, we don't have to pollute our functions with this.

src/utility/svd.jl
```diff
@@ -131,7 +149,9 @@ function ChainRulesCore.rrule(
             n_vals = length(Sdc)
             lvecs = Vector{Vector{scalartype(t)}}(eachcol(Uc))
             rvecs = Vector{Vector{scalartype(t)}}(eachcol(Vc'))
-            minimal_info = KrylovKit.ConvergenceInfo(length(Sdc), nothing, nothing, -1, -1)  # Just supply converged to SVD pullback
+            minimal_info = KrylovKit.ConvergenceInfo(length(Sdc), nothing, nothing, -1, -1)  # Only num. converged is used
+            minimal_alg = GKL(; tol=1e-6)  # Only tolerance is used for gauge sensitivity
```
lkdvos, line 153: is this algorithm not part of the `SVDAdjoint` struct?

src/utility/util.jl
```diff
@@ -197,3 +157,26 @@ macro showtypeofgrad(x)
...
+    diffable_ex = quote
+        if Zygote.isderiving()
+            $ex
+        else
+            Threads.@threads $ex
+        end
+    end
+
+    return esc(diffable_ex)
+end
```
lkdvos, line 182: I like this idea, however I think we could do better by doing some more manual work. In particular, our use case involves a parallel map, where each element is independently computed and returned. Thus, the derivative is also just a parallel map, i.e. we could just compute the derivatives element-wise in parallel as well.

Nevertheless, that might require a bit more work, so I think it is fine to keep this for now and change that later. I would like to go to something like [OhMyThreads.jl](https://github.com/JuliaFolds2/OhMyThreads.jl), if necessary with some extra `rrule`, so I'll already link that here.

test/ctmrg/gaugefix.jl
```diff
@@ -43,17 +45,17 @@ end
     psi = _make_symmetric(psi)
 
     Random.seed!(987654321)  # Seed RNG to make random environment consistent
+    # psi = InfinitePEPS(physical_space, peps_space; unitcell)
     ctm = CTMRGEnv(psi; Venv=ctm_space)
 
     verbosity = 1
     alg = CTMRG(;
-        tol=1e-10, miniter=4, maxiter=400, verbosity, trscheme=truncdim(dim(ctm_space))
+        tol=1e-10, miniter=4, maxiter=5000, verbosity, trscheme=FixedSpaceTruncation()
```
lkdvos, line 53: This sounds like an excessive maxiter, do we need that?

lkdvos (commented):
I like many things, and I like some things a tiny bit less but this is probably because I don't understand them :)
I left some comments in the code with my thoughts, and mostly wanted to add that we should probably now merge #52 first, and possibly #51 as well?
Otherwise, as this is quite a big change, it might be nice to have another zoom and discuss this in a bit more detail next week. I am not entirely following and probably if you walk me through it's going to be a lot easier

pbrehmer:
> I left some comments in the code with my thoughts, and mostly wanted to add that we should probably now merge #52 first, and possibly #51 as well?

Yes, it's a good to idea to merge them first. I will take a look at them as well!

> it might be nice to have another zoom and discuss this in a bit more detail next week

Let's definitely have another zoom. I also realized that there are quite a few changes which might not be obvious from reading the code, so it would be useful to explain it a bit more :)

test/ctmrg/gaugefix.jl
pbrehmer: Yes definitely, that was just me trying around to stabilize the test. I will put it back to reasonable settings, once I stabilized it.

src/algorithms/ctmrg.jl
pbrehmer: A specialized HPC logger would be a great solution, since I didn't want to pollute the main functions with it but it is quite necessary if you want to work in a HPC environment (which I currently do).

leburgel: I usually handle this by using a [`FormatLogger`](https://julialogging.github.io/reference/loggingextras/#LoggingExtras.FormatLogger) or a [`FileLogger`](https://julialogging.github.io/reference/loggingextras/#LoggingExtras.FileLogger-Tuple{Any}) from LoggingExtras.jl as the global logger, both of which have an `always_flush` option for live printing to `stderr`. This does not help with `stdout`, but since we probably anyway want to output all the verbosity as log messages and not plain print statements this might be enough.

pbrehmer: Thanks for the tip! I actually think a `FormatLogger` would be enough since with #51 all the `stdout` messages I previously needed are now also pushed into `stderr` via `@infov`. So that should solve the problem.

src/utility/util.jl
pbrehmer: I really like the idea of having a differentiable parallel map; that was actually what I hoped to find somewhere but haven't found yet. Writing the simple macro was just a quick solution that already speeds up CTMRG significantly for large unit cells or larger bond dimensions. So I'm also fine with keeping it like that for now, but I would also love to have a differentiable parallel map at some point.

src/algorithms/ctmrg_all_sides.jl
```diff
@@ -0,0 +1,177 @@
...
+        rnext = _next(r, size(state, 1))
+        cprev = _prev(c, size(state, 2))
+        cnext = _next(c, size(state, 2))
+        Q[dir, r, c] = if dir == NORTHWEST
+            northwest_corner(
+                env.edges[WEST, r, cprev],
+                env.corners[NORTHWEST, rprev, cprev],
+                env.edges[NORTH, rprev, c],
+                state[r, c],
+            )
```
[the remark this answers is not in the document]
pbrehmer: That's a good idea, I have implemented it like that now and a saves a few lines

src/algorithms/ctmrg_all_sides.jl
```diff
@@ -0,0 +1,177 @@
...
+    rc_combinations = collect(Iterators.product(axes(state)...))
+    @fwdthreads for (r, c) in rc_combinations
+        rprev = _prev(r, size(state, 1))
+        rnext = _next(r, size(state, 1))
+        cprev = _prev(c, size(state, 2))
+        cnext = _next(c, size(state, 2))
+        @diffset @autoopt @tensor corners[NORTHWEST, r, c][χ_S; χ_E] :=
+            P_right[WEST, rnext, c][χ_S; χ1 D1 D2] *
+            Q[NORTHWEST, r, c][χ1 D1 D2; χ2 D3 D4] *
+            P_left[NORTH, r, c][χ2 D3 D4; χ_E]
```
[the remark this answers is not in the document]
pbrehmer: Wow this actually took me way too long since `@diffset` behaved really weirdly. Apparently
```julia
C_NW = _contract_new_corner(
            P_right[WEST, rnext, c], Q[NORTHWEST, r, c], P_left[NORTH, r, c]
)
@diffset corners[NORTHWEST, r, c] = C_NW
```
produces correct results but assigning directly to the destination
```julia
@diffset corners[NORTHWEST, r, c] = _contract_new_corner(
            P_right[WEST, rnext, c], Q[NORTHWEST, r, c], P_left[NORTH, r, c]
)
```
somehow produces wrong results but also doesn't error. What a headache... At some point we should maybe really try to scrap `@diffset`

src/utility/svd.jl
pbrehmer: In the case that `SVDAdjoint.fwd_alg isa IterSVD` yes, but now also algorithms with `SVDAdjoint.fwd_alg isa FixedSVD` have to use that rrule. However, they only carry U, S and V. That means in that case you don't have access to the original algorithm that produced the U, S and V (also not necessarily `GKL`).

src/algorithms/ctmrg.jl
```diff
@@ -28,16 +28,25 @@ end
 """
     CTMRG(; tol=Defaults.ctmrg_tol, maxiter=Defaults.ctmrg_maxiter,
           miniter=Defaults.ctmrg_miniter, verbosity=0,
-          svd_alg=TensorKit.SVD(), trscheme=FixedSpaceTruncation())
+          svd_alg=TensorKit.SVD(), trscheme=FixedSpaceTruncation(),
+          ctmrgscheme=:AllSides)
```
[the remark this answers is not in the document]
pbrehmer: In general the new scheme leads to much faster differentiation since you don't have to differentiate the gauge fixing anymore. In that way I think it's the better default.

On the other hand, it seems to have problems with the Z2 gauge fixing tests with unit cells, so I don't know whether there is a more general problem at hand or not.

src/algorithms/ctmrg_all_sides.jl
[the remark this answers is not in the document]
pbrehmer: Upon further inspection, I found that this has to do with the parallelization together with `@diffset`. I'm not sure what is happening but somehow `@diffset` together with `Threads.@threads` must produces some sort of race condition. (Without `@diffset` there definitely shouldn't be a race condition.)

src/algorithms/ctmrg.jl
```diff
@@ -523,3 +359,44 @@ function LinearAlgebra.norm(peps::InfinitePEPS, env::CTMRGEnv)
...
+        @autoopt @tensor transferv[χ_TL D_Tab D_Tbe χ_TL; χ_BL D_Bab D_Bbe χ_BR] :=
+            env.edges[EAST, r, _next(c, end)][χ_TR D1 D2; χ_BR] *
+            peps[r, c][d; D_Tab D1 D_Bab D3] *
+            conj(peps[r, c][d; D_Tbe D2 D_Bbe D4]) *
+            env.edges[WEST, r, _prev(c, end)][χ_BL D3 D4; χ_TL]
+
+        function lintransfer(v, t)
+            @tensor v′[-1 -2 -3 -4] := t[-1 -2 -3 -4; 1 2 3 4] * v[1 2 3 4]
+            return v′
+        end
```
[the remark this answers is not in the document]
pbrehmer: Thanks for the pointer. I'm not even sure if it make sense to compute the correlation length like that for unit cells. I probably need to something similar to `gauge_fix`, where we consider an entire row or column of edges. Then I will just reuse that machinery

src/algorithms/ctmrg.jl
```diff
@@ -28,16 +28,25 @@ end
...
+depending on `verbosity` (0, 1 or 2).
+
+The projectors are computed from `svd_alg` SVDs where the truncation scheme is set via
+`trscheme`.
+
+In general, two different schemes can be selected with `ctmrgscheme` which determine how
+CTMRG is implemented. It can either be `:LeftMoves`, where the projectors are succesively
+computed on the western side, and then applied and rotated. Or with `AllSides`, all projectors
+are computed and applied simultaneously on all sides, where in particular the corners get
+contracted with two projectors at the same time.
```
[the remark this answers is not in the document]
pbrehmer: Yes, I like that much better; changed it!

pbrehmer:
I tried to fix the issue we were talking about earlier, where `FixedSVD` does not produce an element-wise converged environment if the supplied U, S and V come from a `IterSVD`, however, it still doesn't work. The U's and V's are normalized in the same way as with `TensorKit.SVD` and the singular values and the complete decompositions match. (I added a small extra testset for this.) This is really weird and I'm not so sure where to look next.

For now, I have added a safeguard in the constructor of `PEPSOptimize` which forbids using `IterSVD` together with the `:fixed` mode.

Once the tests turn green (I might need to stabilize them first), I would be fine with merging. What do you think? @lkdvos

src/algorithms/ctmrg.jl
```diff
@@ -330,13 +142,13 @@ function ctmrg_iter(state, env::CTMRGEnv{C,T}, alg::CTMRG) where {C,T}
...
-        env, _, _, ϵ₀ = left_move(state, env, alg.projector_alg)
+        env, info = left_move(state, env, alg.projector_alg)
         state = rotate_north(state, EAST)
         env = rotate_north(env, EAST)
-        ϵ = max(ϵ, ϵ₀)
+        ϵ = max(ϵ, info.ϵ)
     end
 
-    return env, ϵ
+    return env, (; ϵ)
```
lkdvos, line 151: While I like the namedtuples for the information, I don't think we should make the names unicode symbols, as that can be really annoying on certain platforms.

src/algorithms/ctmrg_gauge_fix.jl
```diff
@@ -0,0 +1,216 @@
...
+        ρinit = TensorMap(
+            randn,
+            scalartype(T),
+            MPSKit._lastspace(Tsfinal[end])' ← MPSKit._lastspace(M[end])',
+        )
+        ρprev = transfermatrix_fixedpoint(Tsprev, M, ρinit)
+        ρfinal = transfermatrix_fixedpoint(Tsfinal, M, ρinit)
+
+        # Decompose and multiply
+        Qprev, = leftorth(ρprev)
```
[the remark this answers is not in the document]
lkdvos: Was there a problem with the in-place versions? In principle, this should still be compatible with AD, because of how we structured our rrules for functions that destroy their inputs.

src/algorithms/peps_opt.jl
```diff
@@ -41,13 +86,36 @@ the CTMRG runs can be started on the converged environments of the previous opti
...
+    gradient_alg::G
+
+    function PEPSOptimize(  # Inner constructor to prohibit illegal setting combinations
+        boundary_alg::CTMRG{S},
+        optimizer,
+        reuse_env,
+        gradient_alg::G,
+    ) where {S,G}
+        if gradient_alg isa GradMode
+            if S == :sequential && G.parameters[1] == :fixed
```
lkdvos, line 102: I don't really like accessing the internals of a type like this. In particular, note that this would fail if you make a new subtype: `struct MyGradMode{T,F} <: GradMode{F} end`.
The more idiomatic way is something like:
```julia
iterscheme(alg::GradMode{F}) where {F} = F
```
which would then automatically extract the correct parameter for all subtypes

src/algorithms/peps_opt.jl
```diff
@@ -41,13 +86,36 @@ the CTMRG runs can be started on the converged environments of the previous opti
...
+                throw(ArgumentError(":sequential and :fixed are not compatible"))
+            elseif boundary_alg.projector_alg.svd_alg.fwd_alg isa IterSVD &&
+                G.parameters[1] == :fixed
```
lkdvos, line 105: Same comment here

src/algorithms/peps_opt.jl
```diff
@@ -102,23 +170,65 @@ Evaluating the gradient of the cost function for CTMRG:
...
+# Here f is differentiated from an pre-computed SVD with fixed U, S and V
+function _rrule(
+    gradmode::GradMode{:fixed},
+    ::RuleConfig,
+    ::typeof(MPSKit.leading_boundary),
+    envinit,
+    state,
+    alg::CTMRG{C},
+) where {C}
+    @assert C == :simultaneous
```
lkdvos, line 211:
```suggestion
    @assert C === :simultaneous
```
symbols should be slightly more efficient like this

lkdvos (changes requested):
I left some small comments, mostly the accessing type parameters I find important. Otherwise, there still seems to be some problems with the tests, so this might still require some stabilizing.
For the rest, I think I am happy with merging once these things are resolved!

src/algorithms/ctmrg.jl
pbrehmer: Yes, I think you mentioned this before and I forgot. Somehow I like unicode symbols a bit too much...

src/algorithms/peps_opt.jl
pbrehmer: The `.parameters` thing is a bad habit of mine! I should definitely write functions for that.

src/algorithms/ctmrg.jl
lkdvos: Me too :upside_down_face:

pbrehmer:
This should hopefully stabilize and speed up the test suite a bit. I decided to cut the p-wave superconductor from the `gradients.jl` test because it was very slow (especially the `nothing` setting which ADs everything) and also unstable. I don't think we have that example figured out properly yet, the differentiation doesn't seem to be very efficient and stable yet. Maybe I can try to make that run better in the future.

However, I modified the `gradparts.jl` test and actually added it to the test suite (it was just ignored previously). It now only tests for `ctmrg_iter` but for different CTMRG schemes (and on fermionic spaces.) That should be the most relevant setting anyway but I guess it would be good to add more functions to that test eventually.

Let me know if the modifications to the tests are fine, but I think these should run relatively stable and hopefully faster now!

lkdvos:
I have to say that I am not super happy with cutting the pwave tests, as there now is no full test for a fermionic model. In particular, this is very much not ideal, as this is definitely one of those things where we need to continuously check if we didn't mess up any of the twists by making changes, sometimes requiring multiplication and sometimes contraction, which would now not show up at all for example in the expectation_values etc.

pbrehmer:
I can totally see that; let me check if I can get it to run properly again...

lkdvos:
I would be completely fine with actually loading a pre-computed near-groundstate from disk if that solves the problem

pbrehmer:
That would be a good idea in the future, but currently I'm struggling to get converging gradients for the `:simultaneous` and `:fixed` schemes. Somehow the finite difference tests for `ctmrg_iter` work but the adjoint linear problem doesn't converge properly or the contangents are significantly sensitive to the gauge choice. While this occasionally comes up also with the other models with `|gauge| ~ 1e-9`, here for the fermionic model we have `|gauge|~ 1e-2` which really shouldn't happen.

I can't really tell yet why there is such a difference between the `:sequential` (the old way) and `:simultaneous` (the new way) schemes. For now, I can disable the `:simultaneous` tests for the p-wave superconductor and then it should run as before.

pbrehmer:
@lkdvos Do you have an idea where these `InexactError`s in the `gradparts.jl` test such as `Int64(99.99)` come from? I don't really understand what is happening under the hood and if this is easy to fix...

Otherwise the test should finally run through !

lkdvos:
I am not sure if we really want to include the gradpart tests. Quinten added these to debug on a bit finer level, but there is really no reason to check if the individual steps are compatible with finite differencing if a full gradient also works. (Also, they really are hacked together a bit...)

In particular, while I didn't investigate in detail, the `InexactError` is typically the rrule tests that attempts to generate random tangents for things like the algorithm etc, where somehow it thinks that it should generate `maxiter=99.9` type things. You could avoid this by manually passing the tangent to the `test_rrule` function, `arg \vdash NoTangent()` might do the trick

pbrehmer:
Hmm I guess you got a point, that it is a bit hacked together and that the full gradient is already tested elsewhere. I will then disable the test but it might be useful to leave it in the tests folder if needed in the future?

pbrehmer (commented):
I guess all of the requested changes are now in place :-)

lkdvos (approved):

PR #57, opened 2024-07-16
Add `MPSKit.correlation_length` method for PEPS

pbrehmer:
This PR will add a method to compute correlation lengths of a PEPS unit cell with an CTMRG environment. To that end we're gonna extend the `MPSKit.correlation_length` function and use the MPSKit glue machinery.

### DIFF
--- a/src/PEPSKit.jl
+++ b/src/PEPSKit.jl
@@ -71,7 +71,7 @@ module Defaults
 end
 
 export SVDAdjoint, IterSVD, NonTruncSVDAdjoint
-export FixedSpaceTruncation, ProjectorAlg, CTMRG, CTMRGEnv
+export FixedSpaceTruncation, ProjectorAlg, CTMRG, CTMRGEnv, correlation_length
 export LocalOperator
 export expectation_value, costfun
 export leading_boundary
--- a/src/algorithms/ctmrg.jl
+++ b/src/algorithms/ctmrg.jl
@@ -328,3 +328,35 @@ function LinearAlgebra.norm(peps::InfinitePEPS, env::CTMRGEnv)
 
     return total
 end
+
+"""
+    correlation_length(peps::InfinitePEPS, env::CTMRGEnv; howmany=2)
+
+Compute the PEPS correlation length based on the horizontal and vertical
+transfer matrices. Additionally the (normalized) eigenvalue spectrum is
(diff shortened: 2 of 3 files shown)


pbrehmer:
I had to extend the `MPSKit.mixed_fixpoints` method to also return the left and right eigenvalues. There might be better ways to do this, @lkdvos ?

Also currently the `left_transfer` contraction order seems to be incompatible with the environment edge spaces, so I will fix that tomorrow hopefully

lkdvos:
I think it would be nicer if we structure it a bit like in MPSKit and just add a specialized method to get the transfer spectrum (this might be of independent interest anyways).
https://github.com/QuantumKitHub/MPSKit.jl/blob/1ea5689a1370607a0f200ea4f56a84a363a81cfb/src/algorithms/toolbox.jl#L43-L56

This makes it so we don't have to change the return type (which I would not mind, but then I definitely also would do it in MPSKit), but also so you can probe correlation lengths in different sectors, as often the dominant correlation length is not necessarily associated to trivial observables.

pbrehmer:
Thanks for the pointer, I wasn't aware of the internal of MPSKit. I like the `transfer_spetrum` method much better.

The way it is implemented now is functional but I find it a bit inelegant that we need to redefine the spaces of the edges in the `below` multi-line MPS in order for the `transfer_left` contractions to work.  But maybe I'm overlooking something and there is an easy way to avoid this step?

pbrehmer:
The updated row/column indexing broke the VUMPS convergence in the boundary MPS test. I'm not sure why it doesn't work but it must have broken something under the hood. Maybe @lkdvos can you take a quick look once you have time?

lkdvos:
Ok! if tests pass, this should be good to go for me

lkdvos (dismissed):

pbrehmer:
Sorry, still wanted to add some small test coverage for the `correlation_length` function. I just added some basic checks to the `heisenberg.jl` and `tf_ising.jl` tests. This should run through.

Do you want to do the files reorganization PR before we tag the next release? I mean it's not super important but it would be nice to have (and probably not that much work). On the other hand, there's probably more important stuff to be done elsewhere... :-)

lkdvos:
I did want to have it done before the next release, there's something nice about having a cleaned up released version :)

lkdvos (approved):

PR #58, opened 2024-07-18
Reorganize files and clean up

pbrehmer:
In this PR we will reorganize the file structure a bit. Additionally, it will contain small clean ups and improvements.


pbrehmer:
@lkdvos I wanted to include the `ctmrgscheme` function and the cleaned up return values in `fixedpoint` in the release so I made it part of the "file reorganization" PR. (Especially the return values were needed, since the previous way actually ignored the convergence history which is of course important to know.)

Other than that, I think we only need to tidy up the files before tagging a new release :)

lkdvos:
Do you want me to do the first re-organisation? I don't mind doing it, we should just avoid doing it both at the same time

pbrehmer:
Yes, I think it would probably be more sensible if you do it since I don't have a good overview over all the MPSKit glue/boundary MPS stuff :)

lkdvos:
I'm definitely still in a WIP stage, I'll try to continue over the next couple days, but I'll need some time :)

lkdvos:
@pbrehmer , I started separating out some of the contractions, but I am running out of time a bit. If you or @leburgel are up for it, any chance you could do the same to the other contractions? I think it will make the code a lot more maintainable (and it takes out some of the duplicates)

pbrehmer:
> If you or @leburgel are up for it, any chance you could do the same to the other contractions?

Yes, I'll take a look this afternoon and start working on it! Having a separate catalog of contractions is a thing I wanted to implement for a longer time now anyway.

pbrehmer:
@lkdvos The majority of all contractions are now in the separate contraction files. However, I wasn't sure what to do with `derivatives.jl` ,`transferpepo.jl` and `transferpeps.jl`. Those files anyway need some overall cleaning up at some point (I think you mentioned that too) so I didn't move out any of the contractions yet.

Other than that, I hope that the tests run through now but I do think everything should work properly now. Let me know if there's anything else that I could help with.

lkdvos:
I'm honestly not entirely sure I want to touch the boundarymps stuff already. I think I'll have to really think about how the multiline stuff works before I want to spend a lot of time on it, so I'd say we postpone that for now?
I have to say I am quite happy with the progress for now, so once the tests turn green I'm ok with merging and tagging

pbrehmer:
I'd be totally fine with letting the boundary MPS stuff be for the moment. We did make quite some progress over the last weeks, so I'm also happy with that. The reorganization has helped a lot I think.

I will fix the tests later or tomorrow morning, then we can merge and tag!

pbrehmer:
Note to myself: always also test for non-symmetric unit cells...

Alright, I think this should be good to go once the tests pass (which they now hopefully do).

lkdvos (approved):

PR #60, opened 2024-08-13
Fix virtual space error for larger unit cells

lkdvos:
There was an error in the fixedspace truncation which cycled the spaces around, thus leading to errors in the gaugefixing.
This should now be resolved.

Additionally, I removed the gaugefixing procedure from the regular CTMRG calls, as the gaugefixing is repeated in the gradient rules anyways.

I tried to make both schemes of CTMRG use the same indexing convention for the projectors too now, this was an absolute mess so let's try and keep our coordinates the same now :laughing:


pbrehmer (approved):
Wow, thanks for catching that. I know the pain of figuring out the indexing conventions a bit too well, so I really appreciate that you fixed it :-)

PR #63, opened 2024-08-19
Define + for LocalOperators

Gertian:
I defined addition for Localoperators

### DIFF
--- a/src/operators/localoperator.jl
+++ b/src/operators/localoperator.jl
@@ -5,16 +5,31 @@ struct LocalOperator{T<:Tuple,S}
     lattice::Matrix{S}
     terms::T
 end
-function LocalOperator(lattice::Matrix{S}, terms::Pair...) where {S}
+function LocalOperator(lattice::Matrix{S}, terms::Pair... ; min_norm_operators = eps()^(3/4)) where {S}
     lattice′ = PeriodicArray(lattice)
+    relevant_terms = []
     for (inds, operator) in terms
+        # Check if the indices of the operator are valid with themselves and the lattice
         @assert operator isa AbstractTensorMap
         @assert numout(operator) == numin(operator) == length(inds)
         for i in 1:length(inds)
             @assert space(operator, i) == lattice′[inds[i]]
+        end        
+        # Check if we already have an operator acting on the coordinates
+        i = findfirst(existing_inds -> existing_inds == inds, map(first, relevant_terms))
+        if !isnothing(i) # We are adding to an existing operator
+            new_operator = relevant_terms[i][2] + operator
+            if norm(new_operator) > min_norm_operators
+                relevant_terms[i] = (inds => new_operator)
+            else
+                deleteat!(relevant_terms, i)
+            end
+        else # It's a new operator, add it if its norm is large enough
+            norm(operator) > min_norm_operators && push!(relevant_terms, inds => operator)
         end
     end
-    return LocalOperator{typeof(terms),S}(lattice, terms)
+    relevant_terms = Tuple(relevant_terms)
+    return LocalOperator{typeof(relevant_terms), S}(lattice, relevant_terms)
 end
 
 """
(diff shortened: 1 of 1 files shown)


src/operators/localoperator.jl
```diff
@@ -57,3 +57,17 @@ function Base.repeat(O::LocalOperator, m::Int, n::Int)
     end
     return LocalOperator(lattice, terms...)
 end
+
+function Base.:+(O1::LocalOperator, O2::LocalOperator)
+    if O1.lattice != O2.lattice
```
lkdvos, line 62: I think we already have a function for this check: `checklattice`.
Could you use that one (or overload it if it does not work already?)

src/operators/localoperator.jl
```diff
@@ -57,3 +57,17 @@ function Base.repeat(O::LocalOperator, m::Int, n::Int)
...
+        throw("lattices should match")
+    end
+    terms = []
+    for (inds, operator) in O1.terms
+        push!(terms, inds => operator)
+    end
+    for (inds, operator) in O2.terms
+        push!(terms, inds => operator)
+    end
+    return LocalOperator(O1.lattice, terms...)
```
lkdvos, line 72: I think the following might be a little easier:
```suggestion
    return LocalOperator(O1.lattice, O1.terms..., O2.terms...)
```
This being said, it might be worth it to check if there are duplicate coordinates, and in those cases add the operators together. This should improve efficiency.

lkdvos (commented):
Maybe in the meantime it might be convenient to also add `-`, and scalar multiplication?

Gertian:
> Maybe in the meantime it might be convenient to also add `-`, and scalar multiplication?

Sure, I didn't need those so I didn't think of it. 
I'll get to it in the near future.

src/operators/localoperator.jl
Gertian: I don't see how this would work ? The LocalOperator constructor takes a lattice and an Ntuple ? Now you're giving it 3 arguments no ? Or is this some sort of julia magic that I don't understand ?

src/operators/localoperator.jl
Gertian: This has been done !

src/operators/localoperator.jl
Gertian: I didn't do the suggested change. But I did implement the check for duplicate coordinates

Gertian:
It might be useful to also check when H1 - H2 contains operators that are 0 that can therefore be thrown away.
The issue is that this requires some sort of tolerance which is a bit ambiguous. 

Any suggestions/preferences ? Or do we just ignore this subtlety ?

src/operators/localoperator.jl
lkdvos: The constructor takes a lattice and a comma-separated-list of operators. Notice how the splatting (`...`) syntax just unpacks something into such a list, the following things are equivalent:
`LocalOperator(lattice, terms...) === LocalOperator(lattice, terms[1], terms[2], ..., terms[end])`
Thus, there is really no need to first make a vector that contains both, and then unpack them into such a list again. You might as well immediately unpack and concatenate.

lkdvos:
Honestly, maybe we just add a default tolerance for that? something like `norm < eps^(3/4)` (`1e-12`) sounds reasonable to me.
I think the cleaner approach is then to just ignore this in the `+` methods, and just handle this in the constructor (where we can have the tolerance as a kwarg).

Gertian:
Okay. I'll implement it like that.

lkdvos:
I made some slight changes and added a very basic docstring, thanks for getting this started!
Once the tests turn green, I think this is good to go from my side.

Gertian:
> I made some slight changes and added a very basic docstring, thanks for getting this started! Once the tests turn green, I think this is good to go from my side.

This all looks great for me. And you're welcome ! :)

pbrehmer (approved):
Also looks great to me, thanks for that! :-)

PR #65, opened 2024-08-28
Improve CTMRG error computation

lkdvos:
This should provide a slightly more robust error computation, which will no longer output `Inf` when the spaces are still changing. The only downside is that it might say that something is converged while it is not, because it only checks the difference on the overlap of the changing spaces.
In principle, we could still remedy this, but I would be okay with just having this as is and seeing if it ever shows up as an issue.

### DIFF
--- a/src/algorithms/ctmrg/gaugefix.jl
+++ b/src/algorithms/ctmrg/gaugefix.jl
@@ -139,10 +139,16 @@ function fix_global_phases(envprev::CTMRGEnv, envfix::CTMRGEnv)
     return CTMRGEnv(cornersgfix, edgesgfix)
 end
 
+#=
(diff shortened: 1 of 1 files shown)


lkdvos:
From the codecov it also seems like we are not testing any ctmrg run with changing spaces, so this might also be something to look into.

pbrehmer:
Thanks for the clean up, that's definitely an improvement! Previously (as you hint at in the code comment), changing spaces during a CTMRG run would error due differences between tensors on different spaces. But now this should possible. Maybe I will take a look after I'm back from vacation and come up with a simple test for varying spaces (maybe increasing the environment dimensions during the run or something like that...).

pbrehmer (approved):

lkdvos:
Definitely enjoy the holiday first ;)

PR #68, opened 2024-09-24
Change default values

pbrehmer:
This PR will change and more consistently use the default values in `PEPSKit.Defaults`.


lkdvos (approved):
I do think we definitely need some work on our defaults and kwarg interface, but I'm happy to already include these changes. I didn't check to see if the performance is better now, but I trust you:)

PR #70, opened 2024-09-27
Implement multi-threading using OhMyThreads and make it differentiable

pbrehmer:
Here we'll replace the `@fwdthreads` macro with `tmap` and `foreach` calls. Additionally, we will code up reverse rules such that the backwards pass also runs in parallel.


pbrehmer:
@lkdvos I could use some help on this one... I still find writing rrules very confusing sometimes, so could you maybe take a look at the rrule for `dtforeach`, once you have time? And while the `dtmap` rrule already seems to work, it might not be perfect yet :-)

Other than that we have to think about how to pass along the threading kwargs to the `dtmap` and `dtforeach` calls. One option is to store those kwargs inside the `CTMRG` struct but that feels a bit wrong. (Also there are some calls which wouldn't have access to a `CTMRG` instance.) Not sure what would be the best solution for that. Would global variables work where the user can set that somehow?

I'll also add some rrule tests in the end.

lkdvos:
I think tforeach is going to be a bit hard, because it has no outputs, and is thus necessarily in-place... I have no clue how the zygote buffer magic works, so I can't really say I know how to deal with that either.

I should have some more time to think this through next week though!

For the global variables, I would maybe suggest ScopedVariables.jl instead, this is a little more flexible and shouldn't incur too much runtime costs

pbrehmer:
> I think tforeach is going to be a bit hard, because it has no outputs, and is thus necessarily in-place... I have no clue how the zygote buffer magic works, so I can't really say I know how to deal with that either.

How about we just stick to `tmap` then? The only slightly annoying thing is having to separate multiple return values at different indices but that should not incur too much overhead.

> For the global variables, I would maybe suggest ScopedVariables.jl instead, this is a little more flexible and shouldn't incur too much runtime costs

Wasn't aware of ScopedValues.jl yet, that looks like a great solution. But I don't quite understand the necessity of a scoped value here since we never need to access the threading settings inside a multi-threaded map, right? In any case, we can probably just have a global scoped `Dict` with the threading settings that are passed on to the `dtmap` calls, and that can be mutated by some `set_thread_settings` function.

Anyways, I will give these things a go and then we can review next week, when you have time :)

src/utility/diffable_threads.jl
```diff
@@ -0,0 +1,71 @@
...
+)
+    el_rrules = tmap(A; kwargs...) do a
+        rrule_via_ad(config, f, a)
+    end
+    y = map(first, el_rrules)
+    function dtmap_pullback(dy_raw)
+        dy = unthunk(dy_raw)
+        backevals = tmap(CartesianIndices(A); kwargs...) do idx
+            last(el_rrules[idx])(dy[idx])
+        end
```
lkdvos, line 24: I think something like this should also work:
```suggestion
        backevals = tmap(el_rrules, dy; kwargs...) do (_, el_rrule), dy_
            last(el_rrule)(dy_)
        end
```

src/utility/diffable_threads.jl
```diff
@@ -0,0 +1,71 @@
...
+        df = ProjectTo(f)(sum(first, backevals))
```
lkdvos, line 25: Somehow it should be possible to do a `tmapreduce` and combine all of this, but I'll try and look into it a bit more

src/utility/diffable_threads.jl
```diff
@@ -0,0 +1,71 @@
...
+        dA = tmap(CartesianIndices(A); kwargs...) do idx
```
lkdvos, line 26: Here, I don't think it is really worth it to use `tmap`, there are no more computations (I think), it should just be unpacking the values...

lkdvos (commented):
I think I should have elaborated a bit more on what I had in mind with the scoped values: I would keep the threading strategies in scoped values, such that they can always be accessed as if they are global values. (this prevents them from bloating all of our algorithms)
The benefit of having them as a scoped value however means that users could still change them by calling the peps function from within a scope with a modified scoped value, thus changing the scheduler.

src/utility/diffable_threads.jl
pbrehmer: Oh yes definitely, probably creates more overhead

pbrehmer: > I think I should have elaborated a bit more on what I had in mind with the scoped values: I would keep the threading strategies in scoped values, such that they can always be accessed as if they are global values. (this prevents them from bloating all of our algorithms) The benefit of having them as a scoped value however means that users could still change them by calling the peps function from within a scope with a modified scoped value, thus changing the scheduler.

Alright, thanks for clarifying! The way the threading kwargs are defined should be fine anyway, right? I like having them inside the `Defaults`

pbrehmer:
I was trying to fix the Zygote error but with no luck, I really don't know how to handle the `NoTangent`s (`dA` is an `Array{NoTangent,3}`) as they are being converted to `Nothing`. I'll copy the error here for future reference:

```
ERROR: MethodError: no method matching length(::Nothing)
The function `length` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  length(::Zygote.Grads, Any...; kwargs...)
   @ Zygote ~/.julia/packages/MacroTools/Cf2ok/src/examples/forward.jl:17
  length(::Combinatorics.Partition)
   @ Combinatorics ~/.julia/packages/Combinatorics/Udg6X/src/youngdiagrams.jl:8
  length(::Base.MethodSpecializations)
   @ Base reflection.jl:1317
  ...

Stacktrace:
  [1] productfunc(xs::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}, dy::Array{Nothing, 3})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/lib/array.jl:278
  [2] (::Zygote.var"#collect_product_pullback#744"{Base.Iterators.ProductIterator{Tuple{…}}})(dy::Array{Nothing, 3})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/lib/array.jl:300
  [3] (::Zygote.var"#2941#back#745"{Zygote.var"#collect_product_pullback#744"{…}})(Δ::Array{Nothing, 3})
    @ Zygote ~/.julia/packages/ZygoteRules/M4xmc/src/adjoint.jl:72
  [4] ctmrg_renormalize
    @ ~/repos/PEPSKit.jl/src/algorithms/ctmrg/ctmrg.jl:383 [inlined]
  [5] (::Zygote.Pullback{Tuple{…}, Any})(Δ::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface2.jl:0
  [6] ctmrg_iter
    @ ~/repos/PEPSKit.jl/src/algorithms/ctmrg/ctmrg.jl:163 [inlined]
  [7] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::Tuple{CTMRGEnv{…}, Nothing})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface2.jl:0
  [8] #221
    @ ~/repos/PEPSKit.jl/src/algorithms/ctmrg/ctmrg.jl:118 [inlined]
  [9] (::Zygote.Pullback{Tuple{…}, Any})(Δ::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface2.jl:0
 [10] #35
    @ ~/.julia/packages/LoggingExtras/cFgEq/src/verbosity.jl:117 [inlined]
 [11] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface2.jl:0
 [12] (::Zygote.var"#ad_pullback#61"{Tuple{…}, Zygote.Pullback{…}})(Δ::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/chainrules.jl:264
 [13] with_logger_pullback
    @ ~/.julia/packages/ChainRules/vdf7M/src/rulesets/Base/CoreLogging.jl:12 [inlined]
 [14] (::Zygote.ZBack{ChainRules.var"#with_logger_pullback#862"{…}})(dy::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/chainrules.jl:212
 [15] #withlevel#34
    @ ~/.julia/packages/LoggingExtras/cFgEq/src/verbosity.jl:113 [inlined]
 [16] (::Zygote.Pullback{Tuple{…}, Any})(Δ::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface2.jl:0
 [17] withlevel
    @ ~/.julia/packages/LoggingExtras/cFgEq/src/verbosity.jl:107 [inlined]
 [18] (::Zygote.Pullback{Tuple{…}, Any})(Δ::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface2.jl:0
 [19] withlevel
    @ ~/.julia/packages/LoggingExtras/cFgEq/src/verbosity.jl:107 [inlined]
 [20] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface2.jl:0
 [21] leading_boundary
    @ ~/repos/PEPSKit.jl/src/algorithms/ctmrg/ctmrg.jl:115 [inlined]
 [22] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::CTMRGEnv{TrivialTensorMap{…}, TrivialTensorMap{…}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface2.jl:0
 [23] #37
    @ ./REPL[24]:2 [inlined]
 [24] (::Zygote.Pullback{Tuple{…}, Tuple{…}})(Δ::Float64)
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface2.jl:0
 [25] (::Zygote.var"#78#79"{Zygote.Pullback{Tuple{…}, Tuple{…}}})(Δ::Float64)
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface.jl:91
 [26] withgradient(f::Function, args::InfinitePEPS{TrivialTensorMap{ComplexSpace, 1, 4, Matrix{ComplexF64}}})
    @ Zygote ~/.julia/packages/Zygote/NRp5C/src/compiler/interface.jl:213
 [27] top-level scope
    @ REPL[24]:1
```

lkdvos:
I'm honestly not so sure how no one has ever run into this, but I seem to have been able to circumvent the issue by just not differentiating through the `collect(Iterators.product())`calls. It might have something to do with our `tmap`gradient returning an array of `nothing` instead of simply `nothing`, but I don't really have the time to investigate that and this seems to work.

pbrehmer:
I noticed that the default `ntasks` is always set to one since during pre-compilation `Threads.nthreads` always returns one. That's why I defaulted to `ntasks=4` since on most machines (if Julia is started with the corresponding number of threads) this will multi-thread CTMRG in the four spatial directions. What do you think about that?

Other than that this seems mergeable, thanks a lot for pinpointing and circumventing the error! It is really weird how this Zygote problem never came up before...

lkdvos:
I'm not entirely sure that I love having everything multithreaded by default, even if there are no threads available...

I'm actually quite confused by this behavior, I feel like this must be something that got changed in more recent versions of Julia, as we are also using this pattern in MPSKit, and at some point that was definitely working, but I might be wrong. Anyways, I asked the discourse:
https://discourse.julialang.org/t/evaluation-order-of-threads-nthreads/121989

I'll investigate a bit further, but in the meantime, how would you feel about something like the commit I just had?
It's a bit iffy since we are reaching into a private function of OhMyThreads, but it really does exactly what we want: you are allowed to specify `ntasks`, you can give a `Scheduler`, or a `Symbol`, ... The main thing is to write some documentation that captures all the `kwargs`, which can vary depending on the scheduler.
From what I can see, the default there is to use a `DynamicScheduler` with chunking enabled, which is probably reasonable, and it does a runtime check for the number of threads. I think the overhead of that should be fine, and overall it seems quite flexible that way.

If you agree, the PR requires a bit more cleanup in docstrings and removal of the previous infrastructure, but I am running low on time ;)

pbrehmer:
I really like the proposal! In fact, I was initially thinking about doing something like this but then I got too lazy I guess :-) I will write up the docstring and finish the rest tomorrow. Thanks!

pbrehmer:
This should finish the PR. I decided to not write out all schedulers and keyword arguments in the docstring since I would essentially just copy the [OhMyThreads docs](https://juliafolds2.github.io/OhMyThreads.jl/stable/refs/api/#Schedulers) which already contain a full description of the possible settings.

src/PEPSKit.jl
```diff
@@ -112,24 +110,24 @@ module Defaults
...
     # OhMyThreads scheduler defaults
     const scheduler = Ref{Scheduler}()
     function set_scheduler!()
         scheduler[] = Threads.nthreads() == 1 ? SerialScheduler() : DynamicScheduler()
         return nothing
     end
-    function set_scheduler!(scheduler...; kwargs...)
-        scheduler[] = OhMyThreads.Implementation._scheduler_from_userinput(scheduler...; kwargs...)
+    function set_scheduler!(sc; kwargs...)
+        scheduler[] = OhMyThreads.Implementation._scheduler_from_userinput(sc; kwargs...)
```
lkdvos, line 121: I do think you want the splatting here, as I think it would also work if you supply just `kwargs`, and not a `Symbol` or `Scheduler`. Alternatively, I vaguely recall seeing `NotGiven()` as a default in many of the OhMyThreads functions, so then I would at least supply that as a default argument. (also, `sc` is not really a better variable name than `scheduler` ;)

pbrehmer: I think actually just supplying kwargs is not allowed in `_scheduler_from_userinput` - I looked at the implementation and it turns out that that case is explicitly caught and throws an error, that's why I removed the splatting :)

Of course we can supply `NotGiven()` by default, then the kwargs are just passed onto the `DynamicScheduler` constructor. I'll do that tomorrow and then we can merge.

The problem with `scheduler` as a name was that it interfered with the `Defaults.scheduler` constant such that it was erroring whenever I called `set_scheduler!`, that's why I changed it.

lkdvos: Oh, that obviously makes sense, my bad

src/utility/diffable_threads.jl
```diff
@@ -50,14 +50,17 @@ or equivalently with
...
 For a detailed description of all schedulers and their keyword arguments consult the
 [`OhMyThreads` documentation](https://juliafolds2.github.io/OhMyThreads.jl/stable/refs/api/#Schedulers).
 
 To reset the scheduler to its default value, one calls `set_scheduler!` without passing
 arguments which then uses the default `DynamicScheduler()`. If the number of used threads is
 just one it falls back to `StaticScheduler()`.
 """
-set_scheduler!(scheduler; kwargs...) = Defaults.set_scheduler!(scheduler; kwargs...)
+set_scheduler!(scheduler=OhMyThreads.Implementation.NotGiven(); kwargs...) =
+    Defaults.set_scheduler!(scheduler; kwargs...)
```
lkdvos, line 63: why not just export the `set_scheduler` function from `Defaults`?

pbrehmer: Oh sure, not sure why I did it this way initially

pbrehmer:
Whoops somehow the tests are starting to fail which was not the case before the merge, need to investigate what's happening...

lkdvos:
I'm definitely not a big fan of fixing problems by just turning off the tests, so let's maybe try an see if we can reproduce this anyways? Given that it happens on all platforms, presumably it is a Julia specific problem, but I would still love to either know what causes it, so we can avoid it, or just fix it in the first place...

I guess the reason it only started turning up now is that we were not testing on julia 1.10 before, and instead had 1.9 an 1.11

pbrehmer:
> I'm definitely not a big fan of fixing problems by just turning off the tests, so let's maybe try an see if we can reproduce this anyways?

I definitely agree that turning off tests is not the way to go. However, I haven't managed to reproduce it in any way on Ubuntu with Julia 1.10.6 (and also other Julia versions) - the tests just run through without erroring or causing a segmentation fault, regardless of the number of threads.

> Given that it happens on all platforms, presumably it is a Julia specific problem

Weirdly, the examples run through on the LTS on macOS (whereas they do not on Ubuntu and Windows). So there might also be some interaction with the OS and multi-threading?

Unfortunately, I don't have access to macOS or Windows machines so I don't really know how to approach this or narrow this down. Also, I don't have that much more time to spend on this (and probably neither do you), so that's why I was looking for a quick solution :-) Anyways, if you come up with ideas to try out, let me know and I can revisit this next week.

pbrehmer:
Just a quick thought: Could it be that the OhMyThreads multi-threading is interfering with OpenBLAS multi-threading? If the CI runners have only 4 cores in total then choosing `JULIA_NUM_THREADS=4` might lead to problems if OpenBLAS also  tries to claim more than one thread.

lkdvos:
This is unlikely to be the case (but of course without an actual solution I can't exclude anything), because of Julia's task-based parallelism, and this basically being handled by the operating system. In principle there is nothing stopping you from running 20 tasks on a single threaded setup, the OS will switch between these tasks as usual, and everything should still work the same. Of course, there will only ever be one task running in that case, while you still pay the overhead of the tasks and context switching, so typically this hinders performance a bit, but it shouldn't break anything.

I'll try to spend some more time to investigate this week, I definitely don't feel comfortable releasing something that breaks on the LTS on all platforms, so if we can't figure this out we might have to restrict to v1.11, which I'd rather avoid

pbrehmer:
Oh well, you're right. I just tried to come up with some reason why the tests run through on all my Ubuntu machines on 1.10.6 and why they don't on the CI runners. Also the fact that the macOS examples do run through on 1.10.6 is very weird. This seems to happen deterministically (I reran the CI earlier) but it is somehow platform-specific.

Are the tests running through for you locally on an LTS release? I'm currently stumped on this one.

lkdvos:
If this runs through, I'll re-enable the tests for all OS, and then it is definitely my fault. I think I might have screwed up something with the caching of precompiled data, which might have not included the julia version in the hash, thus possibly leading to these weird kinds of errors. (Just guessing, still need to check) 
If this is the case, I'll fix and merge?

pbrehmer:
Alright, I see no more segmentation faults, thanks for taking a look and fixing it! This looks mergeable now :-)

lkdvos:
Let's just wait a second until the tests actually all turn green (seems like they are not running now for some reason that I don't understand)

PR #72, opened 2024-10-02
Implement sparse SVD with function handles

pbrehmer:
In this PR we will use function handles to perform the iterative SVD of the half-infinite environment in CTMRG. To that end, a struct will be implemented that stores all necessary tensors for the environment contraction and on which `tsvd!` can dispatch.

Before doing that, a long-standing issue will be resolved where `IterSVD` would lead to singular vectors which, when supplied to `FixedSVD`, would not lead to element-wise converged environments and thereby making it not possible to use the iterative SVD during PEPS optimization.

### DIFF
--- a/src/algorithms/ctmrg/ctmrg.jl
+++ b/src/algorithms/ctmrg/ctmrg.jl
@@ -80,11 +80,7 @@ function CTMRG(;
     ctmrgscheme::Symbol=Defaults.ctmrgscheme,
 )
     return CTMRG{ctmrgscheme}(
-        tol,
-        maxiter,
-        miniter,
-        verbosity,
-        ProjectorAlg(; svd_alg, trscheme, verbosity),
+        tol, maxiter, miniter, verbosity, ProjectorAlg(; svd_alg, trscheme, verbosity)
     )
 end
 
@@ -200,7 +196,6 @@ end
 # end
 function ctmrg_expand(dirs, state, envs::CTMRGEnv{C,T}) where {C,T}
     Qtype = tensormaptype(spacetype(C), 3, 3, storagetype(C))
-    # Qtype = EnlargedCorner{C,T,eltype(state),eltype(state)}
     Q = Zygote.Buffer(Array{Qtype,3}(undef, size(envs.corners)))
     drc_combinations = collect(Iterators.product(dirs, axes(state)...))
     @fwdthreads for (dir, r, c) in drc_combinations
--- a/src/algorithms/ctmrg/ctmrg.jl
+++ b/src/algorithms/ctmrg/ctmrg.jl
@@ -192,7 +192,7 @@ function ctmrg_expand(state, envs::CTMRGEnv, ::SimultaneousCTMRG)
 end
 # function ctmrg_expand(dirs, state, envs::CTMRGEnv)  # TODO: This doesn't AD due to length(::Nothing)...
 #     drc_combinations = collect(Iterators.product(dirs, axes(state)...))
-#     return map(idx -> enlarge_corner(idx, envs, state)(idx[1]), drc_combinations)
+#     return map(idx -> EnlargedCorner(state, envs, idx)(idx[1]), drc_combinations)
 # end
 function ctmrg_expand(dirs, state, envs::CTMRGEnv{C,T}) where {C,T}
     Qtype = tensormaptype(spacetype(C), 3, 3, storagetype(C))
@@ -200,51 +200,10 @@ function ctmrg_expand(dirs, state, envs::CTMRGEnv{C,T}) where {C,T}
     dirs_enum = [(i, dir) for (i, dir) in enumerate(dirs)]
     drc_combinations = collect(Iterators.product(dirs_enum, axes(state)...))
     @fwdthreads for (d, r, c) in drc_combinations
-        ec = enlarge_corner((d[2], r, c), state, envs)
+        ec = EnlargedCorner(state, envs, (d[2], r, c))
         Q[d[1], r, c] = ec(d[2])  # Explicitly construct EnlargedCorner for now
     end
     return copy(Q)
-end
-
-"""
-    enlarge_corner((dir, r, c), state, envs)
-
-Enlarge corner by constructing the corresponding `EnlargedCorner` struct.
-"""
-function enlarge_corner((dir, r, c), state, envs)
-    if dir == NORTHWEST
-        return EnlargedCorner(
-            envs.corners[NORTHWEST, _prev(r, end), _prev(c, end)],
-            envs.edges[WEST, r, _prev(c, end)],
-            envs.edges[NORTH, _prev(r, end), c],
-            state[r, c],
-            state[r, c],
-        )
-    elseif dir == NORTHEAST
-        return EnlargedCorner(
-            envs.corners[NORTHEAST, _prev(r, end), _next(c, end)],
-            envs.edges[NORTH, _prev(r, end), c],
-            envs.edges[EAST, r, _next(c, end)],
-            state[r, c],
-            state[r, c],
-        )
-    elseif dir == SOUTHEAST
-        return EnlargedCorner(
-            envs.corners[SOUTHEAST, _next(r, end), _next(c, end)],
-            envs.edges[EAST, r, _next(c, end)],
-            envs.edges[SOUTH, _next(r, end), c],
-            state[r, c],
-            state[r, c],
-        )
-    elseif dir == SOUTHWEST
-        return EnlargedCorner(
-            envs.corners[SOUTHWEST, _next(r, end), _prev(c, end)],
-            envs.edges[SOUTH, _next(r, end), c],
-            envs.edges[WEST, r, _prev(c, end)],
-            state[r, c],
-            state[r, c],
-        )
-    end
 end
 
 # ======================================================================================== #
--- a/src/algorithms/ctmrg/ctmrg.jl
+++ b/src/algorithms/ctmrg/ctmrg.jl
@@ -195,7 +195,7 @@ function ctmrg_expand(state, envs::CTMRGEnv{C,T}, ::SequentialCTMRG) where {C,T}
         Q_nw[r, c] = enlarge_northwest_corner((r, c), envs, state)
     end
 
-    return copy(Q_sw), copy(Q_nw)
+    return stack((copy(Q_sw), copy(Q_nw)); dims=1)
 end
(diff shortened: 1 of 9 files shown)


lkdvos:
I think if you change the initial guess to be a slight deviation from the first singular vector, you'll have better performance. I think you need a small component orthogonal to that vector to start generating new vectors in the krylov subspaces.

pbrehmer:
So in principle this should be a relatively complete implementation of "function handle CTMRG", however it's still untested and there are some things to discuss/fix:

- somehow the way the adjoint action of `HalfInfiniteEnv` is defined leads to errors when calling `tsvd!` and I haven't figured out why
- implementing this took quite a bit of extra compatibility methods and some rewriting, so I'm sure one can improve on some of the design choices

Maybe we could discuss, review and fix these things next week or so @lkdvos.

src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -183,45 +189,28 @@ There are two modes of expansion: `M = :sequential` and `M = :simultaneous`.
...
-            enlarge_northeast_corner((r, c), envs, state)
-        elseif dir == SOUTHEAST
-            enlarge_southeast_corner((r, c), envs, state)
-        elseif dir == SOUTHWEST
-            enlarge_southwest_corner((r, c), envs, state)
-        end
+function ctmrg_expand(state, envs::CTMRGEnv, alg::SimultaneousCTMRG)
+    drc_combinations = collect(Iterators.product(1:4, axes(state)...))
+    return map(drc_combinations) do (dir, r, c)
+        enlarge_corner(Val(dir), (r, c), envs, state, alg)
```
lkdvos, line 201: I'm not sure this solution with `Val` is really what we want, this effectively replaces a runtime `if` check with a runtime dispatch, as the compiler cannot figure out which `enlarge_corner` to call from the types of `dir, r, c`. It might be the case that this is const-propagated, but I don't think this is really necessary, the overhead of the simple if statements should really be completely negligeable...

I am completely fine with moving the `if` clauses into `enlarge_corner`, I would just refrain from using dispatch to select the method here.
Am I missing some specific benefit of doing it this way?

src/algorithms/ctmrg/sparse_environments.jl
```diff
@@ -0,0 +1,265 @@
...
+end
+function TensorKit.block(env::HalfInfiniteEnv, c::Sector)
+    return env
+end
+function TensorKit.tsvd!(f::HalfInfiniteEnv; trunc=NoTruncation(), p::Real=2, alg=IterSVD())
+    return _tsvd!(f, alg, trunc, p)
+end
+function TensorKit.MatrixAlgebra.svd!(env::HalfInfiniteEnv, args...)
+    return TensorKit.MatrixAlgebra.svd!(env(), args...)  # Construct environment densely as fallback
+end
```
lkdvos, line 233: This somehow confuses me, I would not have guessed that if we dispatch to KrylovKit, we would need this TensorKit interop...

src/utility/svd.jl
lkdvos, line 1: I looked at these changes in a bit more detail, and I think we will probably have to commit to a better `svd` interface before we can make this work, as this definitely seems a bit unwieldy.
I think the main issue comes from the fact that we ended up merging the "compute pullback of svd block" step with the "merge pullbacks back into `TensorMap` form" step. I have some time on the flight tomorrow, I'll try and see if I cannot make this work a bit cleaner.

The idea would be something like: if we can have a `stack`-like function for TensorMap's, which is compatible with AD, it should be possible to use the KrylovKit rrule for the first step described above, while letting the AD engine combine the pullbacks automatically. This should then avoid this code-duplication, and make it more robust for function handles as well.

src/algorithms/ctmrg/sparse_environments.jl
```diff
@@ -0,0 +1,265 @@
+const SparseCTMRG = CTMRG{M,true} where {M}  # TODO: Is this really a good way to dispatch on the sparse CTMRG methods?
```
lkdvos, line 1: I would probably put this as a regular field. I am starting to feel less and less happy with all our type parameters, as it makes things a lot less transparent...

lkdvos (commented):
I left some specific remarks in the code, but maybe let me summarize:

- We should probably be a little careful with trying to push too many things into the type domain. The runtime benefit is not obvious, as it would just compile away a couple if checks, and the compiletime will increase, as well as the code complexity.
- The main source of misery here seems to be that we should rethink a good interface for `svdsolve` with both `TensorMap`s and `Function`s. I will attempt to come up with something better (probably tomorrow, otherwise coming week)

As a more general remark, I do really like the sparse environment structs, and would probably structure the code to always construct them first, and then, depending on the algorithm, either convert to a dense environment TensorMap or not. This would probably (hopefully) organise a bit of the code a bit more, such that the responsability of dealing with these objects is centralized to these structs.

src/algorithms/ctmrg/ctmrg.jl
pbrehmer: You're probably not missing anything :-) I just needed to rewrite the `ctmrg_expand` a bit to make it compatible with the `EnlargedCorner` struct and reduce code duplication. Then I somehow decided to move away from the if statement.

I'll move the if clause back inside `enlarge_corner` then!

src/utility/svd.jl
pbrehmer: Yes, I definitely agree. The way the interface is designed right now makes things tricky but there needs to be some compatibility layer that translates between the `svdsolve` calls and the block `TensorMap` structure that also needs to be generic enough to somehow work for function handles. I'd really welcome a better solution than what we have right now.

src/algorithms/ctmrg/sparse_environments.jl
pbrehmer: > As a more general remark, I do really like the sparse environment structs, and would probably structure the code to always construct them first, and then, depending on the algorithm, either convert to a dense environment TensorMap or not. This would probably (hopefully) organise a bit of the code a bit more, such that the responsability of dealing with these objects is centralized to these structs.

When I was writing these struct I was also thinking that it would be more useful to center the CTMRG function around them. That would definitely organize the code a bit more naturally. I'll get to it next week!

pbrehmer:
Had to add some extra `renormalize_corner` methods to make `EnlargedCorner` work with CTMRG. For some reason CTMRG doesn't want to be differentiated anymore (...well when does AD ever work right away anyway?) Tried to find the problem, but no success yet

pbrehmer:
Wow, only now I figured out that the problem with the iterative SVD and the `:fixed` mode leading to non-converging environments was due to wrong daggers in the relative gauge fixing contractions. (Big thanks to Bram on this one!)

Weirdly enough, the TensorKit `tsvd` and iterative SVDs with weird initial guesses retain some sort of symmetry, where the relative phases are just $\pm 1$; if, however, the initial guess is random, the relative phases will be complex which led the gauge fixing procedure to fail since the daggers were in the wrong place... Anyways, this is quite an important fix!

pbrehmer:
This should be finished up to the function handle SVD aspects - but those we'll anyway implement later on. Once the KrylovKit/TensorKit SVD interface is implemented, the function handles should also be very easy to add.

src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -177,51 +176,86 @@ ctmrg_logcancel!(log, iter, η, N) = @warnv 1 logcancel!(log, iter, η, N)
...
     end
-
     return copy(Q)
 end
 
+"""
+    enlarge_corner((dir, r, c), state, envs)
+
+Enlarge corner by constructing the corresponding `EnlargedCorner` struct.
+"""
```
lkdvos, line 213: How would you feel about just making this a constructor overload: `EnlargedCorner(state, envs, coordinates)`?

src/algorithms/ctmrg/sparse_environments.jl
```diff
@@ -0,0 +1,176 @@
...
+    ket::A
+    bra::A′
+end
+
+"""
+    (Q::EnlargedCorner)(dir::Int)
+
+Contract enlarged corner where `dir` selects the correct contraction direction,
+i.e. the way the environment and PEPS tensors connect.
+"""
```
lkdvos, line 23: This is a bit of a weird signature to me. Why do you want to overload "function calling" in order to contract the corner?

src/utility/svd.jl
lkdvos, line 1: I'm still not too happy with this, but I am fine merging while I work on the replacement :)

lkdvos (commented):
I have just a few small comments, otherwise I think I'm okay with merging.

src/algorithms/ctmrg/sparse_environments.jl
pbrehmer: I was thinking of `EnlargedCorner`s as a lazy object just like `HalfInfiniteEnv`, where the function call instantiates it, i.e. contracts the object. For `HalfInfiniteEnv` this is necessary to make it (eventually) compatible with the KrylovKit function handle interface so I wanted to have a similar design for `EnlargedCorner`. I hope that makes sense.

Would you have an alternative which you feel is more fitting here?

src/algorithms/ctmrg/ctmrg.jl
pbrehmer: Oh I'd definitely prefer that! Seems much cleaner

src/algorithms/ctmrg/sparse_environments.jl
lkdvos: I understand that, you want to think about the object as a linear map, and then function application does make sense, I'm just a bit confused about function application on an integer

pbrehmer: Yes, from that point of view it is a bit confusing. I just used the function call to pass along the directional argument. If I think about the integers as the constants `NORTHWEST`, `NORTHEAST` and so on, I don't find it as confusing but the main reason for me to keep it this way is that I don't want to make the design inconsistent with `HalfInfiniteEnv`.

lkdvos: Can you change both to become `TensorMap(sparsetensorstruct, additional_args...; additional_kwargs...)` instead? I don't think "instantiate as an actual tensor" lines up that nicely with "function call"

pbrehmer: Sure, I also like that and it is definitely clearer what happens from a user point of view

lkdvos (approved):

PR #78, opened 2024-10-24
Make `sdiag_inv_sqrt` use pseudo-inverse

### LINKED ISSUE #73: pseudo inverse in sdiag_inv_sqrt?
For large bound dimension CTM environment, the inverse of small svd spectrum can be very large and finally may cause an instability to an inf gradient. It would be helpful to set a cutoff in the sdiag_inv_sqrt and the rrule of it.

pbrehmer:
Implements pseudo-inverse for the inverse square root inside of `sdiag_inv_sqrt` and its reverse-rule.

Closes #73 as suggested by @huangrzh.

### DIFF
--- a/test/ctmrg/gradients.jl
+++ b/test/ctmrg/gradients.jl
@@ -32,7 +32,7 @@ gradmodes = [
         LinSolver(; solver=KrylovKit.GMRES(; tol=gradtol), iterscheme=:diffgauge),
     ],
     [  # Only use :diffgauge due to high gauge-sensitivity (perhaps due to small χenv?)
-        # nothing,
+        nothing,
         GeomSum(; tol=gradtol, iterscheme=:diffgauge),
         ManualIter(; tol=gradtol, iterscheme=:diffgauge),
         LinSolver(; solver=KrylovKit.GMRES(; tol=gradtol), iterscheme=:diffgauge),
--- a/test/ctmrg/gradients.jl
+++ b/test/ctmrg/gradients.jl
@@ -18,12 +18,7 @@ names = ["Heisenberg", "p-wave superconductor"]
 
 gradtol = 1e-4
 boundary_algs = [
-    CTMRG(;
-        tol=1e-10,
-        verbosity=0,
-        ctmrgscheme=:simultaneous,
-        svd_alg=SVDAdjoint(; fwd_alg=TensorKit.SVD(), rrule_alg=GMRES(; tol=1e-10)),
-    ),
+    CTMRG(; tol=1e-10, verbosity=0, ctmrgscheme=:simultaneous),
     CTMRG(; tol=1e-10, verbosity=0, ctmrgscheme=:sequential),
 ]
 gradmodes = [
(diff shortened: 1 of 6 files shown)


pbrehmer:
I'm not super sure about the default inverse cutoff tolerance to choose here; any opinions on this? @lkdvos

lkdvos:
I think TensorKit uses a `tol` which corresponds to `eps(real(float(scalartype(T))))^(3/4) * svals[1]` in the `tsvd` rrule, maybe we could just use the same?

pbrehmer:
I like that TensorKit way of handling the singular value cutoff tolerance. I have decided to again cut the `inv_sqrt_tol` field from `ProjectorAlg` since that is probably anyway a too fine-grained parameter. Also there is not a good consistent way to tweak that same tolerance inside the rrule of `tsvd!` from within PEPSKit, so I think it's better to just leave that parameter away and define a reasonable default.

lkdvos (dismissed):

pbrehmer:
Also updated the p-wave tests a bit and again found it weird that apparently the p-wave differentiation in `:fixed` mode leads to very high gauge-sensitivity. (This is not the case for the `:diffgauge` mode.) My suspicion is that this is improved with the fix in #72, where the gauge-fixing contractions had daggers in the wrong place and perhaps these daggers matter in the fermionic case. I will investigate further with the Hubbard model...

test/ctmrg/gradients.jl
```diff
@@ -36,8 +31,8 @@ gradmodes = [
         LinSolver(; solver=KrylovKit.GMRES(; tol=gradtol), iterscheme=:fixed),
         LinSolver(; solver=KrylovKit.GMRES(; tol=gradtol), iterscheme=:diffgauge),
     ],
-    [
-        nothing,
+    [  # Only use :diffgauge due to high gauge-sensitivity (perhaps due to small χenv?)
```
lkdvos, line 34: did you re check this after the gaugefixing fixes?

test/pwave.jl
```diff
@@ -10,7 +10,13 @@ unitcell = (2, 2)
...
 χbond = 2
 χenv = 16
-ctm_alg = CTMRG(; tol=1e-8, maxiter=150, verbosity=2, ctmrgscheme=:sequential)
+ctm_alg = CTMRG(;
+    tol=1e-8,
+    maxiter=150,
+    verbosity=2,
+    ctmrgscheme=:simultaneous,
+    svd_alg=SVDAdjoint(; rrule_alg=Arnoldi(; tol=1e-9, krylovdim=χenv + 30)),
+)
```
lkdvos, line 19: at the risk of sounding like a broken record, I really don't like that we have to change the settings of all algorithms on every single update of the package. We really have to get some better defaults in, because this is not really feasible in the long run. (not something that has to be fixed in this pr)

lkdvos (dismissed):
Looks good to me. If you checked the gauge sensitivity again after the fixes, Im okay with merging this like this.

test/pwave.jl
pbrehmer: You definitely have a point in repeating that and I completely agree :-) Actually, for most of the tests we should already be able to replace all settings with the default ones I just haven't gotten around to change them. I just thought that I would wait until we anyway redesign our interface and code up some kind of default parameter selection thing.

test/ctmrg/gradients.jl
pbrehmer: Okay I checked again and the gauge-fixing fixes don't make a difference to the high gauge-sensitivity. I will have to investigate and fix that in a different PR. I just haven't played around that much with fermions yet.

I put back in `nothing` to the test, since I accidentally left that commented out as I was trying around.

lkdvos (approved):
OK if tests turn green.

PR #80, opened 2024-10-26
Add Hubbard model

lkdvos:
This overloads the Hubbard model from MPSKitModels.jl

TODO:
- [ ] find some benchmarkdata to test
- [ ] implement fixed particle number

### DIFF
--- a/src/operators/lattices/squarelattice.jl
+++ b/src/operators/lattices/squarelattice.jl
@@ -12,6 +12,8 @@ struct InfiniteSquare <: AbstractLattice{2}
     end
 end
 
+Base.size(lattice::InfiniteSquare) = (lattice.Nrows, lattice.Ncols)
+
 function vertices(lattice::InfiniteSquare)
     return CartesianIndices((1:(lattice.Nrows), 1:(lattice.Ncols)))
 end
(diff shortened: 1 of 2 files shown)


lkdvos:
I should probably be a little more careful with what I meant. Indeed, there is no problem using this function with imposing FermionNumber, which is just a combination of U1 and FermionParity. The subtlety lies in how symmetric tensors work: all tensors always have to have total charge 0. In essence, this means that imposing FermionNumber automatically makes you impose total particle number 0.
It is definitely possible to select a different filling, but this requires injection of U1 charges per site, typically also resulting in unit cells, and I have not yet figured out a good framework to do this.

leburgel:
It seems a Hubbard Hamiltonian was added in #97, so this is probably superseded? The TODOs still remain valid though, would it be better to keep this open, or to close and move the TODOs to an issue?

lkdvos:
Either is fine, I'd say pick the thing that gives the least headaches with merging/branching/...

leburgel:
I filed an issue about the fixed particle number density, but we should anyway only look at that after it's been implemented in MPSKitModels.jl itself in the first place. Regarding the benchmark data, there's a comparison to an auxiliary field Monte Carlo result in the examples (it's not in the actual tests just because the runtime is very long). So I'm in favor of closing this.

PR #83, opened 2024-10-31
Change default SVD algorithm to `SDD` and update tests

pbrehmer:
This PR will change the default SVD algorithm to `TensorKit.SDD` since this generally outperforms `TensorKit.SVD` and is used as the default algorithm in TensorKit (and Base uses `gesdd`). Since this anyways modifies many tests, I will update the tests such that they all use the updated `Defaults` to the fullest possible extent.


lkdvos (approved):
I really like that we are using more Defaults in the tests and reduce the kwargs, looks great for me!

PR #84, opened 2024-11-01
Remove CTMRG redundant logging

Yue-Zhengyuan:
This PR removes a line of redundant logging in the `leading_boundary` function.


lkdvos (approved):
Thanks for that, this is better indeed!

PR #85, opened 2024-11-01
CI updates

lkdvos:
This PR sets the minimal julia version to be 1.10, the current long term stable version.
Additionally, I have updated the CI to reflect this, and used the structure of our tests to have parallelism on the machine level, ie using multiple github runners.


pbrehmer (approved):
Looks great!

Issue #88, opened 2024-11-05
Possible improvement of sequential CTMRG


Yue-Zhengyuan:
This issue is only a suggestion and I may have some misunderstanding... Let me focus on the left move of sequential CTMRG:
```
    -----> 
    C1 - T1 -   r-1
    |    ‖
    T4 = M' =   r
    |    ‖
    C4 - T3 -   r+1
    c-1  c 
```
Currently, it appears that the enlarged NW/SW corners for all sites and projectors on all columns are found at once, and the renormalization of one column is independent of the renormalization of the previous column. However, I only handle one column each time (this is also the approach in the full update). After I update C1, T4, C4 in the (c-1)-th column, when finding the projectors for the c-th column, I will use the *updated* value of C1, T4, C4 in the (c-1)-th column. Naively thinking, this may help speed up convergence a little.

Is my understanding correct? If so, I would also like to know why you decided to use the current approach. Thanks!

pbrehmer:
Thank you for raising this question! You are right, currently all rows and columns are renormalized independently in the `:sequential` mode. One advantage of this is that you can of course parallelize this step which might pay off for larger unit cells. However, in terms of convergence speed it might be advantageous to do what you suggest (and what is probably more conventional anyways): to enlarge and renormalize also the unit cell sequentially.

The reason why we decided to do this is probably because it was part of the old PEPSKit code and there doesn't seem to be an immediate obvious difference in the convergence speed between the `:simultaneous` and `:sequential` modes. That being said, we never really benchmarked the two modes against each other, so it is up for discussion if there are cases where `:sequential` significantly outperforms `:simultaneous` (i.e. where simultaneous renormalization of the unit cell is worse than a sequential renormalization).

Now that I think about it, I find your suggestion to be the better choice for `:sequential`. Let me open up a PR and implement it - it shouldn't be too much of an effort.

pbrehmer:
Implemented in #90 and refactored in #99.

PR #90, opened 2024-11-05
Make `:sequential` act column-wise

pbrehmer:
This PR possibly improves the `:sequential` CTMRG scheme, as suggested by @Yue-Zhengyuan in #88 by updating corners and edges column-wise instead of enlarging and renormalizing all environment tensors independently in the unit cell - previously only the four spatial directions were computed sequentially, now both the spatial directions and the unit cell columns are computed sequentially. This also seems to be the convention when looking at the [literature](https://scipost.org/SciPostPhysLectNotes.25).


pbrehmer:
Weirdly, I find that even for `ntasks=1` the `:sequential` mode is quite a bit slower than `:simultaneous` (for small dimensions). However, I would expect that for larger bond and environment dimensions this difference vanishes since the SVD runtime should dominate the CTMRG iterations.

lkdvos:
Any chance you can benchmark the before and after for the sequential mode? I would love to know if we should really have both options, or can just always enforce one or the other...

pbrehmer:
Sure I will take a look at the performance before and after making the column sequential!

What I find a bit concerning is that the tests fail for large unit cells and evaluating the energy of a random PEPS: somehow the new `:sequential` produces a slightly different result already in the 4th digit or so. On the one hand this is not surprising since the environments are really converged in a different way, but I wouldn't have expected the difference to be this big. I would assume that during optimization this effect is compensated by the optimizer somehow, so it might not be problem.

pbrehmer:
So I played around with the old and new `:sequential` implementations and I found the new implementation to be a bit faster in the reverse pass - but keep in mind that I didn't have the time to thoroughly benchmark this.

I first ran a couple of LBFGS optimization steps with the Heisenberg Hamiltonian to "thermalize" the PEPS. Then I timed a single CTMRG iteration and the differentiation of `leading_boundary`. For $D=3$, $\chi=64$ and a 2x2 unit cell I get for the old `:sequential` implementation

Screenshot from 2024-11-06 17-12-00

The new column-wise dependent `:sequential` seems to differentiate faster and for comparison I also added the same benchmark using `:simultaneous` (which is using 4 Julia threads):

Screenshot from 2024-11-06 16-41-58

I don't really see the necessity to have both the old and new style of `:sequential` implemented and we can probably just stick with the new column-wise dependent implementation. What do you think? @lkdvos

lkdvos:
I changed some small things to excise the "expansion" from the sequential CTMRG scheme. I'm not entirely happy with the result still, but I don't really have the time to invest here.
If the tests turn green, this is good to go for me!

Just to jot down some thoughts:

It seems like the different flavours we have currently really are wildly different, so I would even be happy to just separate them out into two files. Having the different parts of the algorithms interleaved is really more annoying than it helps to organise, I feel.

lkdvos (approved):

pbrehmer:
> It seems like the different flavours we have currently really are wildly different, so I would even be happy to just separate them out into two files. Having the different parts of the algorithms interleaved is really more annoying than it helps to organise, I feel.

Yes, I definitely agree and it would probably facilitate implementing different projector algorithms anyway. I will separate the two flavors then.

Issue #89, opened 2024-11-05
Provide options to silence warning but still print other information of CTMRG


Yue-Zhengyuan:
I feel that the current CTMRG verbosity settings is not well suited for applications that do not involve backprop. 

I frequently encounter states (obtained from simple/full update) that produces degenerate singular values of the CTMs; when performing CTMRG to measure physical quantities, there will be too many warnings; but I only want to look at the change of singular values and norm to determine convergence. Can you please add an option that still prints `Info` (at beginning/end or at each step) but silences all `Warning`s (in particular, the degenerate singular values warning)?

lkdvos:
This is a good suggestion. I never really considered this, but I guess the degenerate singular values are only an issue if we are using back propagation? In that case, it might also be possible to move that warning to only appear whenever the code is being differentiated.

PR #91, opened 2024-11-06
Add simple and full update algorithms

Yue-Zhengyuan:
This PR adds the core part of two (imaginary) time evolution algorithms: simple update ([arXiv 0806.3719](https://arxiv.org/abs/0806.3719)) and full update ([arXiv 1503.05345](https://arxiv.org/abs/1503.05345)). Despite being (much) less accurate than gradient-based ground state searching algorithms, states obtained from imaginary time evolution can usually serve as a good initialization for the former. 

Main changes:

- Added a new struct `SUWeight`, which are diagonal Schmidt weight matrices on the bonds of the iPEPS and serves as a crude approximation of the environment of each site tensor. 
- Three more functions are exported from the main module: `absorb_wt` is used to absorb the `SUWeight` into the iPEPS site tensors; `simpleupdate!` update the iPEPS site tensors and `SUWeight` in place; `fullupdate!` update the iPEPS and the `CTMRGEnv` in place.
- The full update <ins>*depends*</ins> on the column-wise sequential CTMRG (#90) recently implemented by @pbrehmer. I picked out a sub-function `ctmrg_leftmove` from `ctmrg_iter`, which only perform the left move for one column, and is used in the full update code. 
- Added some functions to perform in-place rotation and reflection of the iPEPS and the weights.

Limitations:

- The full update currently cannot handle iPEPS with *bipartite* structure (the unit cell size is 2 x 2, but the tensor at (1, 1) or (1, 2) is the same as the tensor at (2, 2) or (2,1)).
- For my convenience, I made some requirements on whether certain virtual spaces of iPEPS tensors are dual spaces. Hopefully they can be relaxed in the future. 
- Both algorithms can currently only calculate models with nearest neighbor interaction only. Actually they can handle next nearest neighbor terms (see [arXiv 1908.09359](https://arxiv.org/abs/1908.09359) and [arXiv 1711.07584](https://arxiv.org/abs/1711.07584)), but that requires more effort to implement, and is not of my immediate interest. 
- I leave out the part that determines whether the update has converged, since they very much depends on the physical model in interest. 

Improvements that can be done more easily:

- [ ] Improve docstring.
- [ ] Use Julia's more advanced logging system instead of simply printing the information.
- [ ] Add test scripts and illustrating examples.
- [x] Silence the degenerate singular value warning (#89). 

Comments and suggestions are welcome! Hopefully these can be useful additions to the package.

### DIFF
--- a/test/heisenberg_sufu.jl
+++ b/test/heisenberg_sufu.jl
@@ -23,7 +23,7 @@ wts = SUWeight(
 )
 # normalize peps
 for ind in CartesianIndices(peps.A)
-    peps.A[ind] /= PEPSKit.maxabs(peps.A[ind])
+    peps.A[ind] /= norm(peps.A[ind], Inf)
 end
 # Heisenberg model Hamiltonian
 ham = gen_gate()
--- a/src/algorithms/time_evolution/fullupdate.jl
+++ b/src/algorithms/time_evolution/fullupdate.jl
@@ -144,8 +144,8 @@ function update_column!(
             aR, bL, aR2bL2, env; maxiter=maxiter, maxdiff=maxdiff, verbose=false
         )
         costs[row] = cost
-        aR /= maxabs(aR)
-        bL /= maxabs(bL)
+        aR /= norm(aR, Inf)
+        bL /= norm(bL, Inf)
         localfid += local_fidelity(aR, bL, _combine_aRbL(aR0, bL0))
         #= update and normalize peps, ms
 
@@ -163,7 +163,7 @@ function update_column!(
         )
         # normalize
         for c_ in [col, cp1]
-            peps.A[row, c_] /= maxabs(peps.A[row, c_])
+            peps.A[row, c_] /= norm(peps.A[row, c_], Inf)
         end
     end
     # update CTMRGEnv
@@ -187,7 +187,7 @@ function fu_iter!(
     Dcut::Int,
     chi::Int,
     svderr::Float64=1e-9;
-    cheap=false,
+    cheap=true,
 )
     Nr, Nc = size(peps)
     fid, maxcost = 0.0, 0.0
@@ -230,7 +230,7 @@ function fullupdate!(
     rgtol::Float64=1e-6,
     rgmaxiter::Int=10,
     ctmrgscheme=:sequential,
-    cheap=false,
+    cheap=true,
 )
     time_start = time()
     N1, N2 = size(peps)
--- a/src/algorithms/time_evolution/simpleupdate.jl
+++ b/src/algorithms/time_evolution/simpleupdate.jl
@@ -155,7 +155,7 @@ function _su_bondx!(
     # update tensor dict and weight on current bond 
     # (max element of weight is normalized to 1)
     peps.A[row, col], peps.A[row2, col2] = T1, T2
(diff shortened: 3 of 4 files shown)


lkdvos:
Thanks for this work! I will do my best to review in a lot more detail as soon as possible, but I want to thank you in advance for wanting to contribute. This is definitely appreciated!

src/states/suweight.jl
lkdvos, line 1: If I understand everything correctly, it might make sense to bundle this together with the peps itself, ie something like this:
```julia
struct InfiniteWeightPEPS{T,E} <: AbstractPEPS
    vertices::Matrix{T}
    horizontal_weights::Matrix{E}
    vertical_weights::Matrix{E}
end
```

This might be easier to work with?

lkdvos (commented):
My comments consist mostly of some minor naming preferences, and maybe slight alterations to some of the code organization, but overall this looks great!

For function names as well as files, I would favor to just write names out in full, rather than having shorter names. This is definitely not entirely consistent with everything that already exists, as whenever functions in `Base` are defined this is also sometimes shortened (i.e. `prod`, ...). 

Otherwise, let me focus on the simple update code first. (It might even be convenient to split this PR into two separate PRs, for SU and FU separately).
I think the it makes sense to combine the peps tensors and weights into a single struct, perhaps `InfiniteWeightPEPS` could make sense? Considering that the state itself really consists of both, I think it is better to keep this data together, as the PEPS tensors don't really make up a state anymore without the weights, so keeping them in `InfinitePEPS` might be slightly misleading. This would also make it somewhat more explicit to absorb and split off weights again, as you can convert between the types that way.

Code-organisation wise, it would be more in line with the other algorithms to define an algorithm struct that holds the settings:
```julia
struct SimpleUpdate
    dt
    trscheme
    tol
    maxiter
    ...
end
const SU = SimpleUpdate
```
which could then be used in perhaps `fixedpoint` to have an equivalent interface as the gradient-based methods.

For the expectation values, you could leverage the existing functionality by first absorbing the weights, constructing a regular `InfinitePEPS`, and then making use of the `contract_localoperator` function defined [here](https://github.com/QuantumKitHub/PEPSKit.jl/blob/90375ca9b34b8c3fc2c7e3550cbdfb033163ccff/src/algorithms/contractions/localoperator.jl#L16). That function combines constructing the reduced density matrices and applying the operators, hopefully to arrive at a better contraction order.

This is also how the models that are currently in PEPSKit are defined, as a list of local operators (coordinates => tensors). In that way, you should be able to extract the "gates" from these structures as the elements of that list.

If you prefer, I can definitely have a look and bring your implementation a bit in line with the rest. I can implement some changes and suggest them to you via a different PR, so you could have a look and let me know what you think?

test/heisenberg_sufu.jl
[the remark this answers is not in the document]
lkdvos: I don't think there is really an obvious time limit. For these PEPS things, it's really tricky to find things that are representative but still reasonably quick... Maybe by tweaking the tolerances we can make it run a little faster, as it doesn't need to be so precise?

src/states/suweight.jl
lkdvos: Sounds great to me, I mostly care about bundling it with the state, less about the names and structures of the fields.

PR #94, opened 2024-11-14
Use `reallinsolve` for solving the CTMRG gradient linear problem

pbrehmer:
This PR replaces `linsolve` with `reallinsolve` for the CTMRG fixed-point gradient linear problem, as implemented and explained in this [recent PR](https://github.com/Jutho/KrylovKit.jl/pull/105). Since the Jacobians $\partial f / \partial A$ and $\partial f/\partial x$ are not complex linear but only real linear operators (see added test), using `linsolve` directly is not valid in principle.


lkdvos:
`boundarymps` errors are unrelated and should be fixed by https://github.com/QuantumKitHub/PEPSKit.jl/pull/95

pbrehmer:
This PR should be essentially ready to go. It only switches to `reallinsolve` and adds a linearity test for the CTMRG Jacobians. However, we need to wait until KrylovKit tags a new release because as of now, the `reallinsolve` functionality is only available on the master branch.

One thing I'm confused about is that using `FullInfiniteProjector` the Jacobians somehow stop being properly linear even for real coefficients. (See the `jacobian_real_linear.jl` test.) Somehow the full-infinite projector algorithm really doesn't differentiate too well - at least in its current implementation.

lkdvos (approved):

PR #97, opened 2024-11-18
Add simple update algorithm

Yue-Zhengyuan:
This PR adds the core part of the simple update ([arXiv 0806.3719](https://arxiv.org/abs/0806.3719)) algorithm. 

Main changes:

- Added a new struct `InfiniteWeightPEPS`, which is the collection of tensors on lattice vertices and bond weights. The weights are diagonal Schmidt matrices on the bonds of the iPEPS, and serve as a crude approximation of the environment of each site tensor. They are wrapped in the new struct `SUWeight`. 
- Picked out a sub-function `ctmrg_leftmove` from `ctmrg_iter`, which only perform the left move for one column (of the sequential CTMRG (#90)). It will be used in the full update (to be added later). 
- Added some functions to perform rotation and reflection of the iPEPS, the weights and the `LocalOperator`s.
- Added the Hamiltonian of the Hubbard model (#80) and the t-J model in `src/operators/models.jl`.
- Added simple update test in `test/heisenberg.jl`, and an example (by @Sander-De-Meyer) on Hubbard model at half-filling at `examples/hubbard_su.jl`. 

Limitations:

- For convenience, I made some requirements on whether certain virtual spaces of iPEPS tensors are dual spaces. Hopefully they can be relaxed in the future. 
- The current implementation can currently only calculate models with nearest neighbor interaction only. Actually SU can handle next nearest neighbor terms (see [arXiv 1908.09359](https://arxiv.org/abs/1908.09359)), but that will be implemented in the future. 

Things to do before merging:

- [x] Improve docstring.
- [x] ~Add more constructors for `InfiniteWeightPEPS`~ (postponed, when there is absolute need).
- [x] Integrate with Hamiltonians in `LocalOperator` format.
- [x] Use Julia's more advanced logging system instead of simply printing the information.


src/states/infiniteweightpeps.jl
```diff
@@ -0,0 +1,211 @@
...
+"""
+Schmidt bond weight used in simple/cluster update
+"""
+struct SUWeight{E<:PEPSWeight}
+    x::Matrix{E}
+    y::Matrix{E}
+end
+
+function Base.size(wts::SUWeight)
+    @assert size(wts.x) == size(wts.y)
```
lkdvos, line 20: I think if you move this to check to the constructor, it is fair to remove it here.

src/states/infiniteweightpeps.jl
```diff
@@ -0,0 +1,211 @@
...
+    return size(wts.x)
+end
+
+function Base.eltype(wts::SUWeight)
+    @assert eltype(wts.x) == eltype(wts.y)
```
lkdvos, line 25: This is guaranteed by the type signature of `SUWeight`

src/states/infiniteweightpeps.jl
```diff
@@ -0,0 +1,211 @@
...
+function Base.iterate(wts::SUWeight, state=1)
+    nx = prod(size(wts.x))
+    if 1 <= state <= nx
+        return wts.x[state], state + 1
+    elseif nx + 1 <= state <= 2 * nx
+        return wts.y[state - nx], state + 1
+    else
+        return nothing
+    end
+end
```
lkdvos, line 61:
```suggestion
function Base.iterate(wts::SUWeight, state...)
    return iterate(Iterators.flatten((wts.x, wts.y), state...)
end
```

src/states/infiniteweightpeps.jl
```diff
@@ -0,0 +1,211 @@
...
+
+function Base.length(wts::SUWeight)
+    @assert size(wts.x) == size(wts.y)
```
lkdvos, line 64: similar comment about size in constructor

src/states/infiniteweightpeps.jl
```diff
@@ -0,0 +1,211 @@
...
+    return 2 * prod(size(wts.x))
```
lkdvos, line 65: This is slightly strange, in the sense that `length` here is the iteration length, while `size` is the size of the unit cell. Do you ever really use the iteration functionality? I would think it might be more convenient to have this behave as a "matrix" than as an iterator

src/states/infiniteweightpeps.jl
```diff
@@ -0,0 +1,211 @@
...
+                ↓
+                y[r,c]
+                ↓
+    ←x[r,c-1] ← T[r,c] ← x[r,c] ←
+                ↓
+                y[r+1,c]
+                ↓
+```
+"""
+function absorb_wt(
```
lkdvos, line 146:
```suggestion
function absorb_weight(
```

I find that `wt` is not very clear as an abbreviation, especially in function names and kwargs. Would you consider using `weight` instead? (I care less about this for internal variable names, although typically it is slightly more readable to simply write out names)

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -0,0 +1,207 @@
...
+            ↑           ↑
+            |----gate---|
+            ↑           ↑
+            1           2
+            ↑           ↑
+        -1← aR -← 3 -← bL ← -4
+    =#
+    @tensor tmp[:] := gate[-2, -3, 1, 2] * aR[-1, 1, 3] * bL[3, 2, -4]
+    # SVD
+    aR, s, bL, ϵ = tsvd(tmp, ((1, 2), (3, 4)); trunc=truncation_scheme(alg, space(T1, 3)))
```
lkdvos, line 79: Here, you can save a permutation/allocation by writing the output indices explicitly:
```suggestion
    @tensor tmp[-1 -2; -4 -3] := gate[-2, -3, 1, 2] * aR[-1, 1, 3] * bL[3, 2, -4]
    # SVD
    aR, s, bL, ϵ = tsvd!(tmp; trunc=truncation_scheme(alg, space(T1, 3)))
```

src/algorithms/time_evolution/gatetools.jl
lkdvos, line 1: I think it might make more sense to move the rotation functions etc to the `localoperator` file instead

src/utility/util.jl
```diff
@@ -21,6 +21,17 @@ function _elementwise_mult(a::AbstractTensorMap, b::AbstractTensorMap)
     return dst
 end
 
+"""
+Compute S^(pow) for diagonal matrices `S`
+"""
+function sdiag_pow(S::AbstractTensorMap, pow::Real)
```
lkdvos, line 27: Could this function somehow be merged with the one below?

lkdvos (commented):
I tried to go through as much as possible, apologies for the delay.

I would say in general this is looking almost ready to go.

One general comment I have is that we typically tend to put the function signatures in the docstrings, which can be quite convenient (I have a tendency to forget this after a while...).

For the tests, I am not entirely satisfied with the current structure, which seems rather focused on "examples" instead of actual tests. I think the bar could be way lower: testing if the energy is decreasing after a number of iterations might already prove that something is working, or checking that two half steps are approximately equal to one full step could also maybe work?

For the actual examples: I would really advice to merge all heisenberg tests in a single file, where you just add an initial phase where an approximate groundstate is found with simple update, and a similar set of tests for the other examples. Keep in mind that this is not meant to represent a full simulation, rather just check that everything is behaving as expected, so the energy should be somewhat correct but doesn't need more than 1 or 2 digits of accuracy, etc. This should also mean that we could get rid of the utility heisenberg file in the tests?

src/states/infiniteweightpeps.jl
```diff
@@ -0,0 +1,242 @@
...
+function Base.isapprox(wts1::SUWeight, wts2::SUWeight; atol=0.0, rtol=1e-5)
+    return (
+        isapprox(wts1.x, wts2.x; atol=atol, rtol=rtol) &&
+        isapprox(wts1.y, wts2.y; atol=atol, rtol=rtol)
+    )
+end
+
+function compare_weights(wts1::SUWeight, wts2::SUWeight)
+    wtdiff = sum(_singular_value_distance((wt1, wt2)) for (wt1, wt2) in zip(wts1, wts2))
+    return wtdiff / (2 * prod(size(wts1)))
```
lkdvos, line 78: Here you could rewrite this to not use iteration instead, which gives the same result. I'm not entirely convinced that iterating over weights is an intuitive operation, so I would prefer to have `length` return the indexing length, ie the sites in the unitcell
```suggestion
    (L = length(wts1)) == length(wts2) || throw(ArgumentError("size mismatch"))
    return (sum(_singular_value_distance, zip(wt1.x, wt2.x)) + sum(_singular_value_distance, zip(wt1.y, wt2.y))) / (2L)
```

src/states/infiniteweightpeps.jl
lkdvos: I really think that iteration is not the right abstraction for that. It sounds like these things really are mapping over all weights, while iteration is typically used to visit only a few, and to define an order. I don't really see that iterating first the `x` weights and then the `y` weights would constitute a natural order, so I would refrain from defining that.

src/utility/util.jl
lkdvos: I think it might be reasonable to have `tol = pow > 0 ? zero(real(scalartype(S))) : _other_tol)`

lkdvos:
Thanks for explaining that detail about the unit cell, I indeed missed that. I would still prefer to have a single testfile for the heisenberg model, so maybe you could still just merge the two files together, and copy the general structure of that test, but with a larger unit cell?

src/states/infiniteweightpeps.jl
lkdvos: The main problem is that `size` and `length` are incompatible, which is a bit surprising to a user (and might lead to strange bugs when combining broadcasting/iterating/map/.... 

One alternative solution could be that `size = (2, rows, cols)`, to indicate that you have `dir, row, col` for the weights. (you could even define `getindex` and `setindex!` to reflect that behavior). That way, this ambiguity is resolved, and you still have iteration, which I would probably define as:
```julia
Base.iterate(wt, state=0) = iseven(state) ? wt.x[div(state, 2)] : wt.y[div(state, 2)]
```
Which would then return the x and y weights interleaved, but that is more consistent with the `dir, row, col` convention we have throughout PEPSKit.

lkdvos:
If the tests pass right now, we have a bit of a problem in the sense that our tests fail if multithreading is enabled. (This should then be unrelated to this PR, as I accidentally copied over `nthreads = 4` when trying to fix the tests).

pbrehmer:
Wait, wasn't there previously an issue with multi-threading causing the tests to fail (only on CI) because of some issues with GitHub actions? On #99 multi-threading still works fine on the old CI configuration

lkdvos:
Actually, the multithreading in the CI seems to have gotten disabled here: https://github.com/QuantumKitHub/PEPSKit.jl/commit/3a9eb404b9ce4de871654904850968c1f50e2cc0 (the default is just 1 thread on CI)

I fixed the complaints about the required checks, this was a naming thing.

pbrehmer:
Whoops, okay I wasn't aware that multi-threading was disabled all along. But I agree, that's a separate issue we should fix in a separate PR.

Other than that the PR looks pretty good, it's quite a big step forward! I will quickly add the function signatures in the docstrings.

pbrehmer:
I added the function signatures as well as some cosmetic changes, and I merged the Heisenberg tests. This is good to go for me once the tests turn green.

Big big thanks to @Yue-Zhengyuan for the great effort!

lkdvos:
I checked the stacktraces of the failing tests, and I think the issue is a combination of some tolerances and a bug in KrylovKit:
https://github.com/Jutho/KrylovKit.jl/issues/110

The eigensolver in the rrule is failing to converge all requested values, and as a result there is an out of bounds error due to the aforementioned bug in KrylovKit. I think we should be able to avoid running into this altogether, by ensuring that the tolerances are set such that everything converges?

pbrehmer:
> By the way, I'm not quite familiar with defining the rrule. Could you please also take a look at the updated rrule for sdiag_pow and see if it's correct?

It's almost correct - yesterday I missed a tiny error. When deriving the adjoint for complex variables, one also needs to take the complex conjugate corresponding to `spow2'`. Apart from that it looks good! (I will push that now as I'm anyway dealing with the `eltype` thing.)

>... I notice that the definition of eltype is not quite consistent in PEPSKit.

Yes, you're right. However, it's not a big issue since the more relevant way to access the scalar element type is using `scalartype` as defined in [VectorInterface](https://github.com/Jutho/VectorInterface.jl). The definitions for `scalartype` are consistent for `InfinitePEPS` and `CTMRGEnv`. (The difference between the two essentially comes down to the fact that `eltype` is ambiguous for nested structs, whereas `scalartype` clearly returns the underlying scalar element type of the struct.)

Nonetheless, it makes sense to define `eltype` consistently, even though it is not needed much.

pbrehmer:
Well, my feeling is that `InfiniteWeightPEPS` are not really vectors since the elements are defined on different fields? So I don't think it can reasonably support the VectorInterface. Also, I don' think we really need it, since we don't do vector operations with that struct, right?

Regarding `similar(peps, args...)` I think it is fine the way it is. For example, if you have a complex PEPS and want to create a similar real PEPS you can just use `similar(peps, Float64)`.

pbrehmer:
Ahh thanks for catching that, I didn't realize `similar` was defined differently for PEPOs.

Regarding the `args...` of `similar` which should be allowed: I think it is okay if we leave it up to the user which arguments are passed onto the `similar` call which acts on the underlying `TensorMap`s. For example, if the user inputs a domain and codomain such that the tensors are not PEPS/PEPO tensors anymore, the `similar` call will error.

lkdvos:
Actually, the fix for eltype should be the other way around: that function is typically reserved for the type of the result of indexing or iterating, which is why it affects similar. At some point we realized that we need to also have access to the number type, so we introduced scalartype, but apparently the CTMRGEnv didn't get updated. In principle, I'd say it isn't even very well-defined for that, since it could be either an edge or a corner, but for the peps and pepo it really is

pbrehmer:
In that case I would suggest just scrapping `eltype` for `CTMRGEnv`s, since we don't use that anyway and perhaps there is no reasonable and consistent way of defining it for `CTMRGEnv`s. Probably, for all cases where you would want `eltype`, you could just use `scalartype`.

pbrehmer:
So the Heisenberg test does optimize if the derivative of the SVD is performed using `GMRES`. However, you do see that there are many gauge sensitivity warnings and also warnings about (Heisenberg SU(2)?) degeneracies - as expected in Jutho/KrylovKit.jl#110.

This circumvents `Arnoldi` failing to converge all values but of course the underlying problem is still there, which leaves the question of how in general such cases should be handled. Also, in principle we would need Lorentzian broadening here (it's still unclear how to implement that together with the KrylovKit adjoints).

lkdvos (approved):
Good to for me too!

Issue #100, opened 2024-11-29
PEPSOptimize fails for real tensors


Yue-Zhengyuan:
MWE:
```julia
using Random
using PEPSKit
using TensorKit
using OptimKit
using KrylovKit

Dcut, χenv = 4, 16
N1, N2 = 2, 2
Random.seed!(0)
eltype = Float64
k = ℂ
peps = InfinitePEPS(rand, eltype, k^2, k^Dcut; unitcell=(N1, N2))
lattice = InfiniteSquare(N1, N2)
ham = heisenberg_XYZ(lattice; Jx=1.0, Jy=1.0, Jz=1.0)
if eltype <: Real
    ham = LocalOperator(ham.lattice, Tuple(ind => real(op) for (ind, op) in ham.terms)...)
end
envs = CTMRGEnv(rand, eltype, peps, k^χenv)
ctm_alg = CTMRG()
opt_alg = PEPSOptimize(;
    boundary_alg=ctm_alg, optimizer=LBFGS(4; gradtol=1e-3, verbosity=2)
)
result = fixedpoint(peps, ham, opt_alg, envs) # error in this line
```

Error message:
```
LoadError: MethodError: no method matching update!(::CTMRGEnv{TrivialTensorMap{ComplexSpace, 1, 1, Matrix{Float64}}, TrivialTensorMap{ComplexSpace, 3, 1, Matrix{Float64}}}, ::CTMRGEnv{TrivialTensorMap{ComplexSpace, 1, 1, Matrix{ComplexF64}}, TrivialTensorMap{ComplexSpace, 3, 1, Matrix{ComplexF64}}})
The function `update!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  update!(::CTMRGEnv{C, T}, ::CTMRGEnv{C, T}) where {C, T}
   @ PEPSKit ~/GitHub/PEPSKit.jl/src/environments/ctmrg_environments.jl:413

Stacktrace:
  [1] #381
    @ ~/GitHub/PEPSKit.jl/src/algorithms/peps_opt.jl:172 [inlined]
  [2] ignore_derivatives
    @ ~/.julia/packages/ChainRulesCore/6Pucz/src/ignore_derivatives.jl:26 [inlined]
  [3] rrule
    @ ~/.julia/packages/ChainRulesCore/6Pucz/src/ignore_derivatives.jl:40 [inlined]
  [4] rrule
[...]

PR #102, opened 2024-12-03
Warn in case of real environments in `:fixed` mode

pbrehmer:
Catches the issue with real-valued environments and the `fixed` optimization mode, detected in #100. Thanks to @Yue-Zhengyuan for finding this!

### DIFF
--- a/src/algorithms/peps_opt.jl
+++ b/src/algorithms/peps_opt.jl
@@ -145,10 +145,10 @@ function fixedpoint(
     ψ₀::InfinitePEPS{F},
     H,
     alg::PEPSOptimize,
-    env₀::CTMRGEnv{C,T}=CTMRGEnv(ψ₀, field(F)^20);
+    env₀::CTMRGEnv=CTMRGEnv(ψ₀, field(F)^20);
     (finalize!)=OptimKit._finalize!,
     symmetrization=nothing,
-) where {F,C,T}
+) where {F}
     if isnothing(symmetrization)
         retract = peps_retract
     else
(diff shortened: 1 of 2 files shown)


src/algorithms/peps_opt.jl
```diff
@@ -157,6 +157,13 @@ function fixedpoint(
         finalize! = (x, f, g, numiter) -> fin!(symm_finalize!(x, f, g, numiter)..., numiter)
     end
 
+    if scalartype(C) <: Real || scalartype(T) <: Real
```
lkdvos, line 160: I think you can simply call `scalartype(envs)` to avoid having the where clauses.

src/algorithms/peps_opt.jl
```diff
@@ -157,6 +157,13 @@ function fixedpoint(
...
+        env₀ = CTMRGEnv(complex.(env₀.corners), complex.(env₀.edges))
```
lkdvos, line 161:
```suggestion
        env₀ = complex(env₀)
```
Probably cleaner to just define `complex` for the environments

src/algorithms/peps_opt.jl
pbrehmer: Of course I forgot that `CTMRGEnv` supports VectorInterface lol

src/algorithms/peps_opt.jl
pbrehmer: Yes, agree. I also defined `real` for future use

pbrehmer:
I still need to add the big full-infinite environment contractions, then I'll merge the CTMRG refactor once that's pushed and reviewed.

lkdvos (approved):

PR #106, opened 2024-12-14
Add full update algorithm (attempt 1)

Yue-Zhengyuan:
This PR adds the core part of the fast full update ([arXiv 1503.05345](https://arxiv.org/abs/1503.05345)) algorithm. Compared to simple update, the environment around a bond to be updated is more accurately represented by the CTM tensors, instead of diagonal weight matrices. 

Limitations:

- As in simple update, I made some requirements on whether certain virtual spaces of iPEPS tensors are dual spaces. 
- The current implementation can currently only calculate models with nearest neighbor interaction only. Support for next nearest neighbor terms (see [arXiv 1711.07584](https://arxiv.org/abs/1711.07584)) will not be implemented in near future. 

Things to do before merging:

- [x] *Find a robust criterion to change the evolution time step and determine convergence*. ~Currently, the FU stops when the energy is no longer decreasing. However, I feel that this may prevent the algorithm to evolve out from a local minimum. It may be better if some singular value spectrum can be used to determine convergence (as in CTMRG and simple update).~ From my experience using the energy as criterion is good enough for most cases.
- [x] ~Add support for bipartite iPEPS (currently for 2 x 2 unit cell, the algorithm cannot keep the bipartite structure).~ This is just too much work; the CTMRG algorithm also needs to be modified.
- [x] Add a non-trivial test or example. 
- [ ] Improve docstring.
- [ ] Use Julia's more advanced logging system.


lkdvos:
It might already be possible to achieve something like this by simply restricting the number of iterations in the CTMRG contraction. This is definitely not fully equivalent, so I agree that it would be beneficial to have an abstraction to allow different algorithms there.

Yue-Zhengyuan:
The following is the best I can do now for the t-J model: I compare the energy for D = 12 states at t/J = 2.5 with Corboz's results (The energy per hole is calculated using the formula $E_h = (E_s - E_0) / \delta$, where $E_s$ is the energy per site, and $E_0$ is the value of $E_s$ at half-filling (using Monte Carlo value for Heisenberg model)

ehole_D12-vs-Corboz

esite_D12-vs-Corboz

The energy per site (in units of J) difference between my and Corboz's results at doping = 0.1 is about 0.002. Although I only use χ = 100 (it's already quite slow on CPU, taking about 15 minutes per FU step even with spin U(1) symmetry), I feel that the difference mainly comes from the initialization: the singlet pairing $\Delta$ of nearest neighbor bonds in the d-wave states of Corboz (see figure below) is less uniform than mine

Screenshot 2025-03-28 at 10 47 35

We know that around 0.1 doping there's strong competing stripe order, so the non-uniformness may help lower the energy. If I can get Corboz's initialization, I think I can also obtain a better energy. Now I'm somewhat confident about my FU implementation, and can gradually add it to PEPSKit. I shall prioritize this over NTU. 

(Anyway, I think even Corboz himself hardly uses FU now... it's either cluster update (which I'm also going to do) or AD)

Yue-Zhengyuan:
To be re-created later when FU is really ready.

PR #110, opened 2024-12-21
Fix faulty CTMRG contractions

pbrehmer:
Small PR which fixes a wrong `conj` and index mismatch in `ctmrg_contractions.jl`. Thanks to @Yue-Zhengyuan and @Sander-De-Meyer for bringing this up!


lkdvos (dismissed):

pbrehmer:
You're right, thanks for catching that. By the way, one of the next things I'll do in the code is a major overhaul of the documentation. I will try to improve (correct) all docstrings and write up proper examples showcasing all features of the package. So if you have suggestions on the docs, let me know - I'll tag you, once I open up the PR!

lkdvos (approved):

PR #111, opened 2025-01-03
Add CTMRG support for regular 2D partition functions

leburgel:
Rebased from @Sander-De-Meyer's initial work [here](https://github.com/sanderdemeyer/PEPSKit.jl/tree/PartitionFunction).

Some todos left:
- [x] Agree on struct names
- [x] Change indexing convention for local partition function tensor to `W ⊗ S ← N ⊗ E` and document
- [x] Reorganize partition function code a bit (i.e. try not to mix in with PEPS code in the same files if possible)


src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -1,4 +1,4 @@
-const CTMRGEdgeTensor{S} = AbstractTensorMap{S,3,1}
+const CTMRGEdgeTensor{S,N} = AbstractTensorMap{S,N,1}
```
lkdvos, line 1:
```suggestion
const CTMRGEdgeTensor{S,N} = AbstractTensorMap{S,N,1}
const CTMRG_PEPS_EdgeTensor{S} = CTMRGEdgeTensor{S,3}
const CTMRG_PF_EdgeTensor{S} = CTMRGEdgeTensor{S,2}
```
I think it could be nice to have these (unexported) type aliases in the function signatures.
The naming is annoying, as just gluing everything together is ridiculously ugly. I would normally not capitalize acronyms, and use capitalization conventions of CamelCasing etc based on the type of variable name, but it might be too late for that now. `CTMRG_EdgeTensor` and `CTMRG_CornerTensor` might also be reasonable?

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -43,6 +43,42 @@ function enlarge_northwest_corner(
...
+coordinates, environments and state, or by directly providing the tensors.
+
+```
+    C_northwest -- E_north --
+         |            |
+      E_west    -- ket-bra --
+         |            |
+```
+"""
+
```
lkdvos, line 60: Can we join the signatures in a single docstring? (picture also needs update)

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -121,6 +193,42 @@ function enlarge_southeast_corner(
...
+Contract the enlarged southeast corner of the CTMRG environment, either by specifying the
+coordinates, environments and state, or by directly providing the tensors.
+
+```
+          |            |
+    -- ket-bra --    E_east
+          |            |
+    -- E_south -- C_southeast
+```
+"""
```
lkdvos, line 209: same comment about single docstring

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -160,6 +268,42 @@ function enlarge_southwest_corner(
...
+Contract the enlarged southwest corner of the CTMRG environment, either by specifying the
+coordinates, environments and state, or by directly providing the tensors.
+
+```
+          |           |      
+       E_west   -- ket-bra -- 
+          |           |      
+    C_southwest -- E_south -- 
+```
+"""
```
lkdvos, line 284: same comment about single docstring

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -186,6 +330,29 @@ function left_projector(E_1, C, E_2, V, isqS, ket::PEPSTensor, bra::PEPSTensor=k
...
+
+Contract the CTMRG left projector with the higher-dimensional subspace facing to the left.
+
+```
+     C  --  E_2    -- |~~|
+     |       |       |V'| -- isqS --
+    E_1 -- ket-bra -- |~~|
+     |       |
+```
+"""
```
lkdvos, line 344: Same comment about single docstring

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -209,12 +376,36 @@ function right_projector(E_1, C, E_2, U, isqS, ket::PEPSTensor, bra::PEPSTensor=
...
+
+Contract the CTMRG right projector with the higher-dimensional subspace facing to the right.
+
+```
+               |~~| --   E_2   --  C
+    -- isqS -- |U'|      |        |
+               |~~| -- ket-bra -- E_1
+                         |        |
+```
+"""
```
lkdvos, line 390: Same comment about single docstring

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -282,7 +488,8 @@ function half_infinite_environment(
         quadrant1[χ_in D_inabove D_inbelow; χ D1 D2] *
         quadrant2[χ D1 D2; χ_out D_outabove D_outbelow]
 end
-function half_infinite_environment(
+
+function halfinfinite_environment(
```
lkdvos, line 492: Is this rename on purpose? I don't really care either way, but I think @pbrehmer recently changed this, although I might be mistaken

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -350,6 +559,97 @@ function half_infinite_environment(
         conj(E_4[χ6 D9 D10; χ_in])
 end
 
+"""
+    half_infinite_environment(quadrant1::AbstractTensorMap{S,2,2}, quadrant2::AbstractTensorMap{S,2,2})
```
lkdvos, line 563: ensure docstring and function name are the same

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -578,59 +878,283 @@ function full_infinite_environment(
...
-"""
-    renormalize_northwest_corner((row, col), enlarged_envs::CTMRGEnv, P_left, P_right)
-    renormalize_northwest_corner(quadrant::AbstractTensorMap{S,3,3}, P_left, P_right) where {S}
-    renormalize_northwest_corner(E_west, C_northwest, E_north, P_left, P_right, ket::PEPSTensor, bra::PEPSTensor=ket)
+The environment can also be contracted directly from all its constituent tensors.
 
-Apply `renormalize_corner` to the enlarged northwest corner.
-Alternatively, provide the constituent tensors and perform the complete contraction.
+```
+    C_1 --  E_2       --  E_3       -- C_2
```
lkdvos, line 926: I'd be careful with the numbering in the pictures, since this might be interpreted as indexing, which is not correct now. I think it is anyways a lot more readable if we use compass directions as subscripts

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -578,59 +878,283 @@ function full_infinite_environment(
...
 ```
+
+or contract the adjoint environment with `x`, e.g. as needed for iterative solvers.
 """
-function renormalize_northwest_corner((row, col), enlarged_envs, P_left, P_right)
-    return renormalize_northwest_corner(
-        enlarged_envs[NORTHWEST, row, col],
-        P_left[NORTH, row, col],
-        P_right[WEST, _next(row, end), col],
+function full_infinite_environment(
```
lkdvos, line 956: ensure `fullinfinite_...` or `full_infinite_...` is consistent with `halfinfinite_...`

src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -92,6 +92,31 @@ function _singular_value_distance((S₁, S₂))
...
+    S::AbstractTensorMap{E,1,1},
+    V::AbstractTensorMap{E,1,2},
+    Q::AbstractTensorMap{E,2,2},
+    Q_next::AbstractTensorMap{E,2,2},
+) where {E<:ElementarySpace}
+    isqS = sdiag_pow(S, -0.5)
+    P_left = Q_next * V' * isqS
+    P_right = isqS * U' * Q
+    return P_left, P_right
+end
```
lkdvos, line 106: This looks like it should be the same implementation for PEPS?

src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -92,6 +92,31 @@ function _singular_value_distance((S₁, S₂))
...
+    S::AbstractTensorMap{E,1,1},
+    V::AbstractTensorMap{E,1,2},
+    Q::EnlargedPartitionFunctionCorner,
+    Q_next::EnlargedPartitionFunctionCorner,
+) where {E<:ElementarySpace}
+    isqS = sdiag_pow(S, -0.5)
+    P_left = left_projector(Q.E_1, Q.C, Q.E_2, V, isqS, Q.partfunc)
+    P_right = right_projector(Q_next.E_1, Q_next.C, Q_next.E_2, U, isqS, Q_next.partfunc)
+    return P_left, P_right
+end
```
lkdvos, line 118: probably same comment here?

src/algorithms/ctmrg/sequential.jl
```diff
@@ -53,7 +53,10 @@ Compute CTMRG projectors in the `:sequential` scheme either for an entire column
 for a specific `coordinate` (where `dir=WEST` is already implied in the `:sequential` scheme).
 """
 function sequential_projectors(
-    col::Int, state::InfinitePEPS, envs::CTMRGEnv, alg::ProjectorAlgorithm
+    col::Int,
+    state::Union{InfinitePEPS,InfinitePartitionFunction},
```
lkdvos, line 57: Maybe we can explicitly call this union `InfiniteSquareNetwork`?

src/algorithms/ctmrg/sparse_environments.jl
```diff
@@ -15,13 +15,20 @@ struct EnlargedCorner{Ct,E,A,A′}
     bra::A′
 end
 
+struct EnlargedPartitionFunctionCorner{Ct,E,A}
+    C::Ct
+    E_1::E
+    E_2::E
+    partfunc::A
+end
```
lkdvos, line 23: Can we merge this with `EnlargedCorner` as follows:
```julia
struct EnlargedCorner{TC,TE,TA}
    C::TC
    E_1::TE
    E_2::TE
    A::TA
end
# todo: add correct types
EnlargedCorner(C, E1, E2, ket, bra=ket) = EnlargedCorner(C, E1, E2, (bra, ket))
EnlargedCorner(C, E1, E2, pf) = ...
```

src/algorithms/toolbox.jl
```diff
@@ -57,6 +57,46 @@ function LinearAlgebra.norm(peps::InfinitePEPS, env::CTMRGEnv)
     return total
 end
 
+function LinearAlgebra.norm(partfunc::InfinitePartitionFunction, env::CTMRGEnv)
```
lkdvos, line 60: I'm not really convinced that this should overload `norm` for this purpose, given that it's not really a state, and there is no overlap or conj etc...

src/states/abstractpeps.jl
```diff
@@ -48,6 +57,13 @@ virtual indices, conventionally ordered as: O : P ⊗ P' ← N ⊗ E ⊗ S ⊗ W
 """
 const PEPOTensor{S} = AbstractTensorMap{S,2,4} where {S<:ElementarySpace}
 
+"""
+    abstract type AbstractPartitionFunction end
+
+Abstract supertype for a 2D projected entangled-pair state.
+"""
+abstract type AbstractPartitionFunction end
+
```
lkdvos, line 66: Do we need this abstract type? What should it be used for? (docstring is probably wrong)

src/states/infinitepartitionfunction.jl
```diff
@@ -0,0 +1,240 @@
...
+function LinearAlgebra.rmul!(ψ::InfinitePartitionFunction, α::Number)
+    rmul!(ψ.A, α)
+    return ψ
+end
+
+# Used in _add during OptimKit.optimize
+function LinearAlgebra.axpy!(
+    α::Number, ψ₁::InfinitePartitionFunction, ψ₂::InfinitePartitionFunction
+)
+    ψ₂.A .+= α * ψ₁.A
```
lkdvos, line 178:
```suggestion
    axpy!.(α, ψ₁.A, ψ₂.A)
```
Otherwise, this will not actually be an inplace operation, since the rhs alreayd allocates a new matrix with new tensors, which are then added inplace to the lhs matrix (but allocating new tensors again)

src/states/infinitepartitionfunction.jl
```diff
@@ -0,0 +1,240 @@
...
+    ψ₁::InfinitePartitionFunction, ψ₂::InfinitePartitionFunction; kwargs...
+)
+    return all(zip(ψ₁.A, ψ₂.A)) do (p₁, p₂)
+        return isapprox(p₁, p₂; kwargs...)
+    end
+end
+
+# Used in _scale during OptimKit.optimize
+function LinearAlgebra.rmul!(ψ::InfinitePartitionFunction, α::Number)
+    rmul!(ψ.A, α)
```
lkdvos, line 170:
```suggestion
    rmul!.(ψ.A, α)
```
Also inplace usage of the tensors

test/ctmrg/partition_function.jl
```diff
@@ -0,0 +1,140 @@
...
+    return @autoopt @tensor env.corners[NORTHWEST, 1, 1][C_WNW; C_NNW] *
+        env.edges[NORTH, 1, 1][C_NNW D_N; C_NNE] *
+        env.corners[NORTHEAST, 1, 1][C_NNE; C_ENE] *
+        env.edges[EAST, 1, 1][C_ENE D_E; C_ESE] *
+        env.corners[SOUTHEAST, 1, 1][C_ESE; C_SSE] *
+        env.edges[SOUTH, 1, 1][C_SSE D_S; C_SSW] *
+        env.corners[SOUTHWEST, 1, 1][C_SSW; C_WSW] *
+        env.edges[WEST, 1, 1][C_WSW D_W; C_WNW] *
+        O[D_W D_S; D_N D_E]
+end
```
lkdvos, line 105: This could probably be added to PEPSKit itself, maybe something like:
`expectation_value((i, j) => O, partitionfunction, env)`?

lkdvos (commented):
Starting to look good!
I left some remarks alreayd, but some general remarks here:

For the docstrings, I don't think, it makes sense to have separate docstrings for both the partition function and PEPS (and eventually PEPO) implementations. Rather, we could use `A` in the figure, and then specify that `A` might be either bra-ket, partition function or bra-pepo-ket combinations. As a result, we should probably also agree on a style of drawings (let's do this in a separate PR) that uses the same character for drawing tensor connections, and maybe either never draw boxes around tensors or always draw boxes.
(I am assuming that you left some of the docstrings unattached to functions on purposes as this is still a WIP port?)

For naming things, `InfinitePartitionFunction` seems like the best we can do, since I really cannot come up with anything better. Given that it is rather long, how would you feel about `InfinitePF`, similar to `InfinitePEPS`? (Along with `PFTensor` etc)

leburgel:
> For the docstrings, I don't think, it makes sense to have separate docstrings for both the partition function and PEPS (and eventually PEPO) implementations. Rather, we could use `A` in the figure, and then specify that `A` might be either bra-ket, partition function or bra-pepo-ket combinations.

Sounds good, I'll to standardize as much as I can here. I agree we should be more systematic about the diagrams, I'm happy to put some effort into that soon.

> For naming things, `InfinitePartitionFunction` seems like the best we can do, since I really cannot come up with anything better. Given that it is rather long, how would you feel about `InfinitePF`, similar to `InfinitePEPS`? (Along with `PFTensor` etc)

I don't really like the 'PF' acronym, for the simple reason that no one really uses it so there's no way of knowing what it means without actually reading the docstring. This is quite different for 'PEPS' I think. I'd be happy to use the shorter names as internal aliases, but I wouldn't like to export `InfinitePF` like that. Makes sense?

src/algorithms/contractions/ctmrg_contractions.jl
leburgel: I think your suggestion is fine for internal use. I don't think switching to `CTMRG_EdgeTensor` is worth it now. I actually thinking we could probably just drop the `CTMRG` and just use `EdgeTensor` and `CornerTensor`, since this does not sound that much less clear to me and these could at some point also be used in other contraction algorithms. But we can discuss this some other time.

src/states/abstractpeps.jl
leburgel: Right now nothing, the only thing I could think of is finite partition functions and infinite weighted versions in case we want to use simple update/BP style contraction schemes. But I'm fine with removing it for now and adding in an abstract supertype only when there is need for some common interface.

The same could actually be said of `AbstractPEPS` right now. It's used as a supertype in two instances, but there's actually no abstract interface defined anywhere.

src/states/infinitepartitionfunction.jl
leburgel: Good catch. This probably means it should also be changed for `InfinitePEPS` addition, since it was just copied from there.

pbrehmer:
On the topic of naming things: I am a bit confused by calling the `AbstractTensorMap{S,2,2}` a partition function at all. Wouldn't a more generic name be `MPOTensor` (and `InfiniteMPO`)? There could probably be applications where you would want to contract or optimize a generic MPO which doesn't have to represent a partition function physically, so calling it `PartitionFunctionTensor` might be too specific. What do you think?

src/algorithms/contractions/ctmrg_contractions.jl
pbrehmer: Indeed, I thought a bit about this and found `half_infinite_environment` to be the more consistent naming convention. So perhaps we could keep it as `half_infinite_environment`? :-)

leburgel:
> On the topic of naming things: I am a bit confused by calling the `AbstractTensorMap{S,2,2}` a partition function at all. Wouldn't a more generic name be `MPOTensor` (and `InfiniteMPO`)?

I agree using 'partition function' might imply some additional physical properties rather than just a network structure, which is really only what we mean here. The reason we didn't just reuse `MPSKit.MPOTensor` and `MPSKit.InfiniteMPO` is that these are really one-dimensional structures with some additional assumptions as well.

I would personally be okay with using the term 'partition function' for things with no stat mech origin, or at least more than I would be with using 'MPO' for 2D network structures. We could also go for `SquareTensor` and `InfiniteGrid` if we want to be more generic?

src/algorithms/contractions/ctmrg_contractions.jl
leburgel: Okay, I'll replace all of the remaining `halfinfinite` occurrences with `half_infinite`.

pbrehmer:
Well, in principle partition functions could also be a 1D or 3D thing, right? So I feel a bit icky about using `PartitionFunctionTensor` in that way.

I do think it could make sense to use the `MPSKit.MPOTensor` alias because that does not yet imply a 1D structure. For `InfiniteMPO` I agree, we would need a different name (also to avoid a clash with MPSKit). What do you think about `InfiniteMPOGrid` or `InfiniteMPONetwork`?

lkdvos:
To be fair, matrix product operator really does not make too much sense either in the 2D grid case...
The only other thing I can come up with is Projected Entangled Pair Function (PEPF), which might make some sense. 
Ultimately, given that there is no term for this anywhere, we should just pick something that's practical for us though, PEPSKit.InfinitePartitionFunction does have the undertone of being 2D given that it is in PEPSKit. If you really want, we could also just add a type parameter N to give the dimensionality, which is reflected in the array type

pbrehmer:
Alright, you do have a point. Then let's just stick with `PartitionFunctionTensor` and `InfinitePartitionFunction` for practical purposes. A dimensionality type parameter `N` would make sense if we use the partition function outside of a CTMRG/boundary MPS context but to keep it consistent with `InfinitePEPS` and `InfinitePEPO` we can just keep it.

src/networks/tensors.jl
```diff
@@ -1,9 +1,56 @@
...
+           /
+          v
+         /
+  W--<-- --<--E
+       /
+      v
+     /
+    S
+```
+"""
```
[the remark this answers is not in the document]
leburgel: I agree the unicode arrows look much nicer, but I think things become a bit unclear when adding more legs. It feels to me that
```   
        P´  N
        ↑ ↙
    W →   ← E
      ↗ ↓
    S   P
```
is (while definitely prettier) a bit less clear than
```
        P´  N
        |  /
        ^ v
        |/
  W-->-- --<--E
       /|
      ^ v
     /  |
    S   P
```
If I could extend the unicode arrows with some extra lines it would definitely be the way to go, but I can't find anything suitable for the diagonal arrow...

Might just be me though. Any opinions on which is preferable @lkdvos @pbrehmer ?

src/networks/tensors.jl
```diff
@@ -1,9 +1,56 @@
...
+            N
+           /
+          v
+         /
+  W-->-- --<--E
+       /|
+      ^ v
+     /  |
+    S   P
+```
```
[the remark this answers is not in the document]
leburgel: I definitely meant to put these arrow directions for `S` and `W`. The arrows in the diagram only indicate which spaces are in the codomain (corresponding to outgoing arrows) and which are in the domain (corresponding to incoming arrows). According to our choice of tensor map space $P ← N ⊗ E ⊗ S ⊗ W$ all virtual spaces are in the domain and are therefore represented using incoming arrows.

In practice however, you are completely right we usually want to 'flip the arrow' on the south and west spaces so we can make a nice PEPS with a 1-site unit cell. However, if you pass a `S` and `W` space to the `PEPSTensor` constructor, it will put exactly those spaces in the domain. This is what is indicated by the diagram, `S` and `W` are incoming meaning they are placed in the tensor map domain. As a docstring, this is therefore the only correct depiction I think.

The way you then 'flip the arrows' is by intentionally passing **dual** spaces as `S` and `W`, so they effectively correspond to non-dual spaces with outgoing arrows. That way they actually fit on the `N` and `E` spaces of the next PEPS tensor in the state which is what you have in mind I think. However, you need to pass the dual of the spaces yourself in the constructor, which is consistent with the diagram. Note that this is exactly what is done by default in the `PEPSTensor` constructor, where `S` and `W` are chosen as the dual of `N` and `E` respectively:

https://github.com/QuantumKitHub/PEPSKit.jl/blob/5f46c16a83145bd6b0adfe88cc04857381293c9c/src/states/abstractpeps.jl#L26-L27

src/algorithms/contractions/ctmrg_contractions.jl
leburgel: While I agree, since this diagram spans multiple row-col sites properly annotating it with both direction and site indices might become a bit unwieldy. To be honest I wouldn't want to update it since I don't really know what's going on.

I merged all the docstrings so we only need to change it in one place, but since this the diagram was already like this before this PR maybe we can address this in a separate PR?

src/algorithms/ctmrg/ctmrg.jl
leburgel: Missed the fact that I should have removed this one when rebasing onto the current master I think, `build_projectors` must have changed name in the meantime and the implementation is indeed network-agnostic.

src/algorithms/ctmrg/ctmrg.jl
leburgel: Just removed this, see above.

src/algorithms/ctmrg/sequential.jl
leburgel: Done, but I wasn't too sure where to put it. It's fine for now since it's strictly less boilerplate, but I think we could do better in the future.

For one, I would prefer to refer to a pair of `InfinitePEPS` as an `InfiniteSquareNetwork` for the specific purpose here since this is the actual thing being contracted, not an individual PEPS. And then the same for bigger stacks of objects which are effective partition functions.

On the other hand, some abstract `InfiniteSquareNetwork` supertype (or just a union, I don't know) specifically for single layer networks could allow us to remove a lot of duplicates in indexing, size and linear algebra methods. There's so much just copied over from `InfinitePEPS` to `InfinitePartitionFunction` now that it makes me think we could maybe do a type hierarchy based on network geometry rather than object nature (state versus partition function versus operator).

But this is maybe something to be discussed in the future.

test/ctmrg/partition_function.jl
leburgel: I was a bit apprehensive to do this because, while it is an expectation value in the stat mech sense, it is not really the same as the `InfinitePEPS` case where you sandwich an operator between states via a physical level.

But I guess this doesn't matter that much, I'll add it and try to be more generic about the contraction generation so it can handle different kinds of expectation values.

src/algorithms/toolbox.jl
leburgel: Fair. Any other name would be fine for me as well, but I'm having a bit of trouble thinking of a suitable one. Just `value` might be the most 'correct' since the network just represents a number in the end?

src/algorithms/ctmrg/sparse_environments.jl
leburgel: Sounds good to me, I'll give it a go. I like the idea of representing local effective tensors as tuples containing stacked objects. Maybe we can adopt this practice more widely.

src/algorithms/ctmrg/sequential.jl
```diff
@@ -71,7 +71,7 @@ function sequential_projectors(
 end
 function sequential_projectors(
     coordinate::NTuple{3,Int},
-    state::InfinitePEPS,
+    state::Union{InfiniteSquareNetwork},
```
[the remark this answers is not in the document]
leburgel: Yes, good catch, thanks!

src/algorithms/ctmrg/sparse_environments.jl
leburgel: Done, and also did the same for `HalfInfiniteEnvironment` and `InfiniteEnvironment`. To keep the changes minimal I just splat the tuples into the actual contraction functions, which required me to reorder some arguments and remove some defaults. If we want we can also rewrite the actual contraction functions to take tuples for local effective MPO-like tensors, but maybe we can leave this for the future?

Still there's quite a bit of changes here. If everything is covered by the tests then it should be fine, but maybe you can have a look when you have a moment @pbrehmer if I didn't mess anything up in the sparse environments?

src/states/abstractpeps.jl
leburgel: Do you want me to remove this one then, and by extension all abstract supertypes since none of them are really used for anything?

pbrehmer: I think for now it makes sense to remove all abstract types, i.e. `AbstractPEPS`, `AbstractPEPO` and `AbstractPartitionFunction` since they don't seem to have any use. If we really need them in the future we can still think about adding them again.

src/algorithms/ctmrg/sparse_environments.jl
pbrehmer: Thanks! I would leave the local effective MPOs for the future, when we properly get function handle CTMRG (also for normal PEPS) working. I'll take a look at the sparse environments tomorrow!

src/networks/tensors.jl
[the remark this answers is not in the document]
pbrehmer: To be honest, I am fine with both diagrams styles but I do think that unicode arrows look nicer. I am just a bit worried that it starts to look cluttered when drawing larger diagrams or layers of networks (but that would probably always be the case with unicode).

In any case, I would suggest that we settle on one style and then consistently use it in all docstrings. I will open a PR on overhauling the docs in the next few days since that's been on the agenda for a while now and then we can change all docstring diagrams in there (among other things).

src/algorithms/toolbox.jl
leburgel: Well, this didn't quite work as expected since `norm` is called inside the CTMRG `leading_boundary` to give the current PEPS norm in the logging output. For one, it seems a bit wasteful to do the norm contraction every iteration regardless of the logging verbosity. It would probably be good to make the computation conditional on the logging level, no?

While I can then make the partition function tests pass by setting `verbosity=0`, this is clearly not a fix. So we should probably come up with a name that makes sense for any network type and use that in the CTMRG logging. Does `value` still make sense in the more general case?

src/networks/tensors.jl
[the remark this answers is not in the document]
lkdvos: The problem is a bit that the literature always talks about duality and arrows with respect to the codomain, ie as if the tensor is `V1 * V2 * ... <- {}`. Ultimately, it's a bit annoying that a user really has to understand these concepts quite well, and being familiar with TensorKit's way of thinking is somehow a necessity. If we can get around this, I'm definitely all for that.
I guess here the question is mainly: do we want to use arrows to denote domain-codomain or to denote the duality of the spaces. At the end of the day, the latter really does not make much sense since we never enforce this, and a user can always flip the duality without changing anything. I agree that the former can also be a bit confusing, so maybe we just drop the arrows?

src/states/abstractpeps.jl
leburgel: Done.

lkdvos:
I think my preference for the drawings is unicode lines, without arrows. In order to add the missing information about which spaces are domain-codomain, we would then have to specify this in the docs of the individual tensor types, and dedicate a page in the docs for this

src/algorithms/toolbox.jl
lkdvos: I think the cleaner way is to redefine the `ctmrg_log...` functions to take in the state instead, and specialize on what kind it is to determine what should be printed? (along with hiding this behind a verbosity check to avoid unnecessary computations)

leburgel: I agree, but to do this I would need to be able to access the current verbosity in the the `ctmrg_log...` functions inside the `withlevel` block, and I don't immediately see how to do that. I could spend some more time, but since I'm getting warnings that the verbosity macros from LoggingExtras.jl have been deprecated maybe it would be better to reconsider the logging more thoroughly at a later stage?

To patch this for now I added a network-dependent `ctmrg_objective` and made it's computation verbosity dependent. I'm very annoyed at the explicit verbosity level integers inside the `leading_boundary` body now, but I don't immediately see a way around this unless I can manage to figure out the above issue with accessing the verbosity.

Do you think this is sufficient for this PR where we address the remaining issues in a follow-up?

src/algorithms/ctmrg/sequential.jl
leburgel: Sounds good to me, I filed a dedicated issue #113 where we can discuss further.

src/networks/tensors.jl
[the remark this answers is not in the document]
leburgel: I don't think using arrows to indicate duality of spaces is a good idea here, and I agree that using arrows to denote the distinction between domain and codomain spaces is needlessly confusing for people who are not very familiar with TensorKit's ways. I'll remove the arrows and switch to using unicode box drawing characters in these docstrings.

The considerations and warnings about spaces and where they go are really only relevant for the constructors (this was the reason I added them here in the first place), not for the type docstrings themselves.

src/networks/tensors.jl
```diff
@@ -40,32 +87,37 @@ function PEPSTensor(
...
+        P´  N
+        |  /
+        ^ v
+        |/
+  W-->-- --<--E
+       /|
+      ^ v
+     /  |
+    S   P
+```
```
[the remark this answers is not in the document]
leburgel: I just removed the arrows, see also above.

src/networks/tensors.jl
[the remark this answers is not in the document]
leburgel: I removed the arrows and switched to `\diagup` for the diagonal line. I would have preferred to use the diagonal ones suggested above, but I'm having issues with rendering U+27CD and U+27CB on Ubuntu (regardless of the browser), so I went for the more basic ones.

test/ctmrg/partition_function.jl
```diff
@@ -0,0 +1,152 @@
...
```
pbrehmer, line 106: Just a small detail, but in order for `@autoopt` to work here, the index labels need to follow our convention: the environment indices have to start with `χ`.

Also, instead of defining this contraction manually, it would probably pay of to adapt `contract_localnorm` to work for `InfinitePartitionFunction`s. That would make it much easier to work with unit cells or more complicated operators.

test/ctmrg/partition_function.jl
leburgel: I added an `expectation_value` with this signature, only for local (rank-4) `O`s for now though.

test/ctmrg/partition_function.jl
leburgel: I was just in the middle of generalizing and adding this to `src`. Didn't know about the `@autoopt` convention, thanks for that!

I added a generic contraction for a single local tensor (`AbstractTensorMap{S,2,2}`) at a given partition function. Unit cells should be fine, more complicated operators not at this point. It's not immediately clear to me how to do this, since the 'operator' here is contracted into some set of virtual indices instead of being applied to physical indices. For the simplest case it's obvious what this means, but for more complicated cases you would need some conventions for the index ordering on the bigger local tensor.

pbrehmer: Well, the operator in principle is just a `InfinitePartitionFunction` again, right? Just looking at the contraction, `local_contraction` is the same as `contract_localnorm`. The only difference is the object that is contracted inside the environment. So in my head the user should construct an operator as a `InfinitePartitionFunction` and then just call `expectation_value` which only uses `contract_localnorm` under the hood.

src/algorithms/ctmrg/sparse_environments.jl
pbrehmer: Okay, I took a look and it looks good at first glance. Of course proper testing is necessary once we actually run the code. I fixed some docstring typos and formatting which I'll push once I get the rights :)

test/ctmrg/partition_function.jl
leburgel: I don't think I agree with interpreting `O` as an `InfinitePartitionFunction`, it's just a single different local tensor surrounded by some given partition function on all sides. I don't pass the surrounding partition function here since it's not needed in the contraction, but it is really inserting a single (potentially) different `O` tensor at a given site surrounded by the original partition function. In other words, the environment given is not necessarily an environment of a network of all `O`s, so I passing the operator as an `InfinitePartitionFunction` does not make sense I think. It's quite different from the local norm in spirit even though the contraction is the same. Does that make sense?

pbrehmer: Yes that does make sense to me. My point was more that given a unit cell of `AbstractTensorMap{S,2,2}` operators (which corresponds to our `InfinitePartitionFunction`) and some environment, the correct contraction to obtain the expectation value can be performed only using `contract_localnorm`. I just wanted to reuse that structure since we already have that. Making a special case for a single operator via `local_contraction` perhaps seems too specific given that we already have all this cool generic machinery that should be easy to adapt to partition functions.

Maybe we could make a method for `expectation_value` that dispatches on `Matrix{AbstractTensorMap{S,2,2}}` and even a special method that dispatches on a single `AbstractTensorMap{S,2,2}` operator and then internally use `contract_localnorm`?

pbrehmer:
The `InfiniteSquareNetwork` interface already reduces the code duplication among `InfinitePEPS`, `InfinitePEPO` and `InfinitePF` quite a bit. I'm not yet sure, however, how to best incorporate the abstract type in `sparse_environments.jl` and `ctmrg_contractions.jl`. My plan would be to reduce the redundancies by having methods which could dispatch on either `InfinitePF`, `Tuple{InfinitePEPS,InfinitePEPS}` or eventually also `Tuple{InfinitePEPS,InfinitePEPO,InfinitePEPS}`. The goal is that only the methods that really contain the `@tensor` contractions specialize on the underlying tensor type - everything else (i.e. the methods that select the right coordinates) should be generic enough to only require one method per contraction, if that makes sense...

test/ctmrg/partition_function.jl
lkdvos: Just adding on to this: replacing a single tensor is fundamentally different from `InfinitePartitionFunction`, since that one repeats the unitcell over the entire grid, while here you really want to change a single site in the entire grid.
I'm definitely happy with sharing some of the contraction code between the implementation of `contract_localnorm`, but this should probably be decoupled from the actual API functions, since the interpretation really is different.

leburgel: As far as API goes, `expectation_value` calling `contract_local_tensor` sounds fairly 'correct' to me in the current implementation, given that it's really contracting in a single different local tensor. If we want more general 'expectation values', I agree that it would be nice if we were able to reuse the implementation of `contract_localnorm`. However:
- To be able to reuse the actual code, we would need to split out the part of `contract_localnorm` that handles indices and conjugates for the edges and the 'effective MPO part' of the contraction inside the environment. This is basically everything besides the corners. While reusing the machinery in spirit is of course what we want, this will still require separate dedicated code to handle contraction expressions for different types of network stacks, right? Of course this is not a real issue at all, except that
- I really don't know how useful an `expectation_value` for `Matrix{AbstractTensorMap{S,2,2}}` actually is. Usually there's two purposes for 'local' contractions in partition functions, namely really local expectation values or correlation functions. If it's a simple local expectation value, it's a single `AbstractTensorMap{S,2,2}` being contracted which is currently covered. If it's a correlation function, this generically involves two tensor which each have an extra index being contracted across the pair (when imposing symmetries, this is almost always the case). Even for more involved local observables, this would usually involve a single higher-rank tensor (e.g. a plaquette observable) rather than a collection of `AbstractTensorMap{S,2,2}`s. None of these really require or can even be written in terms of contracting a generic `Matrix{AbstractTensorMap{S,2,2}}` inside a given environment.

In summary, while I would be fine with adding support for `expectation_value`s of `Matrix{AbstractTensorMap{S,2,2}}` and then reuse the existing machinery to implement this, I'm really not sure this is what we actually need.

pbrehmer: Alright, I wasn't sure if there is some use to the `Matrix{AbstractTensorMap{S,2,2}}` contraction since I've not yet used 2D partition function PEPS much. And of course I agree that the interpretation of inserting a single operator is different from `InfinitePF`. Then let's keep it the way it is right now.

But if I understood correctly we can't yet contract correlation functions or operator strings with `contract_local_tensor`, right? In that case we would need a version of `contract_localnorm` (with adjusted edge/MPO indices) to expand the environment in the corresponding directions.

leburgel: No, `contract_local_tensor` doesn't support correlation functions,. For that case we would definitely need the logic of `contract_localnorm`, but adapted for MPO-style indices and with an extra contraction for the additional stringy indices. All very doable, but a bit tricky so I was trying to procrastinate this until someone explicitly needed it :)

pbrehmer: Okay I'll give it a try :)

pbrehmer:
There is a design decision to be made if we want to generalize our CTMRG code to also support different ket and bras, as well as PEPOs, and I'm not sure what the best way forward is. Right now, all CTMRG contractions support different `ket` and `bra` but there is no way for the user to pass this onto the CTMRG algorithm. And in principle our code is generic enough to take any `state` (which in principle should be a `InfiniteSquareNetwork`) which is then passed onto `EnlargedCorner` and the CTMRG contractions.

The problem is that there is no `InfinitePEPSKetBra` type which contains both the ket and the bra `InfinitePEPS` which the user can pass to `leading_boundary`. Similarly, if we want to support PEPOs, we would need a `InfiniteKetPEPOBra` type which contains the ket, bra as `InfinitePEPS`s and the `InfinitePEPO`. Should we implement such types (names are up for discussion :P) and subtype as `InfiniteSquareNetwork` or is that redundant somehow? Of course this is not strictly necessary to support 2D partition functions (the code runs as is) but I thought it would pay off to invest some time into generalizing this interface while we're already at it. I hope that makes sense!

leburgel:
> There is a design decision to be made if we want to generalize our CTMRG code to also support different ket and bras, as well as PEPOs, and I'm not sure what the best way forward is. Right now, all CTMRG contractions support different `ket` and `bra` but there is no way for the user to pass this onto the CTMRG algorithm. And in principle our code is generic enough to take any `state` (which in principle should be a `InfiniteSquareNetwork`) which is then passed onto `EnlargedCorner` and the CTMRG contractions.

I think it makes sense to have all the lowest level methods for things like `leading_boundary` and `EnlargedCorner` to work on a 'fully specified' network to be contracted. By this I mean either a partition function, a real pair of ket and bra PEPS, and so on. Passing just an `InfiniteSquareNetwork` to contains extra assumptions on the other layer in the network. This is fine, but it seems natural to make this a layer around the core routines, which always take fully specified disctinct layers. That also solver the issue of not being able to pass distinct ket and bra PEPS, since you could then do this by just skipping this first layer and directly calling the core routines.

> The problem is that there is no `InfinitePEPSKetBra` type which contains both the ket and the bra `InfinitePEPS` which the user can pass to `leading_boundary`. Similarly, if we want to support PEPOs, we would need a `InfiniteKetPEPOBra` type which contains the ket, bra as `InfinitePEPS`s and the `InfinitePEPO`. Should we implement such types (names are up for discussion :P) and subtype as `InfiniteSquareNetwork` or is that redundant somehow?

I don't know if having these as actual types is necessary, it might be but it's not clear to me. Just a union type of tuples for all different possible sandwiches would do the job to start. I certainly don't think it makes sense to make these potential full sandwich types subtypes of `InfiniteSquareNetwork`, since then the hierarchy would become a bit circular I think. At least with the current meaning of `InfiniteSquareNetwork` these new types don't really naturally fall under it I think.

> Of course this is not strictly necessary to support 2D partition functions (the code runs as is) but I thought it would pay off to invest some time into generalizing this interface while we're already at it. I hope that makes sense!

I'm a bit biased of course since I want to start using this as soon as I can, but purely in terms of compartmentalizing features I think it would make sense to just restrict this PR to working partition functions. Just to keep additions from growing bigger and bigger I would break out additional discussions as issues or follow up PRs. Of course I'm not in that much of a rush, but it seems we're moving away from the initial purpose a bit already.

lkdvos:
I'm certainly not against that. I know we previously have done so, and called it [PEPSSandwich](https://github.com/quantumghent/TensorTrack/blob/main/src/mps/PepsSandwich.m), and PEPOSandwich. I do quite like those names 😉

pbrehmer:
> I don't know if having these as actual types is necessary, it might be but it's not clear to me. Just a union type of tuples for all different possible sandwiches would do the job to start. I certainly don't think it makes sense to make these potential full sandwich types subtypes of InfiniteSquareNetwork, since then the hierarchy would become a bit circular I think. At least with the current meaning of InfiniteSquareNetwork these new types don't really naturally fall under it I think.

I have also considered this but the problem with tuples as sandwiches is that you need to be able to index the sandwich. But I do agree that it doesn't make sense to subtype the sandwiches as `InfiniteSquareNetwork`s as well since it would create a circular hierarchy. Maybe it's fine to just add the corresponding indexing methods for tuples, e.g. `Tuple{InfinitePEPS,InfinitePEPS}`, but I'm not sure it will be pretty.

Anyways, I agree that we're putting too many additions into this PR and should probably generalize the CTMRG interface in another PR - I will open up an issue! I will finish up the contraction business and then the PR is good to go for me.

pbrehmer:
Alright, I'd be fine with merging this as is :-) (once the lights turn green)

leburgel:
I added an `expectation_value` method which takes a pair with a single index and a single rank-4 tensor map. While the updated syntax in the test case worked due to the added `AbstractTensorMap{S,M,M}` support, it seemed confusing to me to mix a tuple of indices with a single rank-4 tensor map. It seemed natural to have the simplest and most common case to work as intuitively as possible.

Aside from that, while the generalization of the local tensor contraction is nice, it's not covered by the tests and to be honest I can't immediately think of a relevant test withing the current test model. It's also not really clear to me what the `inds` are supposed to be in case of a single higher-rank tensor map, nor are there checks on the tensor map rank versus the grid size like there is for the matrix case.

I at least added a docstring to indicate what the current possibilities are. I guess since there's no concrete use case for the more general methods right now this is fine for the time being.

pbrehmer:
> Aside from that, while the generalization of the local tensor contraction is nice, it's not covered by the tests and to be honest I can't immediately think of a relevant test withing the current test model.

That's true and I'd like to add tests. Locally, I tested that at least the code runs and captures all the cases I could think of. To test for correctness, it would be nice to at least compare to an Ising correlation function perhaps? I'm not sure if there are exact expressions/integrals we could reference against.

> It's also not really clear to me what the inds are supposed to be in case of a single higher-rank tensor map, nor are there checks on the tensor map rank versus the grid size like there is for the matrix case.

You do need the `inds` since a higher-rank tensor map does not contain any information on how it is contracted with the surrounding environment. To make an example, a `AbstractTensorMap{S,4,4}` could be a 2x2 plaquette or it could be a 3x1 string. The first case would be specified using `inds=((1, 1), (2, 1), (1, 2), (2, 2))` whereas the second case would correspond to `inds=((1, 1), (2, 1), (3, 1))`. That also shows why a general check for the tensor map rank is not as easy as for the matrix case since the number of supplied indices could technically vary given a fixed tensor rank.

leburgel:
> Locally, I tested that at least the code runs and captures all the cases I could think of. To test for correctness, it would be nice to at least compare to an Ising correlation function perhaps? I'm not sure if there are exact expressions/integrals we could reference against.

To just check the code I think we could use a 2x2 patch with one magnetization/energy tensor and the other three just regular partition function tensors, both in matrix form and blocked to cover all cases. 

> You do need the `inds` since a higher-rank tensor map does not contain any information on how it is contracted with the surrounding environment. To make an example, a `AbstractTensorMap{S,4,4}` could be a 2x2 plaquette or it could be a 3x1 string. The first case would be specified using `inds=((1, 1), (2, 1), (1, 2), (2, 2))` whereas the second case would correspond to `inds=((1, 1), (2, 1), (3, 1))`. That also shows why a general check for the tensor map rank is not as easy as for the matrix case since the number of supplied indices could technically vary given a fixed tensor rank.

I see. Then the docstring I added should cover the proper usage I think, and maybe we could add a consistency check on the rank after determining the patch size from the given `inds`? At that point it's clear what the compatible tensor rank should be, right? At least this would be better than to let it throw an error for a badly specified contraction.

src/PEPSKit.jl
```diff
@@ -194,6 +197,7 @@ export fixedpoint
 export absorb_weight
 export su_iter, simpleupdate, SimpleUpdate
 
+export InfinitePartitionFunction, value
```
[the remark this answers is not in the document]
leburgel: I just checked that `value` didn't exist in `Base` before adding the export, but fair point. I'll drop the export.

src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -36,31 +35,36 @@ function MPSKit.leading_boundary(envinit, state, alg::CTMRGAlgorithm)
     TS = map(x -> tsvd(x)[2], envinit.edges)
 
     η = one(real(scalartype(state)))
-    N = norm(state, envinit)
     env = deepcopy(envinit)
     log = ignore_derivatives(() -> MPSKit.IterLog("CTMRG"))
+    function get_objective(state, env, verbosity)
+        return alg.verbosity >= verbosity ? ctmrg_objective(state, env) : nothing
+    end
```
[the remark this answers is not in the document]
leburgel: I thought of that, but it was not clear to me if the expression is really not evaluated if the current verbosity level is too low. Are you sure this will not just compute the objective, but then not log the actual message? I wasn't sure, and the weird thing I did was the simplest way I could think of to be sure.

src/algorithms/toolbox.jl
```diff
@@ -1,11 +1,71 @@
...
+"""
+    expectation_value(inds, O, pf::InfinitePartitionFunction, envs::CTMRGEnv)
+    expectation_value(inds => O, pf::InfinitePartitionFunction, envs::CTMRGEnv)
+
+Compute the expectation value corresponding to inserting a (set of) local tensor(s) `O` at
+position `inds` in the partition function `pf` and contracting the chole using a given CTMRG
+environment `envs`.
+
+Here `inds` can be a single index or a tuple of indices, each specified as a
+`Tuple{Int,Int}` or a `CartesianIndex{2}`. `O` can be either a single tensor or a matrix of
```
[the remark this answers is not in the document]
pbrehmer: The initial reason was that I wanted to have a way to contract any rectangular patch of operators (blocked or not). This is e.g. used to compute the local norm contraction of an `InfinitePF` (what is done through `contract_local_norm` for `InfinitePEPS`). A vector would lose the unit cell information.

While this works as is, it's of course still unclear whether this is a useful approach. Essentially, I'm not tied to this way of computing observables that act on more than one site but I'd argue it would be nice to have the functionality still in this PR. This would allow to at least compute correlation functions of partition functions.

src/algorithms/toolbox.jl
```diff
@@ -1,11 +1,71 @@
...
+tensors respectively, each specified as an `AbstractTensorMap{S,2,2}` conforming to the
+[`PartitionFunctionTensor`](@ref) indexing convention.
+
+Alternatively, `O` can be a single higher-rank tensor map, in which case it is inserted
+inside a rectangular region defined by the indices in `inds`, where in addition to the
```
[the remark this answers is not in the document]
pbrehmer: Yes, there is an ambiguity in working with higher-rank tensor maps and supplying the `inds`. The annoying thing in the current implementation is that only really the indices that specify the corners of the rectangular patch matter, so I agree that it's not a nice way of implementing multi-site operators.

Maybe to restore the original way of using the `inds` we could also do the following: Instead of supplying a higher-rank tensor map or matrix of tensor maps, we could just supply a tuple of `Pair{CartesianIndex(2),TensorMap{S,2,2}}` (similar to how a `LocalOperator` works) as well as the partition function to the `contract_local_tensor` function. Then the supplied tensors would just replace the partition function on the supplied indices. (If no tensors are supplied, then the function just computes the norm as in `contract_local_norm`.)

I don't want to further blow up the PR so I'm also happy to go for a simpler alternative. But I think the main contractions are already implemented to generalize the partition function operator evaluation to it should just be a matter of reorganizing that.

src/algorithms/toolbox.jl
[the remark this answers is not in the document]
leburgel: > A vector would lose the unit cell information.

I don't really see this. In that case there's a linear ordering where every tensor matches a corresponding index, so it should still be clear where it goes?

> While this works as is, it's of course still unclear whether this is a useful approach. Essentially, I'm not tied to this way of computing observables that act on more than one site but I'd argue it would be nice to have the functionality still in this PR. This would allow to at least compute correlation functions of partition functions.

The current implementation is still a bit broken though, I think this
https://github.com/sanderdemeyer/PEPSKit.jl/blob/8103897739009f6f74e399b74dc3f8b98897482f/src/algorithms/toolbox.jl#L48
will break if we pass a single index for a partition function with a larger unit cell. The contraction with the normal partition function patch would need a `map` over the indices to work. For the higher-rank-tensor-case this is becomes extra tricky, since the `inds` are a bit unclear there.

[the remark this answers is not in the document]
pbrehmer: Oh that's true. After all, it's still not super clear to me why we even need higher-rank tensors. Can't we just do everything using just a tuple/vector of index-operator pairs, similar to the `LocalOperator` case? That would also make it much easier to fix this.

src/algorithms/toolbox.jl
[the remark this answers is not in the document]
leburgel: > Yes, there is an ambiguity in working with higher-rank tensor maps and supplying the `inds`. The annoying thing in the current implementation is that only really the indices that specify the corners of the rectangular patch matter, so I agree that it's not a nice way of implementing multi-site operators.

For now, I would just remove the higher-rank tensor case. It's a bit random to come up with an interface for this kind of thing if we don't even have a use case in mind. Since this is making things more difficult here, I suggest we drop it altogether and think about reintroducing it once we have a specific need for it.

> Maybe to restore the original way of using the `inds` we could also do the following: Instead of supplying a higher-rank tensor map or matrix of tensor maps, we could just supply a tuple of `Pair{CartesianIndex(2),TensorMap{S,2,2}}` (similar to how a `LocalOperator` works) as well as the partition function to the `contract_local_tensor` function. Then the supplied tensors would just replace the partition function on the supplied indices. (If no tensors are supplied, then the function just computes the norm as in `contract_local_norm`.)

This seems reasonable, and also allows to solve the above issues. I don't really see the need to have a version of `contract_local_tensor` where no tensors are supplied though. It would be more transparent to just do the explicit call where you pass the partition function tensors (I assume you're referring to the denominator in `expectation_value`?). A rename to `contract_local_tensors` would then be appropriate of course...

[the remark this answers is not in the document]
pbrehmer: I would definitely agree that we should scrap the `AbstractTensorMap{S,M,M}` case since that is not nice to work with and to implement.

I could still see that the rest is salvageable :-)  I don't find the case of correlation functions too random - so if someone wants to evaluate e.g. nearest-neighbor correlators that is not super exotic, right? If you want I could still try to clean up/simplify the general contraction mechanism using `Vector{Pair,AbstractTensorMap{S,2,2}}` so that it is more in line with the `LocalOperator` way of computing the expectation value. But if we want to scrap that for now, I'd also be fine with that - so feel free to remove the relevant code bits.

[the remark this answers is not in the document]
lkdvos: In any case, this is completely decoupled from this PR, so we can separate that out to get reorganized :)

src/algorithms/ctmrg/ctmrg.jl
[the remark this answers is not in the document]
lkdvos: Double-checked to be sure, but basically the entire machinery is hidden behind a big if in the beginning to check a global debug level (not really relevant here), and then all actual function calls, string instantiations etc are hidden behind a `CoreLogging.shouldlog()`, which is what handles logging levels (both info-warn-debug and the verbosity flag.
 
```julia
julia> @macroexpand @infov 2 myexpensivecall(x)
quote
    #= /Users/ldevos/.julia/packages/LoggingExtras/cFgEq/src/verbosity.jl:11 =#
    (LoggingExtras.LoggingExtras).deprecate_verbosity(:infov)
    #= /Users/ldevos/.julia/packages/LoggingExtras/cFgEq/src/verbosity.jl:12 =#
    begin
        #= logging/logging.jl:393 =#
        let
            #= logging/logging.jl:394 =#
            var"#48#level" = Base.CoreLogging.Info
            #= logging/logging.jl:396 =#
            var"#49#std_level" = var"#48#level"
            #= logging/logging.jl:397 =#
            if var"#49#std_level" >= (Base.RefValue{LogLevel}(Debug))[]
                #= logging/logging.jl:398 =#
                var"#50#group" = LoggingExtras.Verbosity(2)
                #= logging/logging.jl:399 =#
                var"#51#_module" = Main
                #= logging/logging.jl:400 =#
                var"#52#logger" = (Base.CoreLogging.current_logger_for_env)(var"#49#std_level", var"#50#group", var"#51#_module")
                #= logging/logging.jl:401 =#
                if !(var"#52#logger" === Base.CoreLogging.nothing)
                    #= logging/logging.jl:402 =#
                    var"#53#id" = :Main_3da3c83c
                    #= logging/logging.jl:405 =#
                    if Base.CoreLogging.invokelatest(Base.CoreLogging.shouldlog, var"#52#logger", var"#48#level", var"#51#_module", var"#50#group", var"#53#id")
                        #= logging/logging.jl:406 =#
                        var"#54#file" = "REPL[2]"
                        #= logging/logging.jl:407 =#
                        if var"#54#file" isa Base.CoreLogging.String
                            #= logging/logging.jl:408 =#
                            var"#54#file" = (Base.CoreLogging.Base).fixup_stdlib_path(var"#54#file")
                        end
                        #= logging/logging.jl:410 =#
                        var"#55#line" = 1
                        #= logging/logging.jl:411 =#
                        local var"#56#msg", var"#57#kwargs"
                        #= logging/logging.jl:412 =#
                        begin
                                #= logging/logging.jl:382 =#
                                try
                                    #= logging/logging.jl:383 =#
                                    var"#56#msg" = myexpensivecall(x)
                                    #= logging/logging.jl:384 =#
                                    var"#57#kwargs" = (;)
                                    #= logging/logging.jl:385 =#
                                    true
                                catch var"#70#err"
                                    #= logging/logging.jl:387 =#
                                    Base.invokelatest(Base.CoreLogging.logging_error, var"#52#logger", var"#48#level", var"#51#_module", var"#50#group", var"#53#id", var"#54#file", var"#55#line", var"#70#err", true)
                                    #= logging/logging.jl:388 =#
                                    false
                                end
                            end && Base.CoreLogging.invokelatest(Base.CoreLogging.handle_message, var"#52#logger", var"#48#level", var"#56#msg", var"#51#_module", var"#50#group", var"#53#id", var"#54#file", var"#55#line"; var"#57#kwargs"...)
                    end
                end
            end
            #= logging/logging.jl:418 =#
            Base.CoreLogging.nothing
        end
    end
end
```

src/algorithms/toolbox.jl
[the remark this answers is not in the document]
pbrehmer: Yes let's do that. Then let's remove the multi-site operator functionality from this PR (reverting back to @leburgel original 1-site operator code) and then I will open up another PR where we can discuss this further :-) In there I can add my current idea of what would be the most organized approach with the least ambiguities.

src/algorithms/ctmrg/ctmrg.jl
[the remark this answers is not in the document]
leburgel: Okay, thanks for checking!

src/algorithms/toolbox.jl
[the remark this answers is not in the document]
lkdvos: Before you go all-in with implementing, we might want to discuss first, unless you are okay with possibly having to make large changes if we find a better approach?

lkdvos (approved):
I think we can merge this like this if the tests all pass.

src/algorithms/toolbox.jl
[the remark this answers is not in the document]
pbrehmer: > Before you go all-in with implementing, we might want to discuss first, unless you are okay with possibly having to make large changes if we find a better approach?

Sure, I definitely had discussions in mind. What I was planning could maybe serve as a good starting point for discussion/further refinements.

PR #114, opened 2025-01-10
Fix contraction order optimization for `LocalOperator` contractions

pbrehmer:


lkdvos (approved):
Good catch! I took the time to also just refactor the code a bit, there was quite a bit of duplication etc, I would argue it reads a lot nicer now.
If tests turn green, good to go for me!

pbrehmer:
The refactor looks very nice, thanks for that!

PR #123, opened 2025-02-03
Update `correlation_length` implementation

leburgel:
This updates the implementation of `correlation_length` to:
- Use only the `CTMRGEnv` edges to compute the correlation length from their MPS-like transfer matrix spectrum
- Avoids forcing edge tensor spaces to match in favor of using only standard tensor manipulations
- Allows passing additional keyword arguments through to `MPSKit.transfer_spectrum`, for example to be able to target eigenvalues in specific symmetry sectors

It also makes the function network agnostic, so we can get a correlation length from an environment of any contractible network in principle.


src/algorithms/toolbox.jl
```diff
@@ -160,39 +160,46 @@ function MPSKit.transfer_spectrum(
...
+remaining keyword arguments are passed through to `MPSKit.correlation_length` (e.g. allowing
+to target the correlation length in a specific symmetry sector).
 """
-function MPSKit.correlation_length(peps::InfinitePEPS, env::CTMRGEnv; num_vals=2)
-    T = scalartype(peps)
-    ξ_h = Vector{real(T)}(undef, size(peps, 1))
-    ξ_v = Vector{real(T)}(undef, size(peps, 2))
-    λ_h = Vector{Vector{T}}(undef, size(peps, 1))
-    λ_v = Vector{Vector{T}}(undef, size(peps, 2))
+function MPSKit.correlation_length(env::CTMRGEnv; num_vals=2, kwargs...)
```
lkdvos, line 180: I did somehow like the `correlation_length(state, [env]; kwargs...)` idea, where we could in principle compute the CTMRG environments if they are not provided. I understand that it's no longer needed here, but my head just thinks about the correlation length of a state, not of the utility partial contraction we keep cached.

src/algorithms/toolbox.jl
```diff
@@ -160,39 +160,46 @@ function MPSKit.transfer_spectrum(
 end
 
 """
-    correlation_length(peps::InfinitePEPS, env::CTMRGEnv; num_vals=2)
+Adjoint of an MPS tensor, but permutes the physical spaces back into the codomain.
+Intuitively, this conjugates a tensor and then reinterprets its 'direction' as an MPS
+tensor.
+"""
+function _dag(A::MPSKit.GenericMPSTensor{S,N}) where {S,N}
```
lkdvos, line 167: Is this to avoid the fact that MPSKit always daggers the bottom MPS? It feels like a hack, but I guess there's not really an alternative?

src/algorithms/toolbox.jl
leburgel: I agree it looks weird, but as far as I know any kind of correlation length measure for a 2D network is really computed from (just) the environment (just thinking of CTMRG and boundary MPS here, I don't really know of other ways). An immediate problem with the `correlation_length(state, [env]; kwargs...)` syntax is that you would have to pass it a contraction algorithm and some proper spaces for everything to make sense, since the correlation length you get is very dependent on the details of how the actual boundary is computed. We could put in defaults for this, but that would mean making quite a few practically arbitrary choices I think.

I also don't fully agree that the environment is really just a utility partial contraction here. Even purely from the point of view of optimization, the optimization really occurred with respect to a specific environment in a sense. And as soon as you want to do anything with the resulting state, you basically always use an explicit boundary with certain properties you chose. This, to me at least, all feels quite different from an MPS setting where the environments really are just a cache.

We can definitely try to add a `correlation_length(state, [env]; kwargs...)` method, but it's not immediately clear how to do this in a sensible way.

src/algorithms/toolbox.jl
lkdvos: Works for me, just wanted to check if I got that right.

src/algorithms/toolbox.jl
leburgel: I restored the more general signature, since it makes sense conceptually and we might need it later. I just dispatched the `env::CTMRGEnv` case to an internal method which then only uses the environment, without messing up the official  `MPSKit.correlation_length` signature.

However, the actual return arguments are still entirely specific to an MPS-style correlation length. These would definitely need to be updated to support a more general setting which is inline with the general calling signature. While I don't see a good fix for this right away, it's also not really a problem right now since we just don't have a concrete generic use case. So I think this is good enough for now as is, I hope it's a good middle ground.

lkdvos (approved):
Ok for me when tests pass. Also okay to immediately tag as v0.4.1

PR #124, opened 2025-02-03
Adapt simple update to DiagonalTensorMap

Yue-Zhengyuan:
Changes:

- Restricts some structs and functions about the singular value spectrum (most notably `SUWeight` and `sdiag_pow`) to work with `DiagonalTensorMap`s. 
- Some refactoring of simple update. 
- Added an example of Heisenberg model simple update. It will later be expanded with examples on NTU and full update.


Yue-Zhengyuan:
The failed tasks involve the backprop of `sdiag_pow` which now only works with `DiagonalTensorMap`, although the forward pass is fine. I may need help on fixing this since I'm not familiar with Zygote, ChainRules etc...

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -72,17 +65,25 @@ function _su_bondx!(
...
         -1← aR -← 3 -← bL ← -4
     =#
-    @tensor tmp[-1 -2; -3 -4] := gate[-2, -3, 1, 2] * aR[-1, 1, 3] * bL[3, 2, -4]
+    @tensor tmp[-1 -2; -3 -4] := gate[-2 -3; 1 2] * aR[-1 1 3] * bL[3 2 -4]
     # SVD
-    aR, s, bL, ϵ = tsvd!(tmp; trunc=truncation_scheme(alg, space(T1, 3)))
+    s, ϵ = nothing, nothing
+    try
+        aR, s, bL, ϵ = tsvd!(tmp; trunc=truncation_scheme(alg, space(T1, 3)))
+    catch e_lapack
```
lkdvos, line 81: This construction doesn't work, because the first attempt can already destroy the input data. This is definitely something we also noted and it's hard to come up with a good solution, but usually we choose `SVD` by default, because of its stability, and leave `SDD` to "expert users" which know how to deal with this better.

src/states/infiniteweightpeps.jl
```diff
@@ -1,11 +1,11 @@
 
 """
-    const PEPSWeight{S}
+    const PEPSWeight{T,S}
 
 Default type for PEPS bond weights with 2 virtual indices, conventionally ordered as: ``wt : WS ← EN``. 
 `WS`, `EN` denote the west/south, east/north spaces for x/y-weights on the square lattice, respectively.
 """
-const PEPSWeight{T,S} = AbstractTensorMap{T,S,1,1} where {T<:Number,S<:ElementarySpace}
+const PEPSWeight{T,S} = DiagonalTensorMap{T,S} where {T<:Number,S<:ElementarySpace}
```
lkdvos, line 8: Do we need to restrict this type here? Since `DiagonalTensorMap{T,S} <: AbstractTensorMap{T,S,1,1}`, we can simply use `DiagonalTensorMap` without restricting, which might be useful in the future.

src/states/infiniteweightpeps.jl
```diff
@@ -109,7 +109,10 @@ function InfiniteWeightPEPS(
     vertices = InfinitePEPS(f, T, Pspace, Nspace, Espace; unitcell=unitcell).A
     Nr, Nc = unitcell
     weights = collect(
-        id(d == 1 ? Espace : Nspace) for (d, r, c) in Iterators.product(1:2, 1:Nr, 1:Nc)
+        begin
+            V = (d == 1 ? Espace : Nspace)
+            DiagonalTensorMap(ones(reduceddim(V)), V)
+        end for (d, r, c) in Iterators.product(1:2, 1:Nr, 1:Nc)
     )
```
lkdvos, line 116: Isn't this a little easier to write instead as
```julia
map(Iterators.product(...)) do (d, r, c)
    ...
end
```

src/utility/util.jl
```diff
@@ -45,53 +45,49 @@ function _elementwise_mult(a₁::AbstractTensorMap, a₂::AbstractTensorMap)
...
 
-_safe_pow(a, pow, tol) = (pow < 0 && abs(a) < tol) ? zero(a) : a^pow
+_safe_pow(a::Real, pow::Real, tol::Real) = (pow < 0 && abs(a) < tol) ? zero(a) : a^pow
 
 """
-    sdiag_pow(S::AbstractTensorMap, pow::Real; tol::Real=eps(scalartype(S))^(3 / 4))
+    sdiag_pow(S::DiagonalTensorMap, pow::Real; tol::Real=eps(scalartype(S))^(3 / 4))
 
-Compute `S^pow` for diagonal matrices `S`.
+Compute `S^pow` for DiagonalTensorMap `S`.
```
lkdvos, line 53: I think we could simply add a specialization for `DiagonalTensorMap` instead of removing the old implementation.

lkdvos (commented):
The `rrule` you have looks fine, the tests are just complaining because you removed the implementations of these functions for the `TensorMap` case, which I don't think is necessary.

(In any case we still have to support `TensorMap` for now since the projector onto the diagonal type is currently not working properly, and we might lose that information. See also https://github.com/Jutho/TensorKit.jl/pull/208.

src/states/infiniteweightpeps.jl
Yue-Zhengyuan: OK, I can remove the `{T,S}` restriction.

src/algorithms/time_evolution/simpleupdate.jl
lkdvos: Let me also link this discussion here: https://github.com/QuantumKitHub/MPSKit.jl/issues/109
From what I remember, we concluded that the approach of making the copy to have the option of retrying is actually slower than simply going with the stable algorithm in the first place, so I am not a huge fan of this implementation.

If you would like to have this option however, I think a much cleaner approach is to define a new `alg` struct that implements this. That gives a more centralized implementation, and also avoids this function having the responsibility to do to many things.

src/utility/util.jl
```diff
@@ -45,53 +45,63 @@ function _elementwise_mult(a₁::AbstractTensorMap, a₂::AbstractTensorMap)
     return dst
 end
 
-_safe_pow(a, pow, tol) = (pow < 0 && abs(a) < tol) ? zero(a) : a^pow
+_safe_pow(a::Real, pow::Real, tol::Real) = (pow < 0 && abs(a) < tol) ? zero(a) : a^pow
 
 """
-    sdiag_pow(S::AbstractTensorMap, pow::Real; tol::Real=eps(scalartype(S))^(3 / 4))
+    sdiag_pow(s, pow::Real; tol::Real=eps(scalartype(s))^(3 / 4))
```
lkdvos, line 51: I'm not a huge fan of suddenly changing the name of the variable to be lowercase. Is there a compelling argument for this? Since we are using this in the context of SVD decompositions, the `S` is conventionally uppercase. Ultimately, it doesn't matter too much, but altering code like this in a PR makes it hard to distinguish the actual changes from the cosmetic ones...

Yue-Zhengyuan: I have restricted the `sdiag_pow` to `AbstractTensorMap{T,S,1,1}`, and `S` is conventionally used to denote the ElementarySpace, so I changed the singular value spectrum to lowercase `s`.

lkdvos: Thanks for explaining, that makes sense!

src/algorithms/time_evolution/simpleupdate.jl
lkdvos: What I am saying is that the cost of having to make a copy, in order to make retrying possible is, from what we experimented with before, larger than the performance gain of the fast algorithm compared to the slow. In particular, this would mean that simply using an in-place `SVD` algorithm is faster than an out-of-place retrying algorithm. (These benchmarks were not run by me though, so I'm definitely not excluding the possiblity of that still working out here)

In order to make the safer option without hardcoding that in this function, something like the following can be a good pattern:

```julia
struct RetrySVD <: TensorKit.FactorizationAlgorithm end

function TensorKit.MatrixAlgebra.svd!(A::StridedMatrix{<:BlasFloat}, alg::RetrySVD)
    try
        return TensorKit.MatrixAlgebra.svd!(copy(A), SDD())
    catch err
        err isa LAPACKException || rethrow(err)
        @warn "SDD failed"
        return TensorKit.MatrixAlgebra.svd!(A, SVD())
    end
end
```
(keep in mind I quickly wrote this down, it's untested just to give you an idea of how I might go about doing this)

src/algorithms/time_evolution/evoltools.jl
```diff
@@ -49,3 +49,20 @@ function get_gateterm(gate::LocalOperator, bond::NTuple{2,CartesianIndex{2}})
         return gate.terms[label[1]].second
     end
 end
+
+const _axlabels = Set("trbl")
```
lkdvos, line 53: I'm wondering if, for the sake of uniformity, this might be clearer to implement it through the `NORTH`, `EAST`, ... interface we already have in place. "trbl" is not the most easy to parse, while `[NORTH, EAST, SOUTH, WEST]` is slightly longer but definitely more obvious.

Yue-Zhengyuan: This is borrowed from the implementation of NTU (see [src/algorithms/time_evolution/bondenv/env_tools.jl](https://github.com/QuantumKitHub/PEPSKit.jl/pull/121/files#diff-7d2e4330e1034866cfc2441d5afa1060dfe984f527857e4ed5a91a312e9ab541) in #121) from YASTN (see the original Python code [here](https://github.com/yastn/yastn/blob/ba886e9bff6c2df72b1ae88948b1aec29981c8c1/yastn/tn/fpeps/envs/_env_auxlliary.py)). There are many functions named `edge_t`, `edge_r` etc as well. I did want to replace them with NORTH ... (at least "nesw"), but decided to make the code look similar to YASTN to make the translation easier.

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: I did some quick test with D = 10, and found that `SDD` and `SVD` do not have much performance difference here. Maybe their difference is only significant for larger tensors. I'll then use the in-place `SVD` only.

src/algorithms/time_evolution/evoltools.jl
lkdvos: It is a bit strange to mix conventions like this though, but I'm okay with leaving it for now. Just as a small aside, typically using strings is not the best for performance, since these are heap-allocated objects and the compiler cannot easily optimize over it. Typically, for patterns like this one would use either `Symbol`s, or an `Enum` type to denote all of the possibilities. There are several advantages to that, related to clarity, but also making sure that the input is what you expect it to be, since a string can really be anything. Overall it probably doesn't matter too much so if it's too much work feel free to leave it as is, I just wanted to mention this.

Also, if you copy code/translate code from other open-source software, I would feel more comfortable if you could mention that in some comments somewhere as well, ie `# Based on implementation in YASTN` or something like that. It's also a matter of giving some credit to them for their work :)

Yue-Zhengyuan: For SU I want to settle with the current status. I plan to refactor it out after the PR on NTU is good enough, and I'll surely add credit there.

src/states/infiniteweightpeps.jl
```diff
@@ -1,11 +1,11 @@
 
 """
-    const PEPSWeight{S}
+    const PEPSWeight
 
 Default type for PEPS bond weights with 2 virtual indices, conventionally ordered as: ``wt : WS ← EN``. 
 `WS`, `EN` denote the west/south, east/north spaces for x/y-weights on the square lattice, respectively.
 """
-const PEPSWeight{T,S} = AbstractTensorMap{T,S,1,1} where {T<:Number,S<:ElementarySpace}
+const PEPSWeight = DiagonalTensorMap
```
lkdvos, line 8:
```suggestion
const PEPSWeight{T,S} = AbstractTensorMap{T,S,1,1}
```
I think we can just keep this as is, while still using DiagonalTensorMap under the hood?

lkdvos (commented):
Overall, looks good to me! In principle, the examples could do with some clean up, but maybe we can leave that for when we uniformize the documentation.
(Also, if possible in the future to split of different topics into different PRs that would be great). 

After updating with the master and my final comment this should be good to go for me.

src/states/infiniteweightpeps.jl
Yue-Zhengyuan: I didn't read very carefully... personally I do want to further restrict the weights to be `DiagonalTensorMap`s, which can also help catch implementation errors. In my use cases they are always singular value spectrums. Do you have something else in mind that may one day require the bond weights to go beyond the singular values?

src/algorithms/time_evolution/simpleupdate.jl
pbrehmer: For larger tensors the performance difference can become quite drastic so it's good to keep this in mind for the future (for the CTMRG routine this made big differences).

src/states/infiniteweightpeps.jl
pbrehmer: I think there is no real downside to keeping it as an `AbstractTensorMap{T,S,1,1}`, right? The type is slightly less restrictive and would work just the same. Perhaps there can also be use cases where the PEPS weights don't come out of an SVD necessarily?

pbrehmer (approved):
Thanks for the work @Yue-Zhengyuan, this looks good! As long as we make sure to credit the YASTN code bits and adapt the naming to our internal conventions in the follow-up PR, I'm happy to merge this :)

src/states/infiniteweightpeps.jl
Yue-Zhengyuan: Well then I'll just privately make sure that the weights produced by time evolution are always DiagonalTensorMaps 😄

PR #126, opened 2025-02-03
Add eigsolve-style rrule for CTMRG fixed-point gradient

leburgel:
Solving the Sylvester problem as an eigenvalue equation, applied to the CTMRG fixed-point gradient linear problem.

Idea after:
https://github.com/Jutho/KrylovKit.jl/blob/fb56bbc7ee952bc8a6af6d278c42e78e553aa62e/ext/KrylovKitChainRulesCoreExt/eigsolve.jl#L167-L290

Actual implementation after:
https://github.com/tangwei94/VUMPSAutoDiff.jl/blob/57afc1cbcb2af9c596197c7d07adfb26ebabcd7a/src/vumps.jl#L161-L203

### DIFF
--- a/src/PEPSKit.jl
+++ b/src/PEPSKit.jl
@@ -190,7 +190,7 @@ export FixedSpaceTruncation, HalfInfiniteProjector, FullInfiniteProjector
 export LocalOperator
 export expectation_value, costfun, product_peps, correlation_length
 export leading_boundary
-export PEPSOptimize, GeomSum, ManualIter, LinSolver
+export PEPSOptimize, GeomSum, ManualIter, LinSolver, EigSolver
 export fixedpoint
 
 export absorb_weight
--- a/src/algorithms/peps_opt.jl
+++ b/src/algorithms/peps_opt.jl
@@ -77,6 +77,28 @@ function LinSolver(;
 end
 
 """
+    struct EigSolver(; solver=KrylovKit.Arnoldi(), iterscheme=Defaults.iterscheme) <: GradMode{iterscheme}
+
+Gradient mode wrapper around `KrylovKit.KrylovAlgorithm` for solving the gradient linear
+problem as an eigenvalue problem.
+
+With `iterscheme` the style of CTMRG iteration which is being differentiated can be chosen.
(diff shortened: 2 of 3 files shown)


leburgel:
I had a go at adding this after a suggestion from @Jutho, for now mainly just so I didn't forget about it again. From what I can tell solving the fixed-point gradient linear problem like this is actually very fast and stable for all the examples. It's also the method of choice for the VUMPS pullback in [VUMPSAutoDiff.jl](https://github.com/tangwei94/VUMPSAutoDiff.jl).

However, there seems to be a problem with the gradient tests for `iterscheme==:diffgauge`. One possibility is that there is actually some residual gauge freedom left which makes it such that the modified eigenvalue problem doesn't work because there's no unique eigenvector with eigenvalue 1. I certainly don't really understand what is going on so I don't know an easy fix, but though I should leave this here as a suggestion.

Do you have an idea why `diffgauge` would give so much trouble here @pbrehmer?

pbrehmer:
This is super interesting, thanks for adding this. I will try to take a closer look the next few days.

Regarding `:diffgauge`: I am kind of surprised by these issues since I thought that `:diffgauge` is in principle more stable. In that scheme, we differentiate through the gauge fixing itself so I would expect AD to take care of any residual gauge freedom - but perhaps I'm wrong.

pbrehmer (changes requested):
Thanks again for the addition, this really seems like an efficient approach to differentiating CTMRG! Perhaps this should be the new default? I will try to benchmark against `LinSolver` and see what seems best in a different PR.

leburgel:
> One can also enable `eager=true` so that the Krylov subspace is not expanded to the full specified dimension and this also seems to repair the `:diffgauge` mode and even speeds it up. Perhaps this also speeds up the `:fixed` mode.

Good catch, I had not thought of that! I added a default `gradient_eigsolve` which sets `eager=true`, and enabled eager mode in the tests using `iterscheme=:fixed` as well.

pbrehmer (approved):

lkdvos (approved):

PR #128, opened 2025-02-05
Add bond truncation algorithms for time evolution

Yue-Zhengyuan:
This PR is one of the smaller PRs that will gradually implement the NTU and FU algorithms (instead of doing it all at once). 

Here two algorithms are added for the following problem: in imaginary time evolution algorithm, after applying the 2-site gate
```
    ↑           ↑
    |----gate---|
    ↑           ↑               ↑     ↑
--- a --------- b ---   →   --- a === b ---
```
truncate the bond between the updated tensors `a` and `b` to desired dimension. They make use of a positive-definite "bond environment" tensor `benv` surrounding the tensors `a` and `b`:
```
    ┌-----------------------┐
    |   ┌----┐              |
    └---|    |-- a† == b† --┘
        |benv|   ↑     ↑         =   ⟨a,b|a,b⟩
    ┌---|    |-- a === b ---┐
    |   └----┘              |
    └-----------------------┘
```
Let the new truncated tensors be `a`, `b`. The `ALSTruncation` algorithm uses alternating least square (ALS) method to minimize the cost function
```
f(a,b)  = ‖ |a,b⟩ - |a2,b2⟩ ‖^2
```
The `FullEnvTruncation` algorithm (FET; G. Evenbly PRB 98, 085155 (2018)) iteratively maximizes the fidelity
```
F(a,b) = (⟨a,b|a2,b2⟩⟨a2,b2|a,b⟩) / (⟨a,b|a,b⟩ ⟨a2,b2|a2,b2⟩)
```
They are described in further detail in the docstring. A (theoretical) advantage of FET is that it can dynamically adjust the dimension of each symmetry sector on the bond during the optimization, while ALS is fixed by the initialization. 

A new test file `test/bondenv/optimize.jl` is added to ensure that the fidelity is non-decreasing during the optimization. 

## TODO:

- [x] (Help needed) Make sure the twists (parity tensors) required for fermions are applied correctly.
- [x] Improve log messages.


Yue-Zhengyuan:
@lkdvos @pbrehmer I think this PR is ready for review. Please take a look

src/algorithms/time_evolution/bondenv/optimize_als.jl
```diff
@@ -0,0 +1,276 @@
...
+    Vtrunc = space(s, 1)
+    a, b = absorb_s(a, s, b)
+    a, b = permute(a, (1, 2, 3)), permute(b, (1, 2, 3))
+    ab = _combine_ab(a, b)
+    # cost function will be normalized by initial value
+    cost00 = cost_function_als(benv, ab, a2b2)
+    fid = fidelity(benv, ab, a2b2)
+    cost0, fid0, Δfid = cost00, fid, 0.0
+    verbose && @info "ALS init" * _als_message(0, cost0, fid, NaN, NaN, 0.0)
+    # only optimize when fidelity differ from 1 by larger than 1e-12
```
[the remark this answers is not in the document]
Yue-Zhengyuan: I'll just remove it. In practice we almost always need to optimize.

src/algorithms/time_evolution/bondenv/optimize_als.jl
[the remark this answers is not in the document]
Yue-Zhengyuan: The names indeed come from the paper. The functions `tensor_Ra` etc are only used internally (maybe I can add a leading `_` to indicate this), so I don't bother to invent new longer names for them...

src/algorithms/time_evolution/bondenv/optimize_als.jl
```diff
@@ -0,0 +1,276 @@
...
+Solving the equations
+```
+    Ra a = Sa, Rb b = Sb
+```
+"""
+function solve_ab(
+    tR::AbstractTensorMap{T,S,2,2}, tS::AbstractTensor{T,S,3}, ab0::AbstractTensor{T,S,3}
+) where {T<:Number,S<:ElementarySpace}
+    f(x) = ncon((tR, x), ([-1 -2 1 2], [1 2 -3]))
+    ab, info = linsolve(f, tS, permute(ab0, (1, 3, 2)), 0, 1)
```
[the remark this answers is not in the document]
Yue-Zhengyuan: I would use `@tensor` instead of `*` because of fermion twist signs.

src/utility/fullenv_truncation.jl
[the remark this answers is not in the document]
Yue-Zhengyuan: `utility/fullenv_truncation.jl` handles the standard FET problem:
```
                ┌--------------------┐
                |   ┌----┐           |
                └---|    |---- b0†---┘
    maximize        |benv|              w.r.t. b0' (normalized)
                ┌---|    |---- b0'---┐
                |   └----┘           |
                └--------------------┘
```
which replaces the bond matrix `b0` with a lower-rank one `b0'` that maximizes the fidelity with the original; the bond can then be truncated by SVD of the reduced `b0'`. I think it may have applications outside the time evolution algorithms, so I put it under `utility`. 

However, in `bondenv/optimize_fet.jl`, the bond "tensor" has extra physical indices (in comparison, the ALS code is fully specialized to handle this variant):
```
┌-----------------------┐
|   ┌----┐              |
└---|    |-- a† == b† --┘
    |benv|   ↑     ↑   
┌---|    |-- a' -- b' --┐
|   └----┘              |
└-----------------------┘
```
It requires some conversion to a standard FET problem. My current approach is to first to a QR decomposition of `a` and `b` to move the physical axes to the `Q` tensors; then the bond matrix is the combination of the two `R` matrices. Frankly, in my recent test on t-J model, this conversion works well only when doping is larger than 0.1; for smaller doping the performance is much worse than ALS. I guess this is because I'm unable to optimize the `Q` tensors as well, and I'm currently thinking about a better way to do it.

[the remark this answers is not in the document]
Yue-Zhengyuan: The problem with FET is solved... it's because I didn't truncate the SVD in obtaining the initialization of the new bond matrix `u s vh`:
https://github.com/QuantumKitHub/PEPSKit.jl/blob/8e57b46a124401c96e980b660f5c07e11e1e3cd8/src/utility/fullenv_truncation.jl#L199-L200
I've changed `trunc=truncerr(1e-14)` to `trunc=alg.trscheme`. 

QR decomposition on `a`, `b` just performs a basis transform on the "bond space" and doesn't affect the result.

src/algorithms/time_evolution/bondenv/optimize_fet.jl
```diff
@@ -0,0 +1,62 @@
+"""
+    bond_optimize(benv::BondEnv{T,S}, a::AbstractTensor{T,S,3}, b::AbstractTensor{T,S,3}, alg) where {T<:Number,S<:ElementarySpace}
+
+Truncation of the bond between `a` and `b`.
```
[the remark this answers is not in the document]
Yue-Zhengyuan: I have moved `a`, `b` to the front, and updated the docstring. But honestly it's a bit verbose to explain what is the "bond environment", so maybe the new one is still not very clear. 

https://github.com/Yue-Zhengyuan/PEPSKit.jl/blob/7438cdf79fa3ec0597176b5d53617659beb22518/src/algorithms/time_evolution/bondenv/optimize.jl#L28-L41

`a`, `b` are "reduced tensors" obtained from QR decomposition from the two PEPS tensors connected by a bond. The entire PEPS $|\psi\rangle$ is then the contraction result of `a`, `b` with the rest of the tensors. Without any approximation, `benv` is the network of $\langle \psi|\psi \rangle$ but with `a`, `b` and their conjugates removed. By construction `benv` is a positive map from all indices of `a`, `b` to the indices of their conjugates. 

In practice `benv` must be approximated. One way is to use the CTM tensors. The other is to keep only a few tensors surrounding the bond (as in NTU), i.e. construct a finite state $|\psi'\rangle$, and `benv` is the network of $\langle \psi'|\psi' \rangle$ but with `a`, `b` and their conjugates removed.

The functions to construct a `benv` following these two approaches will be added in my next PR.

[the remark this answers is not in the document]
lkdvos: Thanks for the detailed explanation! At least for me, that helped follow along what is happening. I do agree that it might be a bit too much for a docstring, but given that the documentation pages are currently not very well set up or layed out, I also don't necessarily have a better solution (yet?).
I'm okay with leaving it as is for now, and rather have something that is too verbose than not verbose enough.

src/utility/fullenv_truncation.jl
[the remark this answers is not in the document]
lkdvos: I don't think I fully understood the problem, but with respect to your remark about the utility beyond time evolution, it seems that is true of both these algorithms, since they are both truncations?
My main objection is that the `utility/` folder was meant to contain small utility functions, which are typically not necessarily centered around a PEPS or something like that. As the package is growing, it does tend to accumulate more and more code, but I would argue this really is quite a substantial algorithm, and if we choose to export the function and the algorithm struct, it should be treated as such, and not be a simple `utility` function.
If you prefer this, we could also just move both of these to a new `algorithms/truncation/` folder?

src/algorithms/time_evolution/simpleupdate.jl
lkdvos, line 1: I'm not against this change of kwarg name, but I want to mention that you should be a little careful about this: every time we change the interface of an outer-facing function (such as this one since it is exported), we have to register this as a breaking change, since it breaks users code.

Typically, it is better to bundle these kinds of changes, or to be more careful about what you export, since the other functions are deemed internal.

src/algorithms/contractions/bondenv/als_solve.jl
lkdvos, line 1: One thing that I'm noticing here is that it looks like you have to go back and forth a couple times for the input tensors to permute them to be the correct shape. I'm okay with keeping this at is, but in order to minimize these operations, it might be useful to decide if there is a more convenient "shape" for each of these tensors, that has some nice properties.
For example, we pick our MPSTensor objects as {2,1} tensor maps, because that way we can immediately QR them to get to the left-canonical form. Here, it seems like there might similarly be a more convenient shape for `a`, `b` and the environments that makes sense? It's a bit hard for me to gauge this, since the code where this will be used eventually is not here yet (if I understood that correctly?). This also tends to simplify some of the contractions afterwards.

test/heisenberg.jl
lkdvos, line 1: Could you keep changes like this as separate in the future? These kind of "formatting" alterations have little meaning, and also conflate the git history: if we browse through the test/heisenberg.jl history afterwards, it will say that the last recorded change was this PR, while it has nothing to do with that. It's not that big of a deal of course, so don't worry about it for now, just wanted to mention that.

lkdvos (changes requested):
Overall looks good to me!
I think the important remarks for me are the `utility` folder not really being meant as a spot for implementations, and the mention that this is now a breaking change because of the keyword in `simpleupdate`. Again, I'm okay with making this breaking, just wanted to make sure that this is worth it.

Otherwise, could you merge the latest master, and should be good to go!

test/heisenberg.jl
Yue-Zhengyuan: I have removed the changes from this PR, and will be more careful in the future.

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: This name change is also kind of irrelevant to the purpose of this PR, so I decide to reverted to the old name `check_int` and avoid a breaking change.

src/algorithms/contractions/bondenv/als_solve.jl
Yue-Zhengyuan: I'll try to reduce the number of `permute`s, but it may take a while. I'll also move `fullenv_truncate` to `src/algorithms/time_evolution/` since currently it's not used elsewhere.

Yue-Zhengyuan: I made the following choice: the reduced tensors `a`, `b` are obtained from SVD when applying the gate. So usually I keep the axes order from SVD:
```
        2
        ↑       a[1 2; 3]
    1 -a/b- 3   b[1; 2 3]
```
But in the for-loop in optimization, I reorder the axes as
```
        3
        ↑       a/b[1 2 3;]
    1 -a/b- 2
```
This is now done before entering the for-loop, instead of every time at `_solve_ab`. Currently I can't find an order that works throughout the algorithm. 

In addition, I also changed the codomain/domain of the bond matrix `b` in `fullenv_truncate`:
```
    1 - b - 2   from [1 2;] to [1; 2]
```

src/algorithms/contractions/bondenv/benv_tools.jl
```diff
@@ -0,0 +1 @@
+const BondEnv{T,S} = AbstractTensorMap{T,S,2,2} where {T<:Number,S<:ElementarySpace}
```
Yue-Zhengyuan, line 1: I move the definition of `BondEnv` to a new file. This file will be expanded later.

src/algorithms/contractions/bondenv/als_solve.jl
lkdvos: I think that makes sense, bond "matrix" sounds to me as well like there is one incoming and one outgoing leg.

One minor detail I just noticed is that if I'm not mistaken, you are drawing your PEPS tensors upside down from our convention, is that possible? See for example [the definition of PEPSTensor](https://github.com/QuantumKitHub/PEPSKit.jl/blob/92a7711d51486ad9ee5b0b4241632d9ca8f285fc/src/networks/tensors.jl#L39)

src/algorithms/contractions/bondenv/als_solve.jl
```diff
@@ -0,0 +1,178 @@
...
+    return real(t1) + real(t2) - 2 * real(t3)
+end
+
+"""
+Solve the equations `Ra a = Sa` and `Rb b = Sb`
+"""
+function _solve_ab(
+    Rx::AbstractTensorMap{T,S,2,2}, Sx::AbstractTensor{T,S,3}, x0::AbstractTensor{T,S,3}
+) where {T<:Number,S<:ElementarySpace}
+    f(x) = (@tensor Sx2[:] := Rx[-1 -2; 1 2] * x[1 2 -3])
```
lkdvos, line 175: I think here, if you change `x::AbstractTensorMap{T,S,2,1}`, you can either simplify `f(x) = Rx * x`, or write `@tensor Sx2[-1 -2; -3] := Rx[-1 -2; 1 2] * x[1 2; -3]`.
Because this permutation has to be done anyways to multiply the tensors together, it would effectively save 2 permutations every iteration.

I would thus either suggest either making `x` like that in the first place, or simply doing one permutation before and after the `linsolve` loop.

src/algorithms/time_evolution/fullenv_truncation.jl
```diff
@@ -0,0 +1,251 @@
...
+    u, s, vh = tsvd(b0; trunc=alg.trscheme)
+    # normalize `s` (bond matrices can always be normalized)
+    s /= norm(s, Inf)
+    s0 = deepcopy(s)
+    Δfid, Δs, fid, fid0 = NaN, NaN, 0.0, 0.0
+    for iter in 1:(alg.maxiter)
+        time0 = time()
+        # update `← r -  =  ← s ← v† -`
+        @tensor r[-1 -2] := s[-1 1] * vh[1 -2]
+        @tensor p[-1 -2] := conj(u[1 -1]) * benv[1 -2; 3 4] * b0[3 4]
```
lkdvos, line 207:
```suggestion
        @plansor p[-1 -2] := conj(u[1; -1]) * benv[1 -2; 3 4] * b0[3 4]
```
This is a small trick that works for planar diagrams, and should remove the need for the `_linearmap_twist!` afterwards I think: `@plansor` works like `*` and does not put parity matrices while contracting ket with bra.

src/algorithms/time_evolution/fullenv_truncation.jl
```diff
@@ -0,0 +1,251 @@
...
+        @tensor B[-1 -2; -3 -4] := conj(u[1 -1]) * benv[1 -2; 3 -4] * u[3 -3]
```
lkdvos, line 208:
```suggestion
        @plansor B[-1 -2; -3 -4] := conj(u[1; -1]) * benv[1 -2; 3 -4] * u[3; -3]
```

src/algorithms/time_evolution/fullenv_truncation.jl
```diff
@@ -0,0 +1,251 @@
...
+        _linearmap_twist!(p)
+        _linearmap_twist!(B)
```
lkdvos, line 210:
```suggestion
```

src/algorithms/time_evolution/fullenv_truncation.jl
```diff
@@ -0,0 +1,251 @@
...
+        r, info_r = linsolve(x -> B * x, p, r, 0, 1)
```
lkdvos, line 211:
```suggestion
        r, info_r = linsolve(Base.Fix1(*, B), p, r, 0, 1)
```

src/algorithms/time_evolution/fullenv_truncation.jl
```diff
@@ -0,0 +1,251 @@
...
+        @tensor b1[-1; -2] := u[-1 1] * r[1 -2]
+        u, s, vh = tsvd(b1; trunc=alg.trscheme)
```
lkdvos, line 213:
```suggestion
        u, s, vh = tsvd!(b1; trunc=alg.trscheme)
```
I think you can safely destroy `b1` here and avoid a copy?

src/algorithms/time_evolution/fullenv_truncation.jl
```diff
@@ -0,0 +1,251 @@
...
+        s /= norm(s, Inf)
+        # update `- l ←  =  - u ← s ←`
+        @tensor l[-1 -2] := u[-1 1] * s[1 -2]
+        @tensor p[-1 -2] := conj(vh[-2 2]) * benv[-1 2; 3 4] * b0[3 4]
+        @tensor B[-1 -2; -3 -4] := conj(vh[-2 2]) * benv[-1 2; -3 4] * vh[-4 4]
+        _linearmap_twist!(p)
+        _linearmap_twist!(B)
+        l, info_l = linsolve(x -> B * x, p, l, 0, 1)
+        @tensor b1[-1; -2] := l[-1 1] * vh[1 -2]
```
lkdvos, line 222:
```suggestion
        @tensor b1[-1; -2] = l[-1 1] * vh[1 -2]
```
Here we can re-use the allocated memory of `b1` from before since these tensors are the same size I think

src/algorithms/time_evolution/fullenv_truncation.jl
lkdvos: (with the caveat that `@plansor` is strict about the semicolon in `[a b; c]`, it needs to be in the correct place)

src/algorithms/time_evolution/fullenv_truncation.jl
```diff
@@ -0,0 +1,251 @@
...
+        r, info_r = linsolve(x -> B * x, p, r, 0, 1)
+        @tensor b1[-1; -2] := u[-1 1] * r[1 -2]
+        u, s, vh = tsvd(b1; trunc=alg.trscheme)
+        s /= norm(s, Inf)
+        # update `- l ←  =  - u ← s ←`
+        @tensor l[-1 -2] := u[-1 1] * s[1 -2]
+        @tensor p[-1 -2] := conj(vh[-2 2]) * benv[-1 2; 3 4] * b0[3 4]
+        @tensor B[-1 -2; -3 -4] := conj(vh[-2 2]) * benv[-1 2; -3 4] * vh[-4 4]
+        _linearmap_twist!(p)
+        _linearmap_twist!(B)
```
lkdvos, line 220: Same comments about `@plansor` here

src/algorithms/time_evolution/optimize.jl
```diff
@@ -0,0 +1,173 @@
...
+                @info "ALS conv" * message
+            elseif cancel
+                @warn "ALS cancel" * message
+            else
+                @info "ALS iter" * message
+            end
+        end
+        converge && break
+    end
+    a, s, b = tsvd(_combine_ab(a, b); trunc=truncspace(Vtrunc))
```
lkdvos, line 126:
```suggestion
    a, s, b = tsvd!(_combine_ab(a, b); trunc=truncspace(Vtrunc))
```

test/bondenv/optimize.jl
```diff
@@ -0,0 +1,45 @@
...
+using LinearAlgebra
+using KrylovKit
+
+Random.seed!(10)
+maxiter = 500
+check_interval = 20
+trscheme = truncerr(1e-10) & truncdim(8)
+Vext = ℂ[FermionParity](0 => 100, 1 => 100)
+Vint = ℂ[FermionParity](0 => 5, 1 => 5)
+Vphy = ℂ[FermionParity](0 => 1, 1 => 2)
```
lkdvos, line 15:
```suggestion
Vext = Vect[FermionParity](0 => 100, 1 => 100)
Vint = Vect[FermionParity](0 => 5, 1 => 5)
Vphy = Vect[FermionParity](0 => 1, 1 => 2)
```
I think the other syntax is deprecated (although there might still be references to this in the docs)

lkdvos (commented):
I had some more time to go through and possibly avoid some more permutations/allocations, which might speed up the algorithm a bit and reduce some memory usage.

src/algorithms/time_evolution/fullenv_truncation.jl
Yue-Zhengyuan: I still need the fermion twists when contracting `u†`, `benv` and `b0`, and only adjust the domain/codomain of the resulting tensor. From your description I think `@plansor` is not suitable here?

src/utility/fullenv_truncation.jl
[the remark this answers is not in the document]
Yue-Zhengyuan: The problem is that the current FET implementation seems to be more prone to a bad initialization for some reason. Sometimes it will converge to a bad solution with a lower fidelity than ALS. In the t-J full update, it turns out using an un-truncated `u`, `s`, `vh` as initialization is not good. So I actually default the bond truncation method to ALS. 

Yes, both FET and ALS perform optimized truncation of a bond. However, the tensor shapes in ALS are also too specialized for time evolution, so I won't put it in a separate `truncation` folder. For now I'm also content with placing FET under `time_evolution` until it has found other applications.

src/algorithms/contractions/bondenv/als_solve.jl
Yue-Zhengyuan: Indeed... I was too accustomed to drawing ket tensors in the bottom, with physical axes pointing upwards. It just appears more natural to me that an operator acts *upon* something 😅 I can change my drawings, though

src/algorithms/time_evolution/fullenv_truncation.jl
Yue-Zhengyuan: In the following I need to calculate the fidelity between the initial `b0` and the optimized `b1`, so I can't destroy it here.

src/algorithms/time_evolution/fullenv_truncation.jl
lkdvos: I think it should all work out actually, but I should work this out properly. It's more or less based on the following intuition:
`twist` tends to show up whenever you are combining things that have to do with inner products with `@tensor`, here the U/V are unitary (orthogonal rows/columns need an inner product), and so while computing U' * something you have to add twists. For matrices (two-leg tensors), since they have total charge 0, you can actually pull this twist through, it does not matter on which leg you put this. I think you needed the twists in the first place because you are mixing `@tensor` and `*` calls, while `@plansor` writes everything in terms of `*` which tends to be what you want with inner products.

lkdvos: This being said, if the current implementation works, that's definitely also fine

src/algorithms/contractions/bondenv/als_solve.jl
lkdvos: Honestly, I think I could come up with a valid reason for any convention 😀 I'm sorry we didn't accidentally have the same one, I know this is annoying to change

src/algorithms/contractions/bondenv/als_solve.jl
Yue-Zhengyuan: > this permutation has to be done anyways to multiply the tensors together

Could you please explain more on this? Here `x` can be either `a` (`{T,S,2,1}`) or `b` (`{T,S,1,2}`). Here I already permute `a`, `b` to `{T,S,3,0}` before the linsolve for-loop:

https://github.com/QuantumKitHub/PEPSKit.jl/blob/3918cbe00bd64dd5f4d23a3c3d6613f5ba56515f/src/algorithms/time_evolution/optimize.jl#L67-L73

src/utility/fullenv_truncation.jl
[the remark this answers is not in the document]
Yue-Zhengyuan: On second thought... now I moved both to `algorithms/bond_truncation`. For ALS, if we ever need truncation on the bond between two tensors of other shapes, I guess we just "overload" the old functions.

src/algorithms/contractions/bondenv/als_solve.jl
lkdvos: Yeah sorry, this is getting a bit technical:
The way tensor contractions are implemented in general (which tends to be state-of-the-art as far as we are aware of) is by mapping them to matrix multiplication. Basically, a pairwise contraction works by TTGT (transpose-transpose-gemm-transpose): you transpose A, transpose B, reshape to a matrix and multiply, and finally transpose the result.
This means that the following are equivalent, and the former is under the hood rewritten to the latter:
```julia
@tensor Sx2[:] := Rx[-1 -2; 1 2;] * x[1 2 -3]
# long version, being explicit with ;
@tensor x2[-1 -2; -3] := x[-1 -2 -3]
@tensor Sx3[-1 -2; -3] := Rx[-1 -2; 1 2] * x2[1 2; -3]
@tensor Sx2[-1 -2 -3] := Sx3[-1 -2; -3]
```
Usually the multiplication step is dominant, but for symmetries this is only true at rather large bond dimensions, and it can significantly affect performance if you can avoid that.
Here, you can see that by solving the linear system in the `TensorMap{2,1}` setting, every function application consists only of multiplications, avoiding the permutations altogether.

lkdvos: If you want to have a look at the details, here is the inner contraction routine, which outlines the permute-permute-mul-permute approach: https://github.com/Jutho/TensorKit.jl/blob/0a66eb24427a5d86875eba8b77f297fdd4de2d9e/src/tensors/tensoroperations.jl#L294

Yue-Zhengyuan: Now I understand. I have changed both `S` and `x` to `TensorMap{2,1}`.

src/algorithms/contractions/bondenv/als_solve.jl
```diff
@@ -0,0 +1,190 @@
...
+
+"""
+Contract the axis between `a` and `b` tensors
+```
+    -- DX - a - D - b - DY --
+            ↓       ↓
+            da      db
+```
+"""
+function _combine_ab(
```
Yue-Zhengyuan, line 141: To reduce permutes when contracting `ab` with `benv`, I also changed the axis order of `ab` from `[DX da; db DY]` to `[DX DY; da db]`. But now I need to permute `ab` before and after entering the ALS for-loop when performing SVD to obtain separate `a`, `b`, like in 
https://github.com/QuantumKitHub/PEPSKit.jl/blob/4a231154a1e18ad9461b14ead0adcef9b1a5d84e/src/algorithms/bond_truncation/bond_truncation.jl#L67-L68

src/algorithms/time_evolution/fullenv_truncation.jl
Yue-Zhengyuan: The simple drop-in replacement of `@tensor` by `@plansor` (with semicolons properly added) and removal of `_linearmap_twist!` don't work. Maybe we just keep the current version.

src/utility/fullenv_truncation.jl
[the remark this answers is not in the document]
Yue-Zhengyuan: Now I added an option `trunc_init` to control whether the SVD of the new bond matrix `u s vh` are initialized with truncation or not. 
https://github.com/QuantumKitHub/PEPSKit.jl/blob/4a231154a1e18ad9461b14ead0adcef9b1a5d84e/src/algorithms/bond_truncation/fullenv_truncation.jl#L16
https://github.com/QuantumKitHub/PEPSKit.jl/blob/4a231154a1e18ad9461b14ead0adcef9b1a5d84e/src/algorithms/bond_truncation/fullenv_truncation.jl#L205
With the current choice of random seed in the test file, FET needs to be initialized without truncation:
https://github.com/QuantumKitHub/PEPSKit.jl/blob/4a231154a1e18ad9461b14ead0adcef9b1a5d84e/test/bondenv/bond_truncate.jl#L34
otherwise it will converge to a sub-optimal solution. But for the t-J model case, initial truncation is necessary. So both work only for some cases.

src/algorithms/time_evolution/fullenv_truncation.jl
lkdvos: Thanks for trying this out!

Yue-Zhengyuan:
Apart from some possible further optimizations on the usage of `permute`, I think this PR is almost done.

src/algorithms/bond_truncation/bond_truncation.jl
```diff
@@ -0,0 +1,179 @@
...
+    @assert !isdual(space(a, 2))
+    @assert !isdual(space(b, 2))
+    @assert codomain(benv) == domain(benv)
+    #= initialize bond matrix using QR as `Ra Lb`
+
+        --- a == b ---   ==>   - Qa - Ra == Rb - Qb -
+            ↓    ↓               ↓               ↓
+    =#
+    Qa, Ra = leftorth(a)
+    Qb, Rb = leftorth(b, ((3, 2), (1,)))
```
Yue-Zhengyuan, line 154: Although `rightorth` also works here and can save one `permute`, I feel more comfortable with `leftorth` by regard the bond environment `benv` as a linear operator acting on `a * b`.

lkdvos (approved):
I think this is okay to go for me. Thanks for working through the comments and feedback with me, really appreciated a lot.

@leburgel and/or @pbrehmer , do you have any remarks or can we merge this?

src/algorithms/bond_truncation/bond_truncation.jl
Yue-Zhengyuan: After some thinking I made sense of some twists required for `rightorth`; I can comfortably use `rightorth` now 😅

pbrehmer:
Looks great, I think we can merge this, thanks @Yue-Zhengyuan!

lkdvos (approved):

PR #129, opened 2025-02-11
Add PrecompileTools precompilation setup

pbrehmer:
Here we add a precompilation setup using PrecompileTools.jl to cut down on the long time-to-first-execution (TTFX) of some functions (in particular anything involving AD takes a very long time to execute). The `precompile.jl` file covers most of the common uses but doesn't iterate through all possible permutations to still ensure that the initial precompilation doesn't take too long. Currently it takes a couple of minutes.

While this really speeds up the TTFX on the user side, it might be annoying from the developer side since any changes will result in re-precompilation of the package. This can be disabled by commenting out `include("precompile.jl)` locally or as described in the [PrecompileTools docs](https://julialang.github.io/PrecompileTools.jl/stable/#Package-developers:-reducing-the-cost-of-precompilation-during-development).


pbrehmer:
In principle this already works quite well I think. The only thing that is undesirable is that now all test individually have to run through the precompilation routine. Also I will try to speed things up further...

pbrehmer:
Precompilation times now seem manageable:
- Ubuntu/Windows CI: ~10 mins
- MacOS CI: ~8 mins
- My local Ubuntu machine: ~7 mins

However, I had to take out all fermionic stuff from the workload since it more than doubles the precompilation times.

pbrehmer:
While having a PrecompileTools setup does speed up TTFX quite a bit, it's not necessarily useful to have this precompiled by default and shipped with the package itself. I will add a page on PrecompileTools when we set up the new docs so that every user can set up their own precompilation routine. Let me close this PR for now :-)

PR #130, opened 2025-02-12
Implement keyword selector

pbrehmer:
In this PR we implement a keyword selector function which parses optimization kwargs onto the corresponding algorithm structs and outputs a `PEPSOptimize` to `fixedpoint`. This simplifies the user interface since one doesn't have to create multiple nested algorithm structs.

The same could be done also for `leading_boundary` since it is another frequently-used function with many algorithmic parameters.

This idea follows KrylovKit's design which supplies a user-friendly kwarg-based method and an "expert" method. See e.g. the [`eigselector`](https://github.com/Jutho/KrylovKit.jl/blob/618457747102692a4d6070507436d03ccca2b4a3/src/eigsolve/eigsolve.jl#L226).


src/algorithms/optimization/peps_optimization.jl
lkdvos, line 1: Is there any way we can bundle the keyword arguments?

```julia
(; gradient_alg=(; tol, maxiter, verbosity), boundary_alg=(; ...))
```
seems a bit more manageable than a long list of keywords with prefixes.
I would maybe also suggest a "top-level" `tol` argument, which would set appropriate tolerances for the subalgorithms, which can then still be overwritten, ...
Similar things for the maxiter and verbosity.

pbrehmer: Sure, bundling seems like a good idea - the keyword arguments are anyway already categorized in bundles. However, this might make things harder when we want to load optimization arguments from a TOML file since the kwargs are now nested in `NamedTuple`s? We could still write some parser though, so we should be fine.

I like having a top-level `tol` argument and for `verbosity` I already actually implemented this. For `maxiter` I don't think there is any reasonable way to realize this: So for example, if I wanted only 50 optimization steps, why would that mean anything for the maximal number of CTMRG iterations or even the maximal number of Arnoldi steps in the gradient problem, etc.? Here I think the best option is just having good defaults that won't interfere with most use cases.

lkdvos: TOML actually does handle nested structures, we just have to think a bit about how we convert to and from, since I think by default that gets converted to a `Dict{String,Any}`, and we really want `Dict{Symbol,Any}`. Nothing that can't be handled I guess.

pbrehmer: What I just realized: with bundling we have the problem that the user needs to supply an entire `gradient_alg` etc. `NamedTuple` where all the corresponding kwargs have to be specified. If I just want to change e.g. the boundary tolerance I would need to specify all `boundary_alg=(; tol=1e-4, ...)` kwargs. That is quite inconvenient or am I missing something?

lkdvos:
```julia
function select_algorithm(; tol, maxiter, verbosity, sub_alg=nothing)
    actual_sub_alg = if sub_alg isa Algorithm
        sub_alg
    elseif sub_alg isa NamedTuple
        sub_tol = get(sub_alg, :tol, tol / 100)
        sub_maxiter = get(sub_alg, :maxiter, Defaults.maxiter)
    else
        throw(ArgumentError("unknown sub_alg: $sub_alg"))
    end
    ...
end
```

lkdvos: Just to give the idea of what I have in mind. Ideally, the sub algorithm also has a selector function, which you can then just pass the `sub_alg` keywords to

pbrehmer:
I will finish this up tomorrow: I want to further test around and I need to finish up the docstrings. I will then also make the tests use the kwarg-based interface since that should get rid of a few lines. Perhaps we can get this merged tomorrow :-)

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -44,19 +48,50 @@ end
...
+to the supplied optimization parameters. The initial environment `env₀` serves as an
+initial guess for the first CTMRG run. By default, a random initial environment is used.
 
-The `finalize!` kwarg can be used to insert a function call after each optimization step
-by utilizing the `finalize!` kwarg of `OptimKit.optimize`.
-The function maps `(peps, env), f, g = finalize!((peps, env), f, g, numiter)`.
-The `symmetrization` kwarg accepts `nothing` or a `SymmetrizationStyle`, in which case the
-PEPS and PEPS gradient are symmetrized after each optimization iteration. Note that this
-requires a symmmetric `peps₀` and `env₀` to converge properly.
+The optimization parameters can be supplied via the keyword arguments or directly as a
```
lkdvos, line 59: I feel like we should probably try and do something about the formatting of this. It might be just me, but this just looks like a big wall of text...

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -144,6 +181,132 @@ function fixedpoint(
...
+        gradient_alg=(;),
+        optimization_alg=(;),
+        (finalize!)=OptimKit._finalize!,
+    )
+
+Parse optimization keyword arguments on to the corresponding algorithm structs and return
+a final `PEPSOptimize` to be used in `fixedpoint`. For a description of the keyword
+arguments, see [`fixedpoint`](@ref).
+"""
+function select_fixedpoint_algorithm(
```
lkdvos, line 199: How do you feel about combining the structure:
```julia
function f(args...; kwargs...)
    alg = select_algorithm(f, args...; kwargs...)
    return f(args..., alg)
end
function select_algorithm(::typeof(f), args...; kwargs...)

end
```

lkdvos (commented):
I sadly won't have time to really look over this in detail today (and Monday is a holiday here), but I will look into it next week. Do you mind if I try out some changes and push them to your branch if I get to it?

src/algorithms/optimization/peps_optimization.jl
pbrehmer: No I agree, the docstring itself looks like a big wall of text, but I find it quite okay the way it is printed in the REPL. I'll format that on Monday :)

src/algorithms/optimization/peps_optimization.jl
pbrehmer: That seems really neat, I'll get to implementing it next week!

src/PEPSKit.jl
```diff
@@ -69,15 +68,22 @@ Module containing default algorithm parameter values and arguments.
...
 - `trscheme=FixedSpaceTruncation()`: Truncation scheme for SVDs and other decompositions
-- `fwd_alg=TensorKit.SDD()`: SVD algorithm that is used in the forward pass
-- `rrule_alg`: Reverse-rule for differentiating that SVD
+- `svd_fwd_alg=TensorKit.SDD()`: SVD algorithm that is used in the forward pass
+- `svd_rrule_type = Arnoldi`: Default solver type for SVD reverse-rule algorithm
+- `svd_rrule_alg`: Reverse-rule algorithm for differentiating a SVD
 
     ```
-    rrule_alg = Arnoldi(; tol=ctmrg_tol, krylovdim=48, verbosity=-1)
+    svd_rrule_alg = svd_rrule_type(; tol=ctmrg_tol, krylovdim=48, verbosity=-1)
```
leburgel, line 78: Would it be worth setting `eager=true` here? I don't know how many iterations this needs on average, but `krylovdim=48` seem fairly high to me so it might be worth it?

src/PEPSKit.jl
```diff
@@ -88,35 +94,43 @@ Module containing default algorithm parameter values and arguments.
...
+- `gradient_alg_maxiter=30`: Maximal number of iterations for computing the CTMRG fixed-point gradient
+- `gradient_alg_iterscheme=:fixed`: Scheme for differentiating one CTMRG iteration
 - `gradient_linsolver`: Default linear solver for the `LinSolver` gradient algorithm
 
     ```
-    gradient_linsolver=KrylovKit.BiCGStab(; maxiter=fpgrad_maxiter, tol=fpgrad_tol)
+    gradient_linsolver=KrylovKit.BiCGStab(; maxiter=gradient_alg_maxiter, tol=gradient_alg_tol)
     ```
 
 - `gradient_eigsolve`: Default eigsolver for the `EigSolver` gradient algorithm
```
leburgel, line 112:
```suggestion
- `gradient_eigsolver`: Default eigsolver for the `EigSolver` gradient algorithm
```

src/PEPSKit.jl
```diff
@@ -134,28 +148,39 @@ module Defaults
...
     const ctmrg_miniter = 4
+    const ctmrg_alg_type = SimultaneousCTMRG
     const sparse = false
     const trscheme = FixedSpaceTruncation()
-    const fwd_alg = TensorKit.SDD()
-    const rrule_alg = Arnoldi(; tol=ctmrg_tol, krylovdim=48, verbosity=-1)
-    const svd_alg = SVDAdjoint(; fwd_alg, rrule_alg)
+    const svd_fwd_alg = TensorKit.SDD()
+    const svd_rrule_type = Arnoldi
+    const svd_rrule_alg = svd_rrule_type(; tol=ctmrg_tol, krylovdim=48, verbosity=-1)
```
leburgel, line 156: Same comment: might be worth setting `eager=true` for the `Arnoldi` case.

src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -7,38 +7,66 @@ for contracting infinite PEPS.
...
-Each CTMRG run is converged up to `alg.tol` where the singular value convergence
-of the corners and edges is checked. The maximal and minimal number of CTMRG
-iterations is set with `alg.maxiter` and `alg.miniter`.
+The algorithm can be supplied via the keyword arguments or directly as an `CTMRGAlgorithm`
+struct. The following keyword arguments are supported:
 
-Different levels of output information are printed depending on `alg.verbosity`, where `0`
-suppresses all output, `1` only prints warnings, `2` gives information at the start and
-end, and `3` prints information every iteration.
+- `alg=SimultaneousCTMRG`: Variant of the CTMRG algorithm; can be any `CTMRGAlgorithm` type
```
leburgel, line 26:
```suggestion
- `alg=Defaults.ctmrg_alg`: Variant of the CTMRG algorithm; can be any `CTMRGAlgorithm` type
```

src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -7,38 +7,66 @@ for contracting infinite PEPS.
...
+
+- `tol=Defaults.ctmrg_tol`: Tolerance checking singular value and norm convergence; also
+  sets related tolerances to sensible defaults unless they are explicitly specified
+
+- `maxiter=Defaults.ctmrg_maxiter`: Maximal number of CTMRG iterations per run
+
+- `miniter=Defaults.ctmrg_miniter`: Minimal number of CTMRG carried out
```
leburgel, line 33: As a more general comment, would it be good to interpolate the actual default values into the docstring where possible (e.g. `maxiter=$(Defaults.ctmrg_maxiter)`? Now these default values really don't say much, given that a user would have to go look them up in the defaults module anyway. Maybe for some things interpolating would give a bad result in terms of formatting, but having at least some concrete values to look at would be nice I think.

src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -80,6 +108,57 @@ end
 @non_differentiable ctmrg_logfinish!(args...)
 @non_differentiable ctmrg_logcancel!(args...)
 
+"""
+    select_algorithm(
+        ::typeof(leading_boundary),
+        env₀::CTMRGEnv;
+        alg=SimultaneousCTMRG,
```
leburgel, line 115:
```suggestion
        alg=Defaults.ctmrg_alg,
```

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -90,12 +134,25 @@ function fixedpoint(
...
-        @warn "the provided real environment was converted to a complex environment since \
-        :fixed mode generally produces complex gauges; use :diffgauge mode instead to work \
-        with purely real environments"
+    # :fixed mode compatibility
+    if !isnothing(alg.gradient_alg) && iterscheme(alg.gradient_alg) == :fixed
+        if scalartype(env₀) <: Real # incompatible with real environments
+            env₀ = complex(env₀)
+            @warn "the provided real environment was converted to a complex environment
+            since :fixed mode generally produces complex gauges; use :diffgauge mode \
+            instead to work with purely real environments"
```
leburgel, line 143: This should probably state how to set `diffgauge`, i.e. as the `iterscheme` of the gradient algorithm. Otherwise someone might end up here when running all the defaults on a real environment, at which point the terms `:fixed` and `:diffgauge` might not mean much to them.

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -144,6 +201,136 @@ function fixedpoint(
...
+        boundary_verbosity = -1
+        gradient_verbosity = -1
+        optimizer_verbosity = -1
+    elseif verbosity == 1 # output only optimization steps and degeneracy warnings
+        boundary_verbosity = -1
+        gradient_verbosity = -1
+        optimizer_verbosity = 3
+    elseif verbosity == 2 # output optimization and boundary information
+        boundary_verbosity = 2
+        gradient_verbosity = -1
```
leburgel, line 242: It would be nice to have some gradient verbosity without having the full boundary verbosity, just to get the warnings if the gradient fails to converge. So I think it would be good to either add another level which does this, or put the gradient verbosity to `1` here.

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -144,6 +201,136 @@ function fixedpoint(
...
+    end
+
+    # parse fixed-point gradient algorithm
+    gradient_algorithm = if gradient_alg isa GradMode
+        gradient_alg
+    elseif gradient_alg isa NamedTuple
+        gradient_kwargs = (;
+            tol=1e-2tol,
+            maxiter=Defaults.gradient_alg_maxiter,
+            alg=typeof(Defaults.gradient_alg),
```
leburgel, line 273: Maybe also make this a default, `Defaults.gradient_alg_type`?

src/algorithms/ctmrg/ctmrg.jl
pbrehmer: I think this should rather be `alg=Defaults.ctmrg_alg_type` but other than that you're right!

src/algorithms/ctmrg/ctmrg.jl
pbrehmer: I would be fine with interpolating this for numbers, bools, symbols and so on. But interpolating structs or parametric types can quickly get out of hand and make the docstring unreadable. I left it like that because KrylovKit does it and I wanted to use the same convention everywhere but I do agree that it might be slightly annoying to always look up default values in the `Defaults` module. @lkdvos Do you have an opinion on this?

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -144,6 +201,136 @@ function fixedpoint(
...
+    end
+
+    # parse boundary algorithm
+    boundary_algorithm = if boundary_alg isa CTMRGAlgorithm
+        boundary_alg
+    elseif boundary_alg isa NamedTuple
+        select_algorithm(
+            leading_boundary,
+            env₀;
+            tol=1e-4tol,
```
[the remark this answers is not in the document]
pbrehmer: Indeed, this is overwritten by `boundary_alg...` if it contains the `tol` keyword (I know it looks slightly sketchy but it's also relatively concise :-))

src/PEPSKit.jl
pbrehmer: I agree, this is probably worth it, so let's try

src/algorithms/optimization/peps_optimization.jl
pbrehmer: Very good point, thanks. I added instructions for passing the right kwargs

pbrehmer:
Okay, apparently setting `eager=true` in the SVD rrule algorithm somehow leads to more `svdsolve` cotangent warnings and consequently the real-linearity in `jacobian_real_linear.jl` is much less accurate. For now I decided that `eager=false` is probably the more stable default but we should keep this in mind since it does probably speed up the SVD reverse pass in many cases.

With regards to the Krylov dimension of the SVD rrule `Arnoldi`: this has to be adjusted according to the environment dimension $\chi$ and I found that one has to choose it quite a bit above $\chi$ to make some things run stably (hence it's `krylovdim=48` by default). The keyword selection takes care of that based on the supplied environment but of course this is just a rule of thumb again.

src/algorithms/ctmrg/ctmrg.jl
lkdvos: I would definitely prefer interpolating here, otherwise it is quite a lot of text for not a lot of information, since you would still have to go and look up the values.

One thing that might actually be reasonable is to follow more closely the `OhMyThreads` approach, where you use symbols instead of the types.
In general, I would say fully instantiated structs are probably not that well-suited as `const` values in the defaults anyways, and should be handled by functions?

lkdvos (commented):
I think it's strange to have a constant krylov dim in the first place, it should typically scale with the number of desired singular values. I'm pretty sure that if you are requesting more values, it automatically just sets this to be larger (since it has to) (This is also the case for the eigensolvers, but there we typically only are interested in the dominant ones, which is why we get away with picking it fixed)

I added a docstring for `select_algorithm(::typeof(leading_boundary), ...)` with some subsections, which I think is a bit easier to follow. Let me know what you think about that.

As mentioned, I think for kwargs that can be algorithms, we should have a general interface that allows for:
* `alg::Algorithm`: directly use the algorithm.
* `(; alg::Union{Symbol,Algorithm}, alg_kwargs...)::NamedTuple`: construct an algorithm from `kwargs`, using `alg` to select between multiple flavours.

pbrehmer:
> I think it's strange to have a constant krylov dim in the first place, it should typically scale with the number of desired singular values. I'm pretty sure that if you are requesting more values, it automatically just sets this to be larger (since it has to) (This is also the case for the eigensolvers, but there we typically only are interested in the dominant ones, which is why we get away with picking it fixed)

Does it really set this automatically? I thought that KrylovKit by default also sets this to some constant value. I was previously getting errors when $\chi$ was too big for the default Krylov dimension, due to [this line]( https://github.com/Jutho/KrylovKit.jl/blob/a39cc4f88f5f7a0935aae52a60372b1d0825773b/src/eigsolve/svdsolve.jl#L153-L154). So, what do you suggest we do here?

> I added a docstring for `select_algorithm(::typeof(leading_boundary), ...)` with some subsections, which I think is a bit easier to follow. Let me know what you think about that.

I really like the docstring and using Markdown subsections. But wouldn't it make more sense to have that description in `leading_boundary` since users actually use that function? Having kwarg descriptions in both the `select_algorithm` and `leading_boundary` docstring is just duplicating the same thing, right?

pbrehmer:
> One thing that might actually be reasonable is to follow more closely the OhMyThreads approach, where you use symbols instead of the types.
In general, I would say fully instantiated structs are probably not that well-suited as const values in the defaults anyways, and should be handled by functions?

Could you elaborate on how to handle defaults with functions instead of structs? I do agree that it is a bit weird to have structs but I also don't mind that way (at the moment). However, I'm also open to redoing the defaults with functions once I get the idea :)

lkdvos:
> Does it really set this automatically?

Apparently not, but it really **should** 😉 . I would suggest we have a krylovdim factor, and simply use `krylovdim = round(Int, howmany * krylovdim_factor)`.

> But wouldn't it make more sense to have that description in leading_boundary since users actually use that function?

Oh yeah, definitely better there, I just wanted some place to try it out and not overwrite too many things.

> Could you elaborate on how to handle defaults with functions instead of structs?

OhMyThreads has the following pattern: `tmap(args...; scheduler = :dynamic, kwargs...)`. Basically, instead of putting an actual `Scheduler` as a default value, you put in all the keywords in terms of basic data types, and only instantiate when you run the function.

This also might help with putting the `Defaults` module first? In any case, I'd suggest we move that to a separate folder.

pbrehmer:
Only using concrete types in `Defaults` is a bit complicated further down the stack, so I had to try around a bit. I decided to use the `select_algorithm` function to map symbols to algorithm types and enable passing algorithms as `NamedTuple`s. I'll finish this up tomorrow.

pbrehmer:
Right now the defaults in the kwarg constructors of various algorithm structs need to interpolate from symbols which is annoying since they can only be defined after the definition of the struct itself. This will hopefully be coming along tomorrow...

lkdvos:
I'm a bit confused as to why you need the symbol-dicts? I think we can get away with the default value for these keyword arguments really being the `Symbol`, and then making the constructors of the `Algorithm`s, or `select_algorithm` handle these?

pbrehmer:
In the end it turned out that supporting `Symbol`s, `Type{Algorithm}` and `Algorithm` does require a bit of care. Most of the complexity is now in the `select_algorithm` methods which also the algorithm struct constructors now call. I tried to update the docstrings to list all the options by also stating the possible `Symbols`.

From the user point of view I still feel that setting parameters via nested `NamedTuple`s is a bit complicated but there is probably no way which would make this substantially less complicated - the PEPS stack is just quite complicated.

src/Defaults.jl
lkdvos, line 1: I like this new version a lot, thanks for cleaning this up!

Small question/suggestion: the current output uses markdown h1 section titles, we could also decide to use h2 or h3 instead (I think I prefer h2, but fine with either). Here is an example of all three
```
  CTMRG
  ≡≡≡≡≡

    •  ctmrg_tol=1.0e-8: Tolerance checking singular value and norm convergence.
...

  SVD forward & reverse
  =====================

    •  trscheme=:fixedspace: Truncation scheme for SVDs and other decompositions.
...

  Projectors
  ––––––––––

    •  projector_alg=:halfinfinite: Default variant of the CTMRG projector algorithm.
...

Comment (lkdvos) [reply src/Defaults.jl:1]: I've changed it for now, but let me know if we want something else, I can definitely change it back

File: src/algorithms/ctmrg/projectors.jl [hunk 13]
```diff
@@ -33,27 +33,59 @@ function truncation_scheme(alg::ProjectorAlgorithm, edge)
...
-    svd_alg::S = Defaults.svd_alg
-    trscheme::T = Defaults.trscheme
-    verbosity::Int = 0
+struct HalfInfiniteProjector{S<:SVDAdjoint,T} <: ProjectorAlgorithm
+    svd_alg::S
+    trscheme::T
+    verbosity::Int
+end
+function HalfInfiniteProjector(; kwargs...)
+    return select_algorithm(ProjectorAlgorithm; alg=:halfinfinite, kwargs...)
```
lkdvos, line 60: This pattern I like a bit less: we already have a constructor, and dispatching that to `select_algorithm` seems like a reroute that does not bring to much to the table. In particular, calling `HalfInfiniteProjector` on something really has to give back a `::HalfInfiniteProjector`, so pushing it to a `select_algorithm(::Type{ProjectorAlgorithm})` feels a bit wrong.

lkdvos (commented):
Thanks a lot for all of the work you've put into this already, it's really coming together nicely and I like the improvements a lot!

I've made some small changes still, and have some general comments for which I apologize in advance, since some of these might contradict things I have hinted at earlier.

I really like the `select_algorithm` pattern for functions, I think that makes a lot of sense and cleanly separates out handling the kwargs from the actual algorithm itself.
I don't like this pattern as much for the `select_algorithm(T::Type)` versions, because there we're not selecting an algorithm for implementing `T(args...)`, we really are just constructing a `alg::T`. I think it makes a bit more sense to keep these as actual constructors. For example, we can also define constructors for abstract supertypes that return a concrete subtype and selects that based on the provided (keyword) arguments.
If you agree with this, I should be able to make some time for implementing this this afternoon, if you are getting sick of working on this for a bit I can definitely help out ;)

I'm a bit torn between having the `select_algorithm` in a separate file for all functions, or having them bundled with their respective implementations (ie closer to where the docstring would be). I think I like both, but I'm slightly worried about the order of having to define things, ie that some types/functions might not exist yet etc. Perhaps that's not really a problem and we can always define `select_algorithm` after everything is already defined though. Just wanted to explicitly bring this up and see what we all think.

I've made a slight formatting change where I added a space before the `:` that comes after the name of the kwarg, because I felt like there are too many `::` and `:` with the symbols etc and wanted to cleanly separate out the code part from the text part a bit more. Let me know if you don't like it and I'll change it back.

Are we now still using the `(; alg=::Type{})` form anywhere? I think I may have suggested this at some point, but given that there are now symbols everywhere, I think it would be fine (preferential even) to just drop that syntax altogether. It seems a bit artificial, and does not really give any benefit over the symbols I think.
I've added some dictionaries instead of the `if` statements for these symbols, mostly to allow them to be externally extensible if anyone would want that.

Again, if you want I can make some time today for working some of these things in, feel free to ask for help :)

src/Defaults.jl
pbrehmer: I'm fine with all of the versions so feel free to choose the one you like best :-)

pbrehmer:
> I don't like this pattern as much for the select_algorithm(T::Type) versions, because there we're not selecting an algorithm for implementing T(args...), we really are just constructing a alg::T. I think it makes a bit more sense to keep these as actual constructors.

Actually I'm also somewhat conflicted on this. I think I went this route since the `select_algorithm` functions compose a bit and I didn't see (at the time) how to easily do this with the algorithm constructors without duplicating a lot of code. Please feel free to change this and implement the corresponding constructors if you feel it makes things a bit more consistent and cleaner! (I am a bit sick of working on this I have to admit ;)) 

> I'm a bit torn between having the select_algorithm in a separate file for all functions, or having them bundled with their respective implementations

I'm also torn on this and I mostly did that to have all selections in one place since they do depend on each other. I think I'd be fine with both ways but I previously did have issues with things not yet being defined which is why I put the `select_algorithm.jl` file at the end of the includes.

> Are we now still using the (; alg=::Type{}) form anywhere? I think I may have suggested this at some point, but given that there are now symbols everywhere, I think it would be fine (preferential even) to just drop that syntax altogether.

No we don't use that form anymore really. Somehow I thought that we wanted to keep this but in hindsight I don't know why I didn't remove this right away. It just makes things more complicated and blows up the docstrings. I would definitely be in favor of removing this.

pbrehmer:
I will remove the `(; alg::Type{}=alg)` syntax now and adapt the docstrings but then feel free to bring the PR over the finish line :-) Of course I'm happy to help!

lkdvos:
I'll try and go through this tomorrow to see if there are any remaining things, but should be more or less finished, so we can merge this Monday

lkdvos (dismissed):
I think this is now okay to merge for me. There are definitely still a couple kinks left, especially with the Krylov dimension, but I think these really deserve their own PRs anyways.

Given that this is somewhat large/opnionated PR, let's get some more reviews in: @Yue-Zhengyuan and @leburgel , do you have any more comments? (Keeping in mind that we can still make changes in follow-up PRs)

pbrehmer:
Thanks for cleaning up the rest, looks great to me! I'd be fine with merging as is.

Regarding the Krylov dimension: I think once we have #150 ready (hopefully soon), then we can hopefully use the TensorKit rrule per default also for the `:fixed` mode - meaning that the Krylov dimension isn't the most pressing issue, I think.

test/boundarymps/vumps.jl
```diff
@@ -33,9 +31,7 @@ end
     mps, env, ϵ = leading_boundary(mps, T, vumps_alg)
```
Yue-Zhengyuan, line 31: Shall we also implement something like
```julia
leading_boundary(mps, T; alg=:vumps, kwargs...)
```

src/algorithms/ctmrg/projectors.jl
```diff
@@ -33,29 +52,65 @@ function truncation_scheme(alg::ProjectorAlgorithm, edge)
...
-@kwdef struct HalfInfiniteProjector{S<:SVDAdjoint,T} <: ProjectorAlgorithm
-    svd_alg::S = Defaults.svd_alg
-    trscheme::T = Defaults.trscheme
-    verbosity::Int = 0
+struct HalfInfiniteProjector{S<:SVDAdjoint,T} <: ProjectorAlgorithm
+    svd_alg::S
+    trscheme::T
+    verbosity::Int
 end
+function HalfInfiniteProjector(; kwargs...)
```
Yue-Zhengyuan, line 78: I think `select_algorithm(ProjectorAlgorithm; ...)` is not used throughout the package? It's even not implemented if I'm not mistaken, and not covered by the tests either. The selection of the projector algorithm is actually done above using the `ProjectorAlgorithm` constructor directly. If not used anywhere, I suggest to remove `select_algorithm` for `ProjectorAlgorithm`.

Yue-Zhengyuan (commented):
Thanks to @pbrehmer for improving the user interface! I only have a few minor suggestions below about things more relevant to my use cases. In addition, we may also update the SU examples `examples/heisenberg_evol/simpleupdate.jl` and `examples/hubbard_su.jl` to use the new `leading_boundary` syntax (I can push the changes to this branch if you don't mind). 

BTW, I think that the old way to run CTMRG by providing the parameters as a combined `ctm_alg` actually still works? Since the new syntax just helps users to combine them, so users don't have to do it by themselves.

src/algorithms/ctmrg/projectors.jl
```diff
@@ -33,29 +52,65 @@ function truncation_scheme(alg::ProjectorAlgorithm, edge)
...
 """
-    struct FullInfiniteProjector{S,T}(; svd_alg=Defaults.svd_alg,
-                                      trscheme=Defaults.trscheme, verbosity=0)
+    struct FullInfiniteProjector{S,T}
 
 Projector algorithm implementing projectors from SVDing the full 4x4 CTMRG environment.
+
+## Keyword arguments
+
+* `svd_alg::Union{<:SVDAdjoint,NamedTuple}=SVDAdjoint()` : SVD algorithm including the reverse rule. See ['SVDAdjoint'](@ref).
```
leburgel, line 91: Is this the actual default? It's a bit hard to follow here so maybe it's equivalent to what's happening, but I can't see where this specific default is used.

src/algorithms/ctmrg/projectors.jl
```diff
@@ -33,29 +52,65 @@ function truncation_scheme(alg::ProjectorAlgorithm, edge)
...
 """
-    struct HalfInfiniteProjector{S,T}(; svd_alg=Defaults.svd_alg,
-                                      trscheme=Defaults.trscheme, verbosity=0)
+    struct HalfInfiniteProjector{S,T}
 
 Projector algorithm implementing projectors from SVDing the half-infinite CTMRG environment.
+
+## Keyword arguments
+
+* `svd_alg::Union{<:SVDAdjoint,NamedTuple}=SVDAdjoint()` : SVD algorithm including the reverse rule. See ['SVDAdjoint'](@ref).
```
leburgel, line 61: Is this the actual default? It's a bit hard to follow here so maybe it's equivalent to what's happening, but I can't see where this specific default is used.

src/algorithms/ctmrg/sequential.jl
```diff
@@ -1,13 +1,24 @@
 """
-    SequentialCTMRG(; tol=Defaults.ctmrg_tol, maxiter=Defaults.ctmrg_maxiter,
-                      miniter=Defaults.ctmrg_miniter, verbosity=0,
-                      projector_alg=typeof(Defaults.projector_alg),
-                      svd_alg=SVDAdjoint(), trscheme=FixedSpaceTruncation())
+    SequentialCTMRG(; tol=$(Defaults.ctmrg_tol), maxiter=$(Defaults.ctmrg_maxiter),
+                      miniter=$(Defaults.ctmrg_miniter), verbosity=$(Defaults.ctmrg_verbosity),
+                      svd_alg=TODO, trscheme=TODO,
+                      projector_alg=TODO)
```
leburgel, line 5: These should still be filled in before merging I guess?

src/algorithms/ctmrg/simultaneous.jl
```diff
@@ -1,13 +1,20 @@
 """
-    SimultaneousCTMRG(; tol=Defaults.ctmrg_tol, maxiter=Defaults.ctmrg_maxiter,
-                      miniter=Defaults.ctmrg_miniter, verbosity=0,
-                      projector_alg=Defaults.projector_alg,
-                      svd_alg=SVDAdjoint(), trscheme=FixedSpaceTruncation())
+    struct SimultaneousCTMRG
```
leburgel, line 2: Is there a reason the kwarg calling signature was removed? I agree it's a bit redundant given the list of keyword arguments below, but I would at least add a line
```julia
SimultaneousCTMRG(; kwargs...)
```
to the signature to make it clear there is a keyword-only call available.

src/algorithms/ctmrg/projectors.jl
lkdvos: Good catch, thanks!

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -1,99 +1,167 @@
...
+
+    return gradient_algorithm
+end
+
 iterscheme(::GradMode{F}) where {F} = F
 
 """
-    struct GeomSum(; maxiter=Defaults.fpgrad_maxiter, tol=Defaults.fpgrad_tol,
-                   verbosity=0, iterscheme=Defaults.iterscheme) <: GradMode{iterscheme}
+    struct GeomSum <: GradMode{iterscheme}
```
leburgel, line 66: Same comment, I would add a line
```julia
GeomSum(; kwargs...)
```
to the signature so people know there's a keyword-only call.

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -1,99 +1,167 @@
...
-    return GeomSum{iterscheme}(maxiter, tol, verbosity)
-end
+GeomSum(; kwargs...) = GradMode(; alg=:geomsum, kwargs...)
+
+GRADIENT_MODE_SYMBOLS[:geomsum] = GeomSum
 
 """
-    struct ManualIter(; maxiter=Defaults.fpgrad_maxiter, tol=Defaults.fpgrad_tol,
-                      verbosity=0, iterscheme=Defaults.iterscheme) <: GradMode{iterscheme}
+    struct ManualIter <: GradMode{iterscheme}
```
leburgel, line 92: Same comment, i would add a line
```julia
ManualIter(; kwargs...)
```

test/boundarymps/vumps.jl
lkdvos: I'm definitely also happy to do this, although that might make more sense in MPSKit directly?

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -1,99 +1,167 @@
...
-)
-    return ManualIter{iterscheme}(maxiter, tol, verbosity)
-end
+ManualIter(; kwargs...) = GradMode(; alg=:manualiter, kwargs...)
+
+GRADIENT_MODE_SYMBOLS[:manualiter] = ManualIter
 
 """
-    struct LinSolver(; solver=KrylovKit.GMRES(), iterscheme=Defaults.iterscheme) <: GradMode{iterscheme}
+    struct LinSolver <: GradMode{iterscheme}
```
leburgel, line 118:
```suggestion
    struct LinSolver <: GradMode{iterscheme}
    LinSolver(; kwargs...)
```

lkdvos:
Thanks for the comments. I would be very happy if you want to add the changes you are suggesting here

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -1,99 +1,167 @@
...
-    return LinSolver{iterscheme}(solver)
+    solver_alg::KrylovKit.LinearSolver
 end
+LinSolver(; kwargs...) = GradMode(; alg=:linsolver, kwargs...)
+
+GRADIENT_MODE_SYMBOLS[:linsolver] = LinSolver
 
 """
-    struct EigSolver(; solver=KrylovKit.Arnoldi(), iterscheme=Defaults.iterscheme) <: GradMode{iterscheme}
+    struct EigSolver <: GradMode{iterscheme}
```
leburgel, line 143:
```suggestion
    struct EigSolver <: GradMode{iterscheme}
    EigSolver(; kwargs...)
```

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -1,99 +1,167 @@
...
 problem as an eigenvalue problem.
 
-With `iterscheme` the style of CTMRG iteration which is being differentiated can be chosen.
-If set to `:fixed`, the differentiated CTMRG iteration is assumed to have a pre-computed
-SVD of the environments with a fixed set of gauges. Alternatively, if set to `:diffgauge`,
-the differentiated iteration consists of a CTMRG iteration and a subsequent gauge fixing step,
-such that `gauge_fix` will also be differentiated everytime a CTMRG derivative is computed.
+## Keyword arguments
+
+* `tol::Real=$(Defaults.gradient_tol)` : Convergence tolerance of the linear solver.
```
leburgel, line 150:
```suggestion
* `tol::Real=$(Defaults.gradient_tol)` : Convergence tolerance of the eigen solver.
```

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -1,99 +1,167 @@
...
+* `maxiter::Int=$(Defaults.gradient_maxiter)` : Maximal number of solver iterations.
+* `verbosity::Int=$(Defaults.gradient_verbosity)` : Output information verbosity of the linear solver.
+* `iterscheme::Symbol=:$(Defaults.gradient_iterscheme)` : Style of CTMRG iteration which is being differentiated, which can be:
+    - `:fixed` : the differentiated CTMRG iteration uses a pre-computed SVD with a fixed set of gauges
+    - `:diffgauge` : the differentiated iteration consists of a CTMRG iteration and a subsequent gauge-fixing step such that the gauge-fixing procedure is differentiated as well
+* `solver_alg::Union{KrylovKit.KrylovAlgorithm,NamedTuple}=(; alg=:$(Defaults.gradient_eigsolver)` : Eigen solver algorithm which, if supplied directly as a `KrylovKit.KrylovAlgorithm` overrides the above specified `tol`, `maxiter` and `verbosity`. Alternatively, it can be supplied via a `NamedTuple` where `alg` can be one of the following:
+    - `:arnoldi` : Arnoldi Krylov algorithm, see the [KrylovKit docs](https://jutho.github.io/KrylovKit.jl/stable/man/algorithms/#KrylovKit.Arnoldi) for details
```
leburgel, line 157: More as a reminder for the future, we should enable external references when we update the documentation.

src/utility/svd.jl
```diff
@@ -10,20 +10,89 @@ using TensorKit:
...
 In case of degenerate singular values, one might need a `broadening` scheme which
 removes the divergences from the adjoint.
+
+## Keyword arguments
+
+* `fwd_alg::Union{Algorithm,NamedTuple}=(; alg::Symbol=Defaults.svd_fwd_alg)`: SVD algorithm of the forward pass which can either be passed as an `Algorithm` instance or a `NamedTuple` where `alg` is one of the following:
+    - `:sdd`: TensorKit's wrapper for LAPACK's `_gesdd`
+    - `:svd`: TensorKit's wrapper for LAPACK's `_gesvd`
+    - `:iterative`: Iterative SVD only computing the specifed number of singular values and vectors, see ['IterSVD'](@ref)
+* `rrule_alg::Union{Algorithm,NamedTuple}=(; alg::Symbol=Defaults.svd_rrule_alg)`: Reverse-rule algorithm for differentiating the SVD. Can be supplied by an `Algorithm` instance directly or as a `NamedTuple` where `alg` is one of the following:
```
leburgel, line 26:
```suggestion
* `rrule_alg::Union{Algorithm,NamedTuple}=(; alg::Symbol=$(Defaults.svd_rrule_alg))`: Reverse-rule algorithm for differentiating the SVD. Can be supplied by an `Algorithm` instance directly or as a `NamedTuple` where `alg` is one of the following:
```

src/utility/svd.jl
```diff
@@ -10,20 +10,89 @@ using TensorKit:
...
+        return new{F,R,B}(fwd_alg, rrule_alg, broadening)
+    end
 end  # Keep truncation algorithm separate to be able to specify CTMRG dependent information
 
+const SVD_FWD_SYMBOLS = IdDict{Symbol,Any}(
+    :sdd => TensorKit.SDD,
+    :svd => TensorKit.SVD,
+    :iterative =>
+        (; tol=1e-14, krylovdim=25, kwargs...) ->
+            IterSVD(; alg=GKL(; tol, krylovdim), kwargs...),
```
leburgel, line 53: This feels a bit hardcoded and opaque, but I guess that's fine for now. I'm not sure if it's worth adding documented defaults for this right now, since I doubt anyone would go this deep in the near future.

test/ctmrg/fixed_iterscheme.jl
```diff
@@ -15,7 +15,7 @@ using PEPSKit:
 χbond = 2
 χenv = 16
 svd_algs = [SVDAdjoint(; fwd_alg=TensorKit.SDD()), SVDAdjoint(; fwd_alg=IterSVD())]
-projector_algs = [HalfInfiniteProjector] #, FullInfiniteProjector]
+projector_algs = [:halfinfinite] #, :fullinfinite]
```
leburgel, line 18: Kind of unrelated, but why is the full infinite projector commented out? What was the problem, and mainly, is there an issue tracking this?

pbrehmer:
> BTW, I think that the old way to run CTMRG by providing the parameters as a combined ctm_alg actually still works?

Indeed, the old syntax still works and its intended as the "expert" mode since that exposes all the algorithm structs at once. So if you want to use that style you can still do so.

leburgel:
Looks great overall, I really think this is a massive improvement. Left some final comments, I'm happy to actually apply the suggestions people agree with, just say so and tag me so there's no conflicts with who's working on what :)

src/Defaults.jl
```diff
@@ -0,0 +1,139 @@
+"""
+    module Defaults
```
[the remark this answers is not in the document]
pbrehmer: I tried to add all options at the docstrings where the kwargs show up but I agree that it is also useful to have this in the `Defaults` directly.

src/algorithms/ctmrg/sequential.jl
pbrehmer: Whoopsie, those were supposed to be removed since the docstrings lists all kwargs below. But I do like the suggestions of at least mentioning them as `SequentialCTRMG(; kwargs...)`.

src/utility/svd.jl
pbrehmer: I also realized this but I think we shouldn't worry too much about this right now and return to that bit once we look at the SVD interface more closely. Anybody that wants to use an iterative SVD can still do so and also change the defaults, even if it is a bit more complicated :)

test/ctmrg/fixed_iterscheme.jl
pbrehmer: Good point: IIRC there were stability issues with the full-infinite scheme and the element-wise convergence. That was actually not the only thing where full-infinite did weird things. I'll take a quick look and file an issue!

test/boundarymps/vumps.jl
pbrehmer: I intentionally left this out for now since I wasn't sure what the best option would be. Eventually, I would also like to have the `alg=:vumps` option but maybe we can postpone that for a separate PR?

Yue-Zhengyuan: Then I'm OK to leave this to a separate PR 😄

pbrehmer:
I can also quickly add all the remaining suggestions, thanks for the comments! :)

src/algorithms/ctmrg/projectors.jl
pbrehmer: Well, it kind of is. That's just a short way of writing that the default `SVDAdjoint` is used. Internally, a `NamedTuple` of kwargs is passed onto the `SVDAdjoint` constructor which defines the defaults for `fwd_alg`, `rrule_alg` and `broadening`. But that's a bit too long to write out I find.

pbrehmer:
Alright, this is mergeable from my side, once the tests turn green!

src/algorithms/ctmrg/projectors.jl
leburgel: Okay, if the outcome is equivalent then this is good enough, just wanted to check because I was confused.

leburgel (approved):

test/ctmrg/fixed_iterscheme.jl
pbrehmer: See #152

Issue #115, opened 2025-01-10
Generalization of the CTMRG `state` type


pbrehmer:
Currently, CTMRG only runs on `InfinitePEPS` as its state type, and with #111 also on `InfinitePartitionFunction`s. However, all CTMRG contractions as well as the sparse storage structs in principle support more general objects, such as PEPS-PEPS or PEPS-PEPO-PEPS sandwiches.

In order to generalize the state type that is specified by the user and passed on to `leading_boundary` we need to agree on a design. Looking at the implementation of CTMRG, it is clear that the type of `state` needs to support indexing operations and rotations. I think there are at least two options:

1. We could create a `PEPSSandwich` and `PEPOSandwich` type. Here the question would be if they should be subtyped as an `InfiniteSquareNetwork` - this would however somehow create a circular hierarchy in its current state.
2. We could just use tuples of `InfiniteSquareNetworks` to represent the sandwiches, e.g. `Tuple{InfintePEPS,InfinitePEPS}`, and define indexing methods such as `getindex` on them.

Another thing to consider is to relax the `InfiniteSquareNetwork` interface such that it is understood as an interface which states that can be contracted using CTMRG should support. This would e.g. mean getting rid of the math and OptimKit operations.

See #111 for further comments.

lkdvos:
Just to make a comment here, I think there are two separate issues you are touching upon.

The first is about *data structures*, in the sense of how we actually lay out the data as julia structs. There are many options, using tuples or custom structs, and my guess is that the performance implications are completely negligible, so I would just choose the thing that is most easy to interpret as a human. In other words, whatever is the most intuitive for each specific instance should be fine.

A second point is about designing *functions*. Here, it is a very common pitfall to think in terms of abstract types and type hierarchies, which is I fear somewhat the direction this question is pushing towards. Very often, prematurely restricting the type signature of functions leads to bad code architectures, since you then are effectively designing code around objects. Julia often shines best by not thinking about the data structures at all, and simply ducktyping everything. This means that rather than thinking about what types of arguments the CTMRG routine should accept, it is a lot better to think about what functions the arguments should support.

TLDR: instead of trying to fit everything into abstract types, I would focus more on trying to avoid specializing methods as much as possible, especially the higher level *parent methods* and only doing that whenever the actual layout of an object is known, writing concrete implementations for *leaf methods*. That way, we can decide on future additions in the future, and simply implement only the leaf methods whenever needed.

PR #131, opened 2025-02-14
Add unified contractible network interface

leburgel:
A first attempt at implementing a concrete contractible network type, as discussed in #115. See comments below for a summary.


src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -27,52 +29,51 @@ Different levels of output information are printed depending on `alg.verbosity`,
 suppresses all output, `1` only prints warnings, `2` gives information at the start and
 end, and `3` prints information every iteration.
 """
-function MPSKit.leading_boundary(state, alg::CTMRGAlgorithm)
-    return MPSKit.leading_boundary(CTMRGEnv(state, oneunit(spacetype(state))), state, alg)
+function MPSKit.leading_boundary(network, alg::CTMRGAlgorithm)
+    return MPSKit.leading_boundary(CTMRGEnv(network, oneunit(spacetype(network))), network, alg)
```
leburgel, line 33: Should we just remove this one? Starting from a product environment by default seems very dangerous if the CTMRG truncation options are not set up for this. I don't think it's unreasonable to just require an initial guess where the user specifies some desirable structure for the environment virtual space, as this is not really a case where a sensible default can be chosen.

src/algorithms/ctmrg/ctmrg.jl
```diff
@@ -27,52 +29,51 @@ Different levels of output information are printed depending on `alg.verbosity`,
...
+                ctmrg_logiter!(log, iter, η, network, env)
             end
         end
         return env, info
     end
 end
 
-# network-specific objective functions
-ctmrg_objective(state::InfinitePEPS, env::CTMRGEnv) = norm(state, env)
-ctmrg_objective(state::InfinitePartitionFunction, env::CTMRGEnv) = value(state, env)
```
leburgel, line 64: Since CTMRG now really runs on a concrete contractible network, the objective is just the value of that network. So I removed `ctmrg_objective` and just use `value` (and then implement the PEPS norm via `value` of the corresponding `InfiniteSquareNetwork`...).

src/algorithms/ctmrg/sparse_environments.jl
```diff
@@ -193,31 +153,25 @@ linear map or adjoint linear map on `x` if `Val(true)` or `Val(false)` is passed
...
 # AbstractTensorMap subtyping and IterSVD compatibility
 # -----------------------------------------------------
 
-function _domain_space(A::NTuple{2,<:PEPSTensor}, n::Int)
-    return domain(A[1])[n] * domain(A[2])[n]'
-end
-function _domain_space(A::NTuple{1,<:PFTensor}, n::Int)
-    return domain(A[1])[n]
+function _domain_space(A, n::Int)
+    return virtualspace(A, n)
```
leburgel, line 170: I streamlined the virtual space handling of the different local sandwiches, so I don't know if this is still necessary? I could just remove `_domain_space` in favor of a direct `virtualspace` call.

src/algorithms/ctmrg/sparse_environments.jl
```diff
@@ -193,31 +153,25 @@ linear map or adjoint linear map on `x` if `Val(true)` or `Val(false)` is passed
...
 end
 
-function _codomain_space(A::NTuple{2,<:PEPSTensor}, n::Int)
-    return domain(A[1])[n]' * domain(A[2])[n]
-end
-function _codomain_space(A::NTuple{1,<:PFTensor}, n::Int)
-    return domain(A[1])[n]'
+function _codomain_space(A, n::Int)
+    return _elementwise_dual(virtualspace(A, n))
```
leburgel, line 174: Same comment here, we can now go through `virtualspace` directly.

src/algorithms/ctmrg/sparse_environments.jl
```diff
@@ -193,31 +153,25 @@ linear map or adjoint linear map on `x` if `Val(true)` or `Val(false)` is passed
...
 end
 
 function TensorKit.domain(env::HalfInfiniteEnv)
```
leburgel, line 177: We could just use `south_virtualspace`?

src/algorithms/toolbox.jl
```diff
@@ -53,90 +53,140 @@ function cost_function(peps::InfinitePEPS, env::CTMRGEnv, O::LocalOperator)
...
-            env.corners[SOUTHEAST, rnext, c][χ5; χ6] *
-            env.corners[SOUTHWEST, rnext, cprev][χ6; χ1]
-        total /= @autoopt @tensor env.corners[NORTHWEST, rprev, cprev][χ1; χ2] *
-            env.edges[NORTH, rprev, c][χ2 D1 D2; χ3] *
-            env.corners[NORTHEAST, rprev, cnext][χ3; χ4] *
-            env.corners[SOUTHEAST, r, cnext][χ4; χ5] *
-            env.edges[SOUTH, r, c][χ5 D1 D2; χ6] *
-            env.corners[SOUTHWEST, r, cprev][χ6; χ1]
-    end
+    return value(InfiniteSquareNetwork(peps), env)
```
leburgel, line 56: I think it makes sense to implemented the `norm` of an `InfinitePEPS` as the value of the corresponding norm network.

src/environments/ctmrg_environments.jl
```diff
@@ -29,17 +29,20 @@ struct CTMRGEnv{C,T}
...
+}
 const SpaceLike = Union{ElementarySpaceLike,ProductSpaceLike}
 
+_elementwise_dual(s::NTuple{N,<:ElementarySpace}) where {N} = dual.(s)
+
 _spacetype(::Int) = ComplexSpace
 _spacetype(::S) where {S<:ElementarySpace} = S
 _spacetype(s::ProductSpaceLike) = _spacetype(first(s))
 
 _to_space(χ::Int) = ℂ^χ
```
leburgel, line 43: Not directly related to this PR, but what would be our stance on dropping support for integers to specify spaces, and instead really require `ElementarySpace` or `ProductSpace` instances? I see how this can be convenient for quickly constructing an `InfinitePEPS`But usually we construct an environment through `CTMRGEnv(network, vspace)`, at which point this means mixing integers and proper instantiated spaces. It's doable, but it really makes the whole logic here unnecessarily complicated in my opinion. It would be worth it if people actually systematically used integer space specification, but I doubt this is the case since none of the examples and tests use that...

src/networks/infinitegridnetwork.jl
```diff
@@ -0,0 +1,169 @@
+# To be deprecated(?): supertype for physics-inspired networks which defines an interface
+# for optimization over these kinds of networks.
+
+"""
+    InfiniteGridNetwork{T,N}
+
+Abstract infinite tensor network consisting of a translationally invariant unit cell
+on a hypercubic lattice.
+"""
+abstract type InfiniteGridNetwork{T,N} end
```
leburgel, line 10: I renamed the old `InfiniteSquareNetwork` to `InfiniteGridNetwork` but kept it as a supertype in an initial pass, just not to break everything too much. Also, the codecov annotations are quite useful in assessing which parts of this assumed common interface are actually used. We need:

- `copy`
- `size` and `axes`
- A `getindex` and a corresponding `rrule`
- Some way to access physical spaces and virtual spaces in all directions
- VectorInterface and basic linear algebra for Zygote accumulation and OptimKit compatibility
- Rotations of the unit cell

Since accessing virtual spaces, physical spaces and rotations is really different for `InfinitePartitionFunction`, `InfinitePEPS` and `InfinitePEPO` and since some of the things here (like `rrule`s for the rotations are no longer used), I think it's reasonable to just remove the supertype and implement the interface for each concrete type separately. That way we can also easily add more network types without subtyping. There will be some duplication, but probably that's better than keeping this around given that it's a bit forced in places and quite restrictive for the future?

Just wanted to check before I actually do this...

src/networks/infinitesquarenetwork.jl
```diff
@@ -1,60 +1,77 @@
...
+            _elementwise_dual(south_virtualspace(A[_prev(d, end), w])) || throw(
+                SpaceMismatch("North virtual space at site $((d, w)) does not match.")
+            )
+            east_virtualspace(A[d, w]) ==
+            _elementwise_dual(west_virtualspace(A[d, _next(w, end)])) ||
+                throw(SpaceMismatch("East virtual space at site $((d, w)) does not match."))
+        end
+        return new{O}(A)
+    end
+end
```
leburgel, line 21: This is the concrete contractible network type discussed in #115 . It wraps a matrix of local rank-4-like objects, and just requires some access to the virtual spaces in all directions, which can be `ProductSpace`s.

Just adding a converter and a corresponding `rrule` for the physics network makes it so all contractions directly work with these objects, where only the contractions have to actually specialize on the local object type.

src/networks/infinitesquarenetwork.jl
```diff
@@ -74,42 +91,60 @@ function Base.isapprox(A₁::InfiniteSquareNetwork, A₂::InfiniteSquareNetwork;
...
+function ChainRulesCore.rrule(
+    ::Type{NWType}, A::Matrix
+) where {NWType<:InfiniteSquareNetwork}
+    network = NWType(A)
+    function InfiniteSquareNetwork_pullback(Δnetwork)
+        Δnetwork = unthunk(Δnetwork)
+        return NoTangent(), unnitcell(Δnetwork)
+    end
+    return network, InfiniteSquareNetwork_pullback
 end
```
leburgel, line 132: I put this in just to be complete, but since we always make contractible networks from states this is never actually used. Should I just remove it?

src/networks/infinitesquarenetwork.jl
```diff
@@ -74,42 +91,60 @@ function Base.isapprox(A₁::InfiniteSquareNetwork, A₂::InfiniteSquareNetwork;
...
+function ChainRulesCore.rrule(
+    ::typeof(Base.getindex), network::InfiniteSquareNetwork, args...
+)
+    O = network[args...]
+
+    function getindex_pullback(ΔO_)
+        ΔO = unthunk(ΔO_)
+        if ΔO isa Tangent # TODO: figure out why this happens in the first place...
+            ΔO = ChainRulesCore.construct(typeof(O), ChainRulesCore.backing(ΔO))
+        end
```
leburgel, line 114: I have no idea why, but this always gets a `Tangent` of the expected local sandwich type instead of a normal instance of the local sandwich type. Everywhere I check with `@showtypeofgrad` the adjoint being accumulated has the expected type, but still this somehow gets a `Tangent`.

The explicit `construct` fixes things so this is probably fine, but it would still be nice to not have to do this...

src/networks/local_sandwich.jl
```diff
@@ -0,0 +1,95 @@
...
+function ChainRulesCore.rrule(::typeof(ket), O::PEPSSandwich)
+    k = ket(O)
+    ket_pullback(Δk) = NoTangent(), (unthunk(Δk), zerovector(bra(O)))
+    return k, ket_pullback
+end
+function ChainRulesCore.rrule(::typeof(bra), O::PEPSSandwich)
+    b = bra(O)
+    bra_pullback(Δb) = NoTangent(), (zerovector(ket(O)), unthunk(Δb))
+    return b, bra_pullback
+end
```
leburgel, line 79: I thought this was all I needed to make sure that adjoints of local `PEPSSandwich`s are properly accumulated, but we still get `Tangent{Any, <:PEPSSandwich}` in the `getindex` pullback for `InfiniteSquareNetwork` (see above).

If anyone has an idea what I'm missing, I'd be very happy to hear.

src/networks/tensors.jl
```diff
@@ -116,3 +124,6 @@ const PEPOTensor{S<:ElementarySpace} = AbstractTensorMap{<:Any,S,2,4}
 Base.rotl90(t::PEPOTensor) = permute(t, ((1, 2), (4, 5, 6, 3)))
 Base.rotr90(t::PEPOTensor) = permute(t, ((1, 2), (6, 3, 4, 5)))
 Base.rot180(t::PEPOTensor) = permute(t, ((1, 2), (5, 6, 3, 4)))
+
+physicalspace(t::PEPOTensor) = space(t, 1) ← space(t, 2)'
```
leburgel, line 128: It made sense to me to make this a `HomSpace` since a PEPO tensor really maps between physcal spaces, but I can make it a `ProductSpace` if this is too confusing.

leburgel (commented):
Left some 'review' comments by way of summary.

src/operators/infinitepepo.jl
```diff
@@ -3,13 +3,14 @@
...
-struct InfinitePEPO{T<:PEPOTensor} <: InfiniteSquareNetwork{T,3}
+struct InfinitePEPO{T<:PEPOTensor} <: InfiniteGridNetwork{T,3}
     A::Array{T,3}
 
     function InfinitePEPO(A::Array{T,3}) where {T<:PEPOTensor}
         # space checks
         for (d, w, h) in Tuple.(CartesianIndices(A))
-            space(A[d, w, h], 1) == space(A[d, w, _next(h, end)], 2)' ||
+            only(codomain(physicalspace(A[d, w, h]))) ==
+            only(domain(physicalspace(A[d, w, _next(h, end)]))) ||
```
lkdvos, line 13: This looks strange to me though, I would expect `physicalspace` of an operator to return the space it is acting on, not the actual homspace/product space mapping it

src/states/infinitepeps.jl
```diff
@@ -103,50 +111,32 @@ function InfinitePEPS(
...
+## InfiniteSquareNetwork interface
 
-    function getindex_pullback(Δpepstensor_)
-        Δpepstensor = unthunk(Δpepstensor_)
-        Δstate = zerovector(state)
-        Δstate[row, col] = Δpepstensor
-        return NoTangent(), Δstate, NoTangent(), NoTangent()
-    end
-    return pepstensor, getindex_pullback
+function InfiniteSquareNetwork(top::InfinitePEPS, bot::InfinitePEPS)
```
lkdvos, line 122:
```suggestion
function InfiniteSquareNetwork(top::InfinitePEPS, bot::InfinitePEPS=top)
```

src/states/infinitepeps.jl
```diff
@@ -103,50 +111,32 @@ function InfinitePEPS(
...
 end
 
-function ChainRulesCore.rrule(::Type{<:InfinitePEPS}, A::Matrix{T}) where {T<:PEPSTensor}
-    peps = InfinitePEPS(A)
-    function InfinitePEPS_pullback(Δpeps)
-        return NoTangent(), Δpeps.A
-    end
-    return peps, InfinitePEPS_pullback
-end
+InfiniteSquareNetwork(top::InfinitePEPS) = InfiniteSquareNetwork(top, top)
```
lkdvos, line 128:
```suggestion
```

src/algorithms/ctmrg/sparse_environments.jl
leburgel: I can definitely change `n` to `dir`.

Regarding the second argument being a direction and not a position, `virtualspace` can also be called on a network where the first arguments correspond to the position in the unit cell and the last argument is the direction. I think this is always well-defined, since we know what an index looks like for a 2D or 3D network (either a set of >1 integers or a `CartesianIndex`), so there's in principle no confusion with the single-integer argument call which is then a direction (and only works on local sandwich objects).

If we're worried about this being confusing, we could make `virtualspace` an exclusively internal thing, and only use calls with concrete directions like `north_virtualspace`. I think this is anyway what I would do (see also below), it was just easier to route all of this through `virtualspace` with an integer direction specifier since then I only need to define one method for every network and local sandwich.

src/algorithms/toolbox.jl
```diff
@@ -53,90 +53,140 @@ function cost_function(peps::InfinitePEPS, env::CTMRGEnv, O::LocalOperator)
...
-            env.edges[NORTH, rprev, c][χ2 D1; χ3] *
-            env.corners[NORTHEAST, rprev, cnext][χ3; χ4] *
-            env.corners[SOUTHEAST, r, cnext][χ4; χ5] *
-            env.edges[SOUTH, r, c][χ5 D1; χ6] *
-            env.corners[SOUTHWEST, r, cprev][χ6; χ1]
+"""
+    value(network::InfiniteSquareNetwork, env::CTMRGEnv)
+
+Return the value (per unit cell) of a given contractible network contracted using a given
+CTMRG environment.
```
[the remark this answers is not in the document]
leburgel: I remember, not exporting was the easiest solution but you're right that now might be a good time to actually address this. I actually like `network_value`. I think `contract` might be a bit similar to `leading_boundary` in what it implies, i.e. it might sound like it's actually doing things instead of just evaluating. And `evaluate` maybe sounds a bit generic, although it could also work in principle.

src/networks/local_sandwich.jl
```diff
@@ -0,0 +1,95 @@
...
+ket(O::PEPSSandwich) = O[1]
+bra(O::PEPSSandwich) = O[2]
+
+function virtualspace(O::PEPSSandwich, dir)
+    return virtualspace(ket(O), dir) ⊗ virtualspace(bra(O), dir)'
+end
+
+Base.rotl90(O::PEPSSandwich) = rotl90.(O)
+Base.rotr90(O::PEPSSandwich) = rotr90.(O)
+Base.rot180(O::PEPSSandwich) = rot180.(O)
```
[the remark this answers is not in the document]
leburgel: Oh, yes, did not think of this very much. I'll use different internal names for the operations (rotations and math) on local sandwiches and then call those from the implementation of the `Base` methods on the networks.

Do we prefer something like `rotl90_localsandwich` and `add_localsandwich` over the simplest `_rotl90` and `_add` to make it obvious these things would need to be implemented for custom local objects?

src/networks/tensors.jl
leburgel: I think I like keeping `physicalspace` like this, but adding `domain_physicalspace` and `codomain_physicalspace`. I can then also implement these for `InfinitePEPO`, which is handy for initializing purposes.

src/operators/infinitepepo.jl
leburgel: How about I use `domain_physicalspace` and `codomain_physicalspace` here? Just having `physicalspace` return a single space is not really sensible I think, since it often happens that the physical domain and codomain are different for symmetric multilayer PEPOs.

src/algorithms/ctmrg/sequential.jl
```diff
@@ -30,58 +30,60 @@ function SequentialCTMRG(;
     )
 end
 
-function ctmrg_iteration(state, env::CTMRGEnv, alg::SequentialCTMRG)
-    truncation_error = zero(real(scalartype(state)))
-    condition_number = zero(real(scalartype(state)))
+function ctmrg_iteration(
+    network::InfiniteSquareNetwork, env::CTMRGEnv, alg::SequentialCTMRG
```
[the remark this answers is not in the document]
leburgel: I added a generic unrestricted
```julia
function ctmrg_iteration(state, env, alg::CTMRGAlgorithm)
    return ctmrg_iteration(InfiniteSquareNetwork(state), env, alg)
end
```
to convert anything to a contractible network so we can still call `ctmrg_iteration` using a PEPS or partition function as input. Given this, I think it's fair to restrict here, no?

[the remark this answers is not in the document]
leburgel: Alternatively, we can make the `ctmrg_iteration` implementation generic, and make the conversion methods catch specific types of input states to convert. Maybe that's better.

src/algorithms/ctmrg/sequential.jl
```diff
@@ -30,58 +30,60 @@ function SequentialCTMRG(;
...
-    sequential_projectors(coordinate::NTuple{3,Int}, state::InfinitePEPS, env::CTMRGEnv, alg::ProjectorAlgorithm)
+    sequential_projectors(col::Int, network::InfiniteSquareNetwork, env::CTMRGEnv, alg::ProjectorAlgorithm)
+    sequential_projectors(coordinate::NTuple{3,Int}, network::InfiniteSquareNetwork, env::CTMRGEnv, alg::ProjectorAlgorithm)
 
 Compute CTMRG projectors in the `:sequential` scheme either for an entire column `col` or
 for a specific `coordinate` (where `dir=WEST` is already implied in the `:sequential` scheme).
 """
 function sequential_projectors(
-    col::Int, state::InfiniteSquareNetwork, env::CTMRGEnv, alg::ProjectorAlgorithm
+    col::Int, network::InfiniteSquareNetwork, env::CTMRGEnv, alg::ProjectorAlgorithm
```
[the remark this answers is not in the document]
leburgel: We don't really, I can drop the restriction if you want.

src/algorithms/ctmrg/sequential.jl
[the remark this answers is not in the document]
lkdvos: I don't think additional conversion methods are really required, these are all really just internal functions, and I just don't see the need to restrict the types. If you want to call these functions manually, it should be up to you to ensure that the data supports the required interface.
Additionally, most of these functions actually don't use any interface for `network` at all, and simply pass them through.
In that sense, I would say that as long as we are not accessing fields or similar data-specific implementations, there really is no need to restrict this.

src/networks/tensors.jl
```diff
@@ -116,3 +124,8 @@ const PEPOTensor{S<:ElementarySpace} = AbstractTensorMap{<:Any,S,2,4}
 Base.rotl90(t::PEPOTensor) = permute(t, ((1, 2), (4, 5, 6, 3)))
 Base.rotr90(t::PEPOTensor) = permute(t, ((1, 2), (6, 3, 4, 5)))
 Base.rot180(t::PEPOTensor) = permute(t, ((1, 2), (5, 6, 3, 4)))
+
+domain_physicalspace(t::PEPOTensor) = space(t, 2)'
+codomain_physicalspace(t::PEPOTensor) = space(t, 1)
+physicalspace(t::PEPOTensor) = codomain_physicalspace(t) ← domain_physicalspace(t)
```
lkdvos, line 130: This still doesn't make that much sense to me, why is the `physicalspace` the combination of in and out?

src/algorithms/ctmrg/sequential.jl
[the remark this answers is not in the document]
leburgel: I see. I added the conversion here simply because the tests use `ctmrg_iteration` directly and this was the quickest way to unbreak them. But it makes sense to make everything other than `leading_boundary` internal, and then add the extra layer in the test. I'll change this and remove the type restriction.

src/networks/tensors.jl
leburgel: Just because in and out might not necessarily the same (especially in the context of multilayer things, see also comment above). I agree that the `HomSpace` might be too much, but I does seem sensible for `physicalspace` to return some info on both the incoming and outgoing space.

If this clashes too much with the MPSKit convention then I would even prefer to remove `physicalspace` for PEPOs and always explicitly use domain and codomain physicalspace. Having `physicalspace` return a single space seems very confusing to me.

src/networks/infinitegridnetwork.jl
pbrehmer: I'd also be okay with removing the supertype! Also, the OptimKit compatibility is probably not necessary for all potential subtypes so I hope the code duplication won't be too bad.

src/environments/ctmrg_environments.jl
pbrehmer: I agree, we never use the `Int` constructors of `CTMRGEnv`s anyway. They really only matter for the state types.

src/networks/infinitesquarenetwork.jl
pbrehmer: I would think that when you work with unit cells the matrix-based constructor would become relevant, right? But probably it doesn't need to be differentiated so I'd be okay with removing it for now if we anyway don't have any reasonable application/test

src/algorithms/toolbox.jl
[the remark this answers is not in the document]
pbrehmer: I think I would prefer `network_value` here

src/algorithms/ctmrg/ctmrg.jl
pbrehmer: I would definitely think it's okay to remove this. As far as I know this is not used anywhere and you're right, starting with a product environment seems like a relatively specific default argument that might incur problems depending on the CTMRG settings

src/algorithms/ctmrg/sparse_environments.jl
leburgel: I updated things to always use things like `north_virtualspace` in actual routines, and only use `virtualspace` to route all the different virtual space accessors through a single method to be overloaded. So things like `virtualspace(O, r, c, dir)` are never really called directly, it would always be something like `north_virtualspace(O, r, c)` where the additional arguments really specify a position like we're used to.

Is this sufficient? I'm kind of okay with it as is, but I can also make it more obvious that it's internal by renaming it to `_virtualspace`, or even get rid of it altogether and always directly overload the methods for all four directions if we think that's better.

src/networks/infinitesquarenetwork.jl
leburgel: I updated things to not use any custom rrules for component getters like `ket` and `bra`, and I can make things work by properly unthunking components and then converting the tangents here. I think this is cleaner than hooking into all the getters to make sure there's no tangent types in the first place, so probably things are just fine like this.

src/networks/tensors.jl
leburgel: I added `domain_physicalspace` and `codomain_physicalspace` and am quite happy with these. There's still the issue of what to do with `physicalspace` then, but resolving here in favor of the comment further below.

src/algorithms/ctmrg/sequential.jl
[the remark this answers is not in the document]
leburgel: I dropped the type restriction and added an explicit `InfiniteSquareNetwork` call whenever we were using `ctmrg_iteration` directly on an `InfinitePEPS` previously. This was mainly in the tests, but also in the fixed-point gradient code.

The way things are set up in `fixedpoint` now we really call `leading_boundary` directly on a state which now has an implicit conversion to `InfiniteSquareNetwork`, and so the `_rrule` really expects a PEPS as input and I had to add a conversion in the function handles we use to compute the fixed-point gradient. Do we think it's okay like this, or do we prefer to pull the conversion out explicitly and implement the pullback with respect to the `InfiniteSquareNetwork` in the `_rrule`?

src/algorithms/ctmrg/sequential.jl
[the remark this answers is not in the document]
leburgel: Dropped the type restriction.

src/networks/local_sandwich.jl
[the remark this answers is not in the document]
leburgel: I (think I) removed all of the piracy for tuples by switching to custom internal methods in the rotations and math for `InfiniteSquareNetwork`. Not too sure about the names though, so would be happy if anyone has an opinion on that.

src/operators/infinitepepo.jl
leburgel: I updated the space checks to explicitly use `domain_physicalspace` and `codomain_physicalspace`. That should solve the issue here. Resolving this and moving the discussion on what to do with `physicalspace` further down.

src/networks/tensors.jl
leburgel: In the end all the accessors for physical PEPO spaces explicitly use either `domain_physicalspace` and `codomain_physicalspace`. Maybe it's best to just remove `physicalspace` then, since it's not entirely clear what to do with this. Like I said, I personally prefer removing it altogether over having `physicalspace` return a single space. What do we think?

src/networks/infinitesquarenetwork.jl
```diff
@@ -1,67 +1,84 @@
...
+
+virtualspace(n::InfiniteSquareNetwork, r::Int, c::Int, dir) = virtualspace(n[r, c], dir)
+
 ## Vector interface
-function VectorInterface.scalartype(::Type{NWType}) where {NWType<:InfiniteSquareNetwork}
-    return scalartype(eltype(NWType))
-end
-function VectorInterface.zerovector(A::NWType) where {NWType<:InfiniteSquareNetwork}
-    return NWType(zerovector(unitcell(A)))
+VectorInterface.scalartype(::Type{<:InfiniteSquareNetwork{O}}) where {O} = scalartype(O)
```
lkdvos, line 56:
```suggestion
VectorInterface.scalartype(::Type{T}) where {T<:InfiniteSquareNetwork} = scalartype(eltype(T))
```

src/networks/infinitesquarenetwork.jl
```diff
@@ -1,67 +1,84 @@
...
+    return InfiniteSquareNetwork(zerovector(unitcell(A)))
 end
 
-## Math
+## Math (for Zygote accumulation)
+
 function Base.:+(A₁::NWType, A₂::NWType) where {NWType<:InfiniteSquareNetwork}
-    return NWType(unitcell(A₁) + unitcell(A₂))
+    return NWType(_add_localsandwich.(unitcell(A₁), unitcell(A₂)))
 end
```
lkdvos, line 65: This signature is a bit strange because it would only allow things that have the exact same type (for example different scalartypes would fail. Maybe better:
```julia
Base.:+(A1::InfiniteSquareNetwork, A2::InfiniteSquareNetwork) = InfiniteSquareNetwork(_add_localsandwich.(unitcell(A1), unitcell(A2)))
```

I would also do the same for the methods below

src/networks/infinitesquarenetwork.jl
```diff
@@ -1,67 +1,84 @@
...
 function Base.:-(A₁::NWType, A₂::NWType) where {NWType<:InfiniteSquareNetwork}
-    return NWType(unitcell(A₁) - unitcell(A₂))
+    return NWType(_subtract_localsandwich.(unitcell(A₁), unitcell(A₂)))
 end
 function Base.:*(α::Number, A::NWType) where {NWType<:InfiniteSquareNetwork}
```
lkdvos, line 69: Maybe also worth adding:
```julia
Base.:*(A::InfiniteSquareNetwork, α::Number) = α * A
```

src/networks/infinitesquarenetwork.jl
```diff
@@ -74,42 +91,50 @@ function Base.isapprox(A₁::InfiniteSquareNetwork, A₂::InfiniteSquareNetwork;
...
-    return A
+## Chainrules
+
+function ChainRulesCore.rrule(
+    ::typeof(Base.getindex), network::InfiniteSquareNetwork, args...
+)
+    O = network[args...]
+
+    function getindex_pullback(ΔO_)
+        ΔO = map(unthunk, ΔO_)
```
lkdvos, line 113: this looks a bit strange to me, what if `args... === i::Int`? or what if `args... === Colon()`? Do these methods work the same for both? I would somehow expect that this method only needs to be defined for the scalar getindex?

This being said, since we implement `getindex` through `getindex(unitcell())`, a `rrule` for `unitcell` should suffice, and then this could also be deleted?

src/networks/infinitesquarenetwork.jl
```diff
@@ -74,42 +91,50 @@ function Base.isapprox(A₁::InfiniteSquareNetwork, A₂::InfiniteSquareNetwork;
...
+function ChainRulesCore.rrule(::typeof(rotl90), network::InfiniteSquareNetwork)
+    network´ = rotl90(network)
+    function rotl90_pullback(Δnetwork_)
+        Δnetwork = unthunk(Δnetwork_)
+        return NoTangent(), rotr90(Δnetwork)
+    end
+    return network´, rotl90_pullback
 end
 
-## FiniteDifferences vectorization
```
lkdvos, line 108: This might be convenient to keep around?

src/networks/local_sandwich.jl
```diff
@@ -0,0 +1,71 @@
...
+
+## PartitionFunction
+
+_rotl90_localsandwich(O::PFTensor) = rotl90(O)
+_rotr90_localsandwich(O::PFTensor) = rotr90(O)
+_rot180_localsandwich(O::PFTensor) = rot180(O)
+
+_add_localsandwich(O1::PFTensor, O2::PFTensor) = O1 + O2
+_subtract_localsandwich(O1::PFTensor, O2::PFTensor) = O1 - O2
+_mul_localsandwich(α::PFTensor, O::PFTensor) = α * O
```
lkdvos, line 29:
```suggestion
_mul_localsandwich(α::Number, O::PFTensor) = α * O
```

src/networks/tensors.jl
lkdvos: I think for me the options are either of the following:
1. `codomain_physicalspace(t) == domain_physicalspace(t) ? codomain_physicalspace(t) : error()`
2. *deleted*

I'm expecting that we are going to run into a similar issue in MPSKit by the way, where I was leaning towards implementing the first for all "operator types".
It's slightly unfortunate that we have directions everywhere, but here we are using the `domain` and `codomain` terminology. Would it still make sense to replace that with `top_physicalspace` and `bottom_physicalspace`? Or, make aliases for that? Since we are introducing this now, I would rather think it through now instead of being stuck with it later...
I'm definitely also okay with keeping it as is, if we agree to that, just asking for opinions here. (and keeping in mind that I will push this through to MPSKit)

src/operators/infinitepepo.jl
```diff
@@ -3,26 +3,32 @@
 
 Represents an infinite projected entangled-pair operator (PEPO) on a 3D cubic lattice.
 """
-struct InfinitePEPO{T<:PEPOTensor} <: InfiniteSquareNetwork{T,3}
+struct InfinitePEPO{T<:PEPOTensor}
     A::Array{T,3}
```
lkdvos, line 8: Can you also add an "unchecked" constructor `InfinitePEPO{T}(A::Array{T,3})` here?

src/states/infinitepeps.jl
```diff
@@ -103,50 +107,128 @@ function InfinitePEPS(
...
+## InfiniteSquareNetwork interface
 
-    function getindex_pullback(Δpepstensor_)
-        Δpepstensor = unthunk(Δpepstensor_)
-        Δstate = zerovector(state)
-        Δstate[row, col] = Δpepstensor
-        return NoTangent(), Δstate, NoTangent(), NoTangent()
+function InfiniteSquareNetwork(top::InfinitePEPS, bot::InfinitePEPS=top)
+    size(top) == size(bot) ||
+        throw(ArgumentError("Top PEPS, bottom PEPS and PEPO rows should have length"))
```
lkdvos, line 145:
```suggestion
        throw(ArgumentError("Top PEPS, bottom PEPS and PEPO rows should have the same length"))
```

src/states/infinitepeps.jl
```diff
@@ -103,50 +107,128 @@ function InfinitePEPS(
...
+
+## Math
+
+function Base.:+(A₁::InfinitePEPS{T}, A₂::InfinitePEPS{T}) where {T}
+    return InfinitePEPS(unitcell(A₁) + unitcell(A₂))
+end
+function Base.:-(A₁::InfinitePEPS{T}, A₂::InfinitePEPS{T}) where {T}
+    return InfinitePEPS(unitcell(A₁) - unitcell(A₂))
+end
+Base.:*(α::Number, A::InfinitePEPS) = InfinitePEPS(α * unitcell(A))
```
lkdvos, line 164: Also `Base.*(A::InfinitePEPS, α::Number)`?

src/algorithms/ctmrg/sequential.jl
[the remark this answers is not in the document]
leburgel: I really think I did, there shouldn't be any more restricted `ctmrg_iteration` definitions after [this commit](https://github.com/QuantumKitHub/PEPSKit.jl/pull/131/commits/6f6e8250d94f13c93e71594c84af0522acf4ee14)?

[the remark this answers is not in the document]
lkdvos: My bad, I didn't realize that the comment stays with the outdated file on my screen. Apologies!

src/networks/tensors.jl
leburgel: I personally think using `domain` and `codomain` for physical spaces is okay. Aside from the fact that using `top` and `bottom` might be confused with north and south at some point, I think it's also not ideal to add a convention that domain is 'top' and codomain is `bottom`. Of course some kind of graphical convention is necessary in implementations, but in principle we should be able to reason about these operators without using top and bottom so long as we don't go into the actual contractions. But most importantly I don't think it's impossible that we ourselves are going to be about which of top or bottom is domain and codomain again at some point.

So I think using directions for all the virtual things, and more proper domain and codomain for actual physcical spaces of states and operators is not really a bad thing. I would also be okay with adopting this in MPSKit, using left and right for virtual spaces but using domain and codomain for physical ones really doesn't seem off-putting to me.

leburgel: Ah, but as to the actual suggestion, I like the first.

src/networks/infinitesquarenetwork.jl
leburgel: I kept it around for `InfinitePEPS`, which is what this was actually implemented for in the first place (but since I repurposed the old supertype name for a new thing it shows up as removed...). Do you think it's useful for the contractible network type? I would assume we only ever optimize over a physicsy network. Of course I can also add it for contractible networks, but I don't immediately see a clear use case.

src/networks/infinitesquarenetwork.jl
leburgel: Yes, we only really need the scalar getindex, I'll specialize this to avoid issues.

We could just go through a `unitcell` `rrule`, but out of the box this would get in a matrix of weird tangents that I would have to convert before being able to construct an `InfiniteSquareNetwork` from this. I can give this a shot, but at first sight I don't know if that's going to be actually cleaner than doing it this way.

src/networks/tensors.jl
leburgel: I implemented your first suggestion, I quite like it so I would be okay with doing it like this systematically.

lkdvos:
Is there a reason to wrap the partition function tensors in a tuple? Why can we not just use the tensors directly?

leburgel:
> Is there a reason to wrap the partition function tensors in a tuple? Why can we not just use the tensors directly?

No, no real reason. The 'reason' I changed this is because in playing around with AD for partition functions and PEPOs I was having issues with the `getindex` pullback for `InfiniteSquareNetwork`. This was getting in all kinds `Tangent`s of nested `Thunk`s with type signatures that were tricky to get a hold of. Implementing all the local sandwiches in a flat iterable allowed me to just use `ΔO = map(unthunk, ΔO_)` in the pullback which fixed all problems I was having. While this was most important for the PEPO sandwiches, I quite liked the idea of having exactly the same implementation for all the local sandwich manipulations by assuming they're flat iterables, so I went ahead an did the same using a (trivial) partition function sandwich.

So this is not necessary at all, it's more that I did not see the harm in it. But maybe it's unnecessarily complicating things. I can switch back to using just the the tensor, if I add an extra `_unthunk_localsandwich` that works for each sandwich type.

src/networks/infinitesquarenetwork.jl
lkdvos: fair enough

leburgel:
I switched the partition function `InfiniteSquareNetwork` to just use a `PFTensor` eltype, and adapted the local interface and getindex pullback to this specialized case.

lkdvos (approved):
For me this is okay to merge.

PR #132, opened 2025-02-14
Refactor boundary MPS contractions

leburgel:
Refactors the boundary MPS contractions to make use of `@autoopt` and the corresponding index labeling conventions. I also added an actual test for the multi-layer PEPO case, seems this was just not there before for some reason.

### DIFF
--- a/src/operators/infinitepepo.jl
+++ b/src/operators/infinitepepo.jl
@@ -111,7 +111,9 @@ TensorKit.space(T::InfinitePEPO, i, j) = space(T[i, j, end], 1)
 function initializePEPS(
     T::InfinitePEPO{<:PEPOTensor{S}}, vspace::S
 ) where {S<:ElementarySpace}
-    Pspaces = map(I -> space(T, I.I...), eachindex(IndexCartesian(), T))
+    Pspaces = map(Iterators.product(axes(T, 1), axes(T, 2))) do (r, c)
+        return space(T, r, c)
+    end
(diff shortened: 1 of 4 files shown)


src/operators/infinitepepo.jl
```diff
@@ -111,10 +111,8 @@ TensorKit.space(T::InfinitePEPO, i, j) = space(T[i, j, end], 1)
 function initializePEPS(
     T::InfinitePEPO{<:PEPOTensor{S}}, vspace::S
 ) where {S<:ElementarySpace}
-    Pspaces = Array{S,2}(undef, size(T, 1), size(T, 2))
-    for i in axes(T, 1)
-        j in axes(T, 2)
-        Pspaces[i, j] = space(T, i, j)
+    Pspaces = map(Iterators.product(axes(T, 1), axes(T, 2))) do (r, c)
+        return space(T, r, c)
     end
```
lkdvos, line 116:
```suggestion
    Pspaces = map(I -> space(T, I.I...), eachindex(IndexCartesian(), T))
```

lkdvos (dismissed):
Looks good to me!

src/operators/infinitepepo.jl
leburgel: Sorry, wasn't paying attention when I commited the suggestion, but I specifically only want to iterate over the first 2 or the three axes so I don't think this will work?

lkdvos (approved):

PR #134, opened 2025-02-17
`CTMRG` support for PEPS-PEPO-PEPS networks

leburgel:
Implements all the contractions required to run CTMRG for PEPO sandwiches with an arbitrary number of layers.

TODO:
- [x] Add a test with a proper numerical check on the outcome

### DIFF
--- a/src/algorithms/contractions/ctmrg_contractions.jl
+++ b/src/algorithms/contractions/ctmrg_contractions.jl
@@ -310,14 +310,28 @@ Alternatively, contract the environment with a vector `x` acting on it
 
 or contract the adjoint environment with `x`, e.g. as needed for iterative solvers.
 """
-function half_infinite_environment(
-    quadrant1::AbstractTensorMap{T,S,3,3}, quadrant2::AbstractTensorMap{T,S,3,3}
-) where {T,S}
-    return @autoopt @tensor env[χ_in D_inabove D_inbelow; χ_out D_outabove D_outbelow] :=
-        quadrant1[χ_in D_inabove D_inbelow; χ D1 D2] *
-        quadrant2[χ D1 D2; χ_out D_outabove D_outbelow]
+@generated function half_infinite_environment(
+    quadrant1::AbstractTensorMap{T,S,N,N}, quadrant2::AbstractTensorMap{T,S,N,N}
+) where {T,S,N}
+    env_e = tensorexpr(
+        :env,
+        (envlabel(:out), ntuple(i -> virtuallabel(:out, i), N - 1)...),
+        (envlabel(:in), ntuple(i -> virtuallabel(:in, i), N - 1)...),
+    )
+    quadrant1_e = tensorexpr(
+        :quadrant1,
+        (envlabel(:out), ntuple(i -> virtuallabel(:out, i), N - 1)...),
+        (envlabel(:NC), ntuple(i -> virtuallabel(:NC, i), N - 1)...),
+    )
+    quadrant2_e = tensorexpr(
+        :quadrant2,
+        (envlabel(:NC), ntuple(i -> virtuallabel(:NC, i), N - 1)...),
+        (envlabel(:in), ntuple(i -> virtuallabel(:in, i), N - 1)...),
+    )
+    return macroexpand(
+        @__MODULE__, :(return @autoopt @tensor $env_e := $quadrant1_e * $quadrant2_e)
+    )
 end
-
 function half_infinite_environment(
     C_1, C_2, E_1, E_2, E_3, E_4, A_1::P, A_2::P
 ) where {P<:PEPSSandwich}
@@ -364,12 +378,6 @@ function half_infinite_environment(
         conj(E_3[χ4 D7 D8; χ5]) *
         conj(C_2[χ5; χ6]) *
         conj(E_4[χ6 D9 D10; χ_in])
-end
-function half_infinite_environment(
-    quadrant1::AbstractTensorMap{T,S,2,2}, quadrant2::AbstractTensorMap{T,S,2,2}
-) where {T,S}
-    return @autoopt @tensor env[χ_in D_in; χ_out D_out] :=
-        quadrant1[χ_in D_in; χ D1] * quadrant2[χ D1; χ_out D_out]
 end
 function half_infinite_environment(
     C_1, C_2, E_1, E_2, E_3, E_4, A_1::P, A_2::P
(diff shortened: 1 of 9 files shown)


leburgel:
I just blindly implemented all the contractions in the `ctmrg_contractions.jl` file, but it seems like quite a few of the methods are never used. Maybe this would also be a good point to do some housekeeping with regard to what we should keep around and what not?

pbrehmer:
Looking at the code coverage, it seems that some contractions in `ctmrg_contractions.jl` got obsolete by adding the sparse `EnlargedCorner` struct which handles the row-column indices such that the row-column contraction methods are not used anymore (`enlarge_northwest_corner` and so on). In my opinion we can get rid of them.

Some of the contractions (`left_projector`, `right_projector`, some `half_infinite_environment`, `renormalize_northwest_corner`, ...) are not yet tested because sparse CTMRG (never explicitly constructing enlarged corners and full environments) is not yet fully implemented due to the complicated KrylovKit/TensorKit SVD interplay. But the building blocks are already there and the rest will be added in the future.

And for some contractions I don't yet get why codecov is flagging them as untested, so perhaps these are false flags or I need to dig deeper?

leburgel:
Removing the row-column-style enlarged corner methods seems good, it seems unlikely we would go back to using these. For the sparse CTMRG contractions, I guess it's fine to leave them as is then and reassess coverage once the things are actually implemented. Otherwise we would essentially be commenting them now and uncommenting them later, which is also a bit silly.

leburgel:
In terms of timeline, when do you think it would be good to try and add this @lkdvos @pbrehmer? It's not really urgent, but it's also quite orthogonal to most other things being worked on so I wouldn't mind getting this in sooner rather than later. The reason I'm asking is because I was hoping to update some bits to work with fermions, starting with the boundary MPS code, and it would be more efficent to have this in before I start that.

pbrehmer:
It does seem that this PR is relatively decoupled from the rest and that it might be finished soon? So I would be fine with still getting this in for v0.5 (in roadmap terms). Actually, it would be nice to have a little PEPO example in the docs anyway. If you think the PR can be realistically finished soon, then go for it, I'd say :-) But that also of course depends on the review capacities since we have quite a few PRs in the pipeline...

leburgel:
From my end I think this is finished and ready for review, I was more asking about review capacity since there's a lot of things going on. I would have liked to improve coverage, but since there's a lot of contractions that will only be used when sparse CTMRG is implemented that's not really possible right now I think. I also don't feel like leaving out those contractions now and implementing them later, just because it might take me some time to remember how these worked as it's a bit involved in places :)

The test can easily be made into an example for the documentation when we get to adding more examples, I quite like it.

pbrehmer:
Sounds great! I won't be able to take a look today but I could review tomorrow (if we want to)

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -310,14 +284,28 @@ Alternatively, contract the environment with a vector `x` acting on it
...
+    )
+    quadrant2_e = tensorexpr(
+        :quadrant2,
+        (envlabel(:NC), ntuple(i -> virtuallabel(:NC, i), N - 1)...),
+        (envlabel(:in), ntuple(i -> virtuallabel(:in, i), N - 1)...),
+    )
+    return macroexpand(
+        @__MODULE__, :(return @autoopt @tensor $env_e := $quadrant1_e * $quadrant2_e)
+    )
 end
```
lkdvos, line 308:
```suggestion
function half_infinite_environment(
    quadrant1::AbstractTensorMap{T,S,N,N}, quadrant2::AbstractTensorMap{T,S,N,N}
) where {T,S,N}
    p1 = (codomainind(quadrant1), domainind(quadrant1) .+ numout(quadrant1))
    p2 = (codomainind(quadrant2), domainind(quadrant2) .+ numout(quadrant2))
    p3 = (codomainind(quadrant1), domainind(quadrant2) .+ numout(quadrant1))
    return tensorcontract(quadrant1, p1, false, quadrant2, p2, false, p3)
end
```
not everything has to be generated :)

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -776,17 +770,29 @@ Apply projectors to each side of a quadrant.
...
-    return @autoopt @tensor corner[χ_in; χ_out] :=
-        P_right[χ_in; χ1 D1 D2] * quadrant[χ1 D1 D2; χ2 D3 D4] * P_left[χ2 D3 D4; χ_out]
-end
-function renormalize_corner(
-    quadrant::AbstractTensorMap{T,S,2,2}, P_left, P_right
-) where {T,S}
-    return @autoopt @tensor corner[χ_in; χ_out] :=
-        P_right[χ_in; χ1 D1] * quadrant[χ1 D1; χ2 D3] * P_left[χ2 D3; χ_out]
+@generated function renormalize_corner(
+    quadrant::AbstractTensorMap{T,S,N,N}, P_left, P_right
```
lkdvos, line 774: Can we be more restricted with the generated functions: 
```suggestion
    quadrant::AbstractTensorMap{<:Any,S,N,N}, P_left::AbstractTensorMap{<:Any,S,N,1}, P_right::AbstractTensorMap{<:Any,S,1,N}
```

or at least make it `AbstractTensorMap`?

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -1014,23 +1020,31 @@ Apply bottom projector to southwest corner and south edge.
...
-function renormalize_bottom_corner(
-    (row, col), env::CTMRGEnv{C,<:CTMRG_PF_EdgeTensor}, projectors
-) where {C}
-    C_southwest = env.corners[SOUTHWEST, row, _prev(col, end)]
-    E_south = env.edges[SOUTH, row, col]
-    P_bottom = projectors[1][row]
-    return @autoopt @tensor corner[χ_in; χ_out] :=
-        E_south[χ_in D1; χ1] * C_southwest[χ1; χ2] * P_bottom[χ2 D1; χ_out]
+@generated function renormalize_bottom_corner(
+    C_southwest, E_south::CTMRGEdgeTensor{T,S,N}, P_bottom
```
lkdvos, line 1030: Same comment about signature here.

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -1050,13 +1064,21 @@ function renormalize_top_corner((row, col), env::CTMRGEnv, projectors)
...
 end
-function renormalize_top_corner(C_northwest, E_north::CTMRG_PEPS_EdgeTensor, P_top)
-    return @autoopt @tensor corner[χ_in; χ_out] :=
-        P_top[χ_in; χ1 D1 D2] * C_northwest[χ1; χ2] * E_north[χ2 D1 D2; χ_out]
-end
-function renormalize_top_corner(C_northwest, E_north::CTMRG_PF_EdgeTensor, P_top)
-    return @autoopt @tensor corner[χ_in; χ_out] :=
-        P_top[χ_in; χ1 D1] * C_northwest[χ1; χ2] * E_north[χ2 D1; χ_out]
+@generated function renormalize_top_corner(
+    C_northwest, E_north::CTMRGEdgeTensor{T,S,N}, P_top
```
lkdvos, line 1068: Same comment about signature here

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -1465,3 +1493,840 @@ Multiply west right singular vectors with gauge signs from the left.
 function fix_gauge_west_right_vecs((row, col), V, signs)
     return signs[WEST, _prev(row, end), col] * V[WEST, row, col]
 end
+
+## WIP: PEPO contractions...
```
lkdvos, line 1497: No longer WIP?

src/algorithms/contractions/ctmrg_contractions.jl
lkdvos, line 1: Definitely fancy with all the generated expressions!

test/ctmrg/partition_function.jl
```diff
@@ -44,8 +43,9 @@ Implements the 2D classical Ising model with partition function
 \\mathcal{Z}(\\beta) = \\sum_{\\{s\\}} \\exp(-\\beta H(s)) \\text{ with } H(s) = -J \\sum_{\\langle i, j \\rangle} s_i s_j
 ```
 """
-function classical_ising(; beta=log(1 + sqrt(2)) / 2, J=1.0)
+function classical_ising(beta=log(1 + sqrt(2)) / 2, J=1.0)
```
lkdvos, line 46: I think not using the kwargs is a bit breaking with the convention we have for all other models. Does it matter?

test/runtests.jl
```diff
@@ -23,9 +23,6 @@ end
         @time @safetestset ":fixed CTMRG iteration scheme" begin
             include("ctmrg/fixed_iterscheme.jl")
         end
-        @time @safetestset "Unit cells" begin
-            include("ctmrg/unitcell.jl")
-        end
```
lkdvos, line 28: Was this on purpose?

test/ctmrg/pepo.jl
```diff
@@ -0,0 +1,139 @@
...
+
+    # check energy
+    n3_final = InfiniteSquareNetwork(psi_final, T)
+    m = PEPSKit.contract_local_tensor((1, 1, 1), M, n3_final, env3_final)
+    nrm3 = PEPSKit._contract_site((1, 1), n3_final, env3_final)
+
+    # compare to Monte-Carlo result from https://www.worldscientific.com/doi/abs/10.1142/S0129183101002383
+    @test abs(m / nrm3) ≈ 0.667162 rtol = 1e-2
+
+    # TODO: figure out issue with energy conventions...
```
lkdvos, line 138: Is this TODO done?

lkdvos (commented):
Left some small comments, otherwise I'm also okay with merging this.

test/ctmrg/partition_function.jl
leburgel: I was planning to try out computing some stat mech observables using AD by differentiating with respect to temperature, at which point I made this a normal argument. But then I didn't actually do that, so it doesn't matter here. If you prefer I can switch it back.

test/ctmrg/pepo.jl
leburgel: Yes, took a bit of fiddling but the test is trustworthy now.

src/algorithms/contractions/ctmrg_contractions.jl
leburgel: Not quite as bad as I anticipated, but still a bit of a pain :)

test/ctmrg/partition_function.jl
lkdvos: Maybe? It clearly also doesn't matter too much, I do like some consistency

lkdvos (dismissed):

lkdvos:
approved as soon as tests actually work :)

lkdvos (approved):

PR #135, opened 2025-02-18
Implement charge shift for `LocalOperator`

leburgel:
Implements `MPSKit.add_physical_charge` for `LocalOperator`, to be able to simulate Hamiltonians at fixed particle densities.

TODO:

- [ ] Add to model Hamiltonians ((Bose)-Hubbard)
- [ ] Test

### DIFF
--- a/src/operators/localoperator.jl
+++ b/src/operators/localoperator.jl
@@ -208,3 +208,65 @@ function Base.rot180(H::LocalOperator)
     )
     return LocalOperator(lattice2, terms2...)
 end
+
+# Charge shifting
+# ---------------
+
+TensorKit.sectortype(O::LocalOperator) = sectortype(typeof(O))
+TensorKit.sectortype(::Type{<:LocalOperator{T,S}}) where {T,S} = sectortype(S)
+
+@generated function _fuse_isomorphisms(
+    op::AbstractTensorMap{<:Any,S,N,N}, fs::Vector{<:AbstractTensorMap{<:Any,S,1,2}}
+) where {S,N}
+    op_out_e = tensorexpr(:op_out, -(1:N), -((1:N) .+ N))
+    op_e = tensorexpr(:op, 1:3:(3 * N), 2:3:(3 * N))
+    f_es = map(1:N) do i
+        j = 3 * (i - 1) + 1
+        return tensorexpr(:(fs[$i]), -i, (j, j + 2))
+    end
+    f_dag_es = map(1:N) do i
+        j = 3 * (i - 1) + 1
+        return tensorexpr(:(fs[$i]), -(N + i), (j + 1, j + 2))
+    end
+    multiplication_ex = Expr(
+        :call, :*, op_e, f_es..., map(x -> Expr(:call, :conj, x), f_dag_es)...
+    )
+    return macroexpand(@__MODULE__, :(return @tensor $op_out_e := $multiplication_ex))
+end
+
+"""
+Fuse identities on auxiliary physical spaces into a given operator.
+"""
+function _fuse_ids(op::AbstractTensorMap{T,S,N,N}, Ps::NTuple{N,S}) where {T,S,N}
+    # make isomorphisms
+    fs = map(1:N) do i
+        return isomorphism(fuse(space(op, i), Ps[i]), space(op, i) ⊗ Ps[i])
+    end
+    # and fuse them into the operator
+    return _fuse_isomorphisms(op, fs)
+end
+
+"""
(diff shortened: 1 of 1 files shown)


leburgel:
I think this is basically ready, but there's a few things I would like some input on:

- For `MPSKitModels.hubbard_model`: I wanted to add the fixed particle number `n` and opened up MPSKitModels.jl to adopt the convention used there, but it turns out the `n` argument is just never used in the [`MPSKitModels.hubbard_model` implementation](https://github.com/QuantumKitHub/MPSKitModels.jl/blob/b2c370e2335446e20e8025d901a92c13764e04a9/src/models/hamiltonians.jl#L308-L325). I imposed half filling in the test through a manual charge shift which covers the intended use case, but we really want to be able to pass a filling to the Hamiltonian constructor. I could fix this here, but probably it makes more sense to add this in MPSKitModels.jl first and then implement the same convention here in a follow-up?
- For `MPSKitModels.bose_hubbard_model`: exactly the same thing, but here the filling was implemented and I just kind of don't agree with the convention. Because of the way the auxiliary leg convention in MPSKit works, it seems to me that the auxiliary charge we add should be the dual of the physical filling `n` being passed to the Hamiltonian constructor. Therefore, I chose to shift by `U1Irrep(-n)` here as opposed to the `U1Irrep(n)` shift in MPSKitModels`. If the docstring says 'a fixed particle number density `n` can be imposed', I think shifting by `U1Irrep(-n)` such that the physical particle density is `n` is really the only sensible thing to do. If others agree, I can file an issue on this in MPSKitModels and change the convention there. If not, I'll flip it here, but that seems extremely confusing. Ultimately, this will be in line with the planned switch in auxiliary leg convention (see [here](QuantumKitHub/MPSKit.jl#260), so I don't think it can hurt to adopt this in the models already.

lkdvos:
I think I'm okay with solving both issues (here and MPSKitModels) simultaneously.
Definitely agree with conventions, also with implementing them.

leburgel:
I filed issues QuantumKitHub/MPSKitModels.jl#46 and QuantumKitHub/MPSKitModels.jl#47 on the problems in MPSKitModels.jl so we can track those separately. For the Bose-Hubbard model that solves things here.

For the Fermi-Hubbard model, there's some plumbing required to translate the integer particle number to an appropriate sector value for a given model symmetry. So I'm inclined to tackle that in MPSKitModels.jl first and then reuse that implementation to add the particle number here in a follow-up, since this is anyway a bit beyond the initially intended scope of this PR

To proceed, I was thinking of
- Finding an external benchmark value for the Bose-Hubbard energy and keep that as the charge shift test (since it's the fastest)
- Removing the XXZ and Fermi-Hubbard tests since they take too much time
- Add example stubs of the XXZ, Bose-Hubbard and Fermi-Hubbard runs currently included as tests, in preparation of some proper examples on imposing more involved internal symmetries and finite charge densities.

leburgel:
Okay, I got my hands on a cylinder-MPS result for the Bose-Hubbard test, added this one as a more narrated example, and moved the XXZ and Fermi-Hubbard tests to the examples folder as stubs.

I also added access and defaults for the linesearch maxiter and maxfg, since limiting this is really helpful for pushing through at the start of the optimization for more involved examples (in particular the Fermi-Hubbard one). That's kind of beyond the scope of this PR, but I set the defaults to the same ones as in OptimKit.jl, so none of the default behavior is changed by this.

If tests pass withing the timeout, I think this is pretty much ready to go.

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -74,6 +74,8 @@ function _OptimizationAlgorithm(;
     tol=Defaults.optimizer_tol,
     maxiter=Defaults.optimizer_maxiter,
     verbosity=Defaults.optimizer_verbosity,
+    ls_maxiter=Defaults.ls_maxiter,
+    ls_maxfg=Defaults.ls_maxfg,
     lbfgs_memory=Defaults.lbfgs_memory,
     # TODO: add linesearch, ... to kwargs and defaults?
```
lkdvos, line 80: Is this TODO now obsolete?

lkdvos (approved):
Looks extremely nice, the examples are really great.

src/algorithms/optimization/peps_optimization.jl
leburgel: Not really, there's still no way to really specify the other line search parameters (e.g. those that affect step acceptance) or pass in an actual line search algorithm instance. So I would leave this, but tackle it later when we actually get to solving our line search issues. For now, we can still pass in a concrete optimization algorithm where we set the line search algorithm parameters beforehand, so it would just be an extra convenience thing.

PR #140, opened 2025-02-23
Enforce `FixedSpaceTruncation` during differentiation

### LINKED ISSUE #138: Use `FixedSpaceTruncation` in CTMRG rrule
Currently, when differentiating through a `ctmrg_iteration` call in the fixedpoint differentiation we are not enforcing any `FixedSpaceTruncation` settings, which effectively means that we require a two-phase CTMRG algorithm if we want to allow for dynamic virtual bond dimensions.
I think this might be resolved by simply enforcing `FixedSpaceTruncation` [here](https://github.com/QuantumKitHub/PEPSKit.jl/blob/92a7711d51486ad9ee5b0b4241632d9ca8f285fc/src/algorithms/optimization/fixed_point_differentiation.jl#L119C32-L119C47)
Mentioned first #136

pbrehmer:
This fixes #138 by overwriting the `trscheme` to `FixedSpaceTruncation` in the CTMRG algorithm during differentiation. This is only necessary for `:diffgauge` since for the `:fixed` mode the SVD is anyway fixed and only accepts `NoTruncation`.

### DIFF
--- a/src/algorithms/optimization/fixed_point_differentiation.jl
+++ b/src/algorithms/optimization/fixed_point_differentiation.jl
@@ -111,13 +111,15 @@ function _rrule(
     alg::CTMRGAlgorithm,
 )
     env, info = leading_boundary(envinit, state, alg)
-    @reset alg.projector_alg.trscheme = FixedSpaceTruncation() # fix spaces during differentiation
+    alg_fixed = @set alg.projector_alg.trscheme = FixedSpaceTruncation() # fix spaces during differentiation
 
     function leading_boundary_diffgauge_pullback((Δenv′, Δinfo))
         Δenv = unthunk(Δenv′)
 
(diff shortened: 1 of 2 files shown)


src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -111,6 +111,7 @@ function _rrule(
     alg::CTMRGAlgorithm,
 )
     env, info = leading_boundary(envinit, state, alg)
+    @reset alg.projector_alg.trscheme = FixedSpaceTruncation() # fix spaces during differentiation
```
lkdvos, line 114: I'm not too familiar with `Accessors`, but can you make this create a new variable (with a new variable name) instead? This usually ruins the type stability of the closure below

PR #141, opened 2025-02-24
Pick out CTMRG left move as a separate function

Yue-Zhengyuan:
The left move of the sequential CTMRG is picked out as a separate function, which will be used in full update algorithm.


lkdvos (approved):
Looks like a reasonable change to me. Ok to merge if tests turn green.

PR #143, opened 2025-02-26
Wrap SU intermediate steps into functions

Yue-Zhengyuan:
This PR refactors the simple update, so that some parts can be reused by full update / neighborhood tensor update in the future.

### DIFF
--- a/src/algorithms/time_evolution/evoltools.jl
+++ b/src/algorithms/time_evolution/evoltools.jl
@@ -66,3 +66,76 @@ function _absorb_weight(
     end
     return t
 end
+
+"""
+Use QR decomposition on two tensors connected by a bond
+to get the reduced tensors
+```
+        2                   1
+        |                   |
+    5 - A ← 3   ====>   4 - X - 2   1 - a ← 3
+        | ↘                 |            ↘
+        4   1               3             2
+
+        2                               1
+        |                               |
+    5 ← B - 3   ====>   1 ← b - 3   4 - Y - 2
+        | ↘                  ↘          |
+        4   1                 2         3
+```
+"""
+function _qr_bond(A::PEPSTensor, B::PEPSTensor)
+    # TODO: relax dual requirement on the bonds
+    @assert isdual(space(A, 3)) # currently only allow A ← B
+    X, a = leftorth(A, ((2, 4, 5), (1, 3)))
+    b, Y = rightorth(B, ((5, 1), (2, 3, 4)))
+    # add twist if X → a / b → Y
+    isdual(space(a, 1)) && twist!(a, 1)
+    isdual(space(Y, 1)) && twist!(b, 3)
+    X = permute(X, (1, 4, 2, 3))
+    Y = permute(Y, (2, 3, 4, 1))
+    return X, a, b, Y
+end
+
+"""
+Reconstruct the tensors connected by a bond from their QR results
+obtained from `_qr_bond`
+```
+        -2                             -2
+        |                               |
+    -5- X - 1 - a - -3     -5 - b - 1 - Y - -3
+        |        ↘               ↘      |
+        -4        -1              -1   -4
+```
+"""
+function _qr_bond_undo(X::PEPSOrth, a::AbstractTensorMap, b::AbstractTensorMap, Y::PEPSOrth)
+    @tensor A[-1; -2 -3 -4 -5] := X[-2 1 -4 -5] * a[1 -1 -3]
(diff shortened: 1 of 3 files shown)


src/algorithms/time_evolution/evoltools.jl
```diff
@@ -66,3 +66,76 @@ function _absorb_weight(
...
+```
+"""
+function _qr_bond(A::PEPSTensor, B::PEPSTensor)
+    # TODO: relax dual requirement on the bonds
+    @assert isdual(space(A, 3)) # currently only allow A ← B
+    X, a = leftorth(A, ((2, 4, 5), (1, 3)))
+    b, Y = rightorth(B, ((5, 1), (2, 3, 4)))
+    # add twist if X → a / b → Y
+    isdual(space(a, 1)) && twist!(a, 1)
+    isdual(space(Y, 1)) && twist!(b, 3)
```
lkdvos, line 94: I don't think this can ever happen: `leftorth` etc should always create non-dual spaces between the factors

lkdvos (approved):
Overall looks good to me.
Something to consider in the future (separate PR) is the same as before: have you considered whether the "linear map" choices are the most convenient?
It seems like permuting everything to `TensorMap{N,0}` might lead to unnecessary permutations again.

src/algorithms/time_evolution/evoltools.jl
Yue-Zhengyuan: I thought so before, until once I encounter:
```julia
using TensorKit # v0.14.4
a = randn(ℂ^2 ← (ℂ^2)')
q, r = leftorth(a)
println(space(q))
println(space(r))
```
Output:
```
ℂ^2 ← (ℂ^2)'
(ℂ^2)' ← (ℂ^2)'
```
This happens when `numin(a) == 1`, and the domain is a dual space. I think `rightorth` also has this subtlety. I'm quite unhappy with this choice in TensorKit.

Yue-Zhengyuan: If you can help change this behavior of TensorKit, I can use this certainty on the "dual-ness" to avoid some twists.

Yue-Zhengyuan:
@lkdvos Or, is it ok if I just merge this one as it is and consider the choice of codomain/domain in the next PR on the construction of bond environment? That can provide more context on which choice is the best

src/algorithms/time_evolution/evoltools.jl
lkdvos: I agree, I’ll look into resolving that. For now, we can keep your fix. Thanks for bringing this to my attention!

lkdvos (approved):
I actually prefer it in separate PRs, as you can tell the small ones are a lot easier to review and get progress on :). 

Ok to merge for me.

src/algorithms/time_evolution/evoltools.jl
Yue-Zhengyuan: @lkdvos What's the progress on this now? It bothers me again when writing SU with next nearest neighbor interactions. Was there actually a reason behind the current design choice that sometimes produces a dual space between Q, R (or L, Q)?

Yue-Zhengyuan: I finally find a case (in the 3-site simple update #171) where the current TensorKit behavior is useful. When the produced R or L is a matrix, it is indeed convenient sometimes to keep its domain equal to codomain. I'm now satisfied 😄 I'll try explaining it in that PR in a comment or the docstring... the twists are driving me crazy

lkdvos: I seemed to have missed your previous comment here, apologies for that.
The *reason* behind this design actually is not so much of a choice, but rather an accidental result from how it is implemented.
The leg inbetween the factorized tensors typically can be constructed as `fuse(domain)` or `fuse(codomain)`. `fuse` by default results in a non-dual space, but whenever there is only one space, you could consider it to be the identity, therefore not necessarily having a non-dual space as output. (This was "fixed" in https://github.com/Jutho/TensorKit.jl/pull/111). However, a similar specialization was still in place in the factorizations themselves, see https://github.com/Jutho/TensorKit.jl/blob/ed09f138e41d8a69a703f6473e4acbb2838f2fb5/src/tensors/factorizations.jl#L343 for example. In principle we should go and fix this, although that definitely counts as a breaking change, and as you might imagine, I'm scared of it breaking code because of the twists... I still think it should be changed to be consistent, so if you want to make a PR for that definitely feel free to do so. If not, this will probably be changed in the next (breaking) release of TensorKit anyways, since we are reworking the factorizations: https://github.com/Jutho/TensorKit.jl/tree/matrixalgebra

PR #144, opened 2025-02-27
Add Neighborhood Tensor Update (NTU)

Yue-Zhengyuan:
(Finally, after being delayed for over a year)

----

This PR adds the Neighborhood Tensor Update (NTU, [arXiv 2107.06635](https://arxiv.org/abs/2107.06635)), which works for both iPEPS and iPEPO, and supports  nearest neighbor 2-site gates and arbitrary MPO gates (in particular, the next-nearest neighbor 3-site MPO).

One caveat to be emphasized is that after applying an N-site gate (N > 2), the updated bonds are truncated sequentially without going back. In other words, it does *not* perform an ALS optimization of all site tensors that are updated, since producing the environment surrounding a general cluster of sites is complicated, and the following ALS optimization will be extremely costly.

When truncating a bond after applying a time evolution gate, the NTU uses some neighboring tensors as the bond environment, instead of using only a few bond weights (simple update) of the full iPEPS approximated by CTM (full update), "interpolating" between the two extremes. Currently the NN, NN+ and NNN bond environments (see [description](https://yastn.github.io/yastn/fpeps/environment_ntu.html) in the YASTN package) are implemented.


Yue-Zhengyuan:
The NTU bondenvs can also be constructed for `InfinitePEPS` without Schmidt weights on bonds (this is the approach in YASTN). But I found that NTU works better (at least for Heisenberg model) if I include the weights. At $D = 4$ with `NTUEnvNN`, the NTU energy for `InfinitePEPS` is -0.66858; for `InfiniteWeightPEPS` it can be further improved a bit to -0.66885. 

I think the reason is quite simple. For `InfinitePEPS`, the `NTUEnvNN` bondenv is (I only draw the ket layer without physical axes; the open ends are contracted with the bra layer)
```
            |   |
        ----T---T----
            |   |
    ---T---()   ()---T---
            |   |
        ----T---T----
            |   |
```
For `InfiniteWeightPEPS` with bond weights, the `NTUEnvNN` bondenv is (assume the weights have been absorbed into the tensors)
```
                |       |
                √λ      √λ
                |       |
        ---√λ---T-------T---√λ---
                |       |
    ---√λ---T---()      ()---T---√λ---
                |       |
        ---√λ---T-------T---√λ---
                |       |
                √λ      √λ
                |       |
```
So the bondenv has an additional circle of $\sqrt{\lambda}$'s, which is intuitively a more accurate than the bondenv for `InfinitePEPS`. This is why I modified the NTU in YASTN to work with `InfiniteWeightPEPS` instead of `InfinitePEPS`.

Yue-Zhengyuan:
Edit: I'll soon add an option to control whether to further absorb the additional circle of weights when constructing the NTU bond environment.

Yue-Zhengyuan:
Some commits are based on #348, #349.

Yue-Zhengyuan:
I'll keep this as a draft for a while to not waste resources on CI.

src/algorithms/contractions/bondenv/benv_tools.jl
```diff
@@ -11,3 +18,198 @@ _prepare_site_tensor(t::PEPSTensor) = t
...
+    # axes to be contracted
+    hair_axs = Tuple(ax for ax in 2:5 if ax ∉ open_axs)
+    # attach hairs to ket
+    ket = twistdual(ket, 1)
+    axes = ntuple(identity, Val(5))
+    for (h, ax) in zip(hairs, hair_axs)
+        twistdual!(ket, ax)
+        h === nothing && continue
+        axes, biperm = _permute_to_first(axes, ax)
+        ket = permute(h, ((1,), (2,))) * permute(ket, biperm)
```
Yue-Zhengyuan, line 115: This is slightly unfortunate that I have to permute `h` every time, because I decide to put the bra legs in front of the ket legs, and all legs in the codomain (which was the result of using `ncon`).

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -176,17 +176,22 @@ end
...
 on the input state `psi` and its environment `env`.
+
+- Using `iter` and `t` to reset the current iteration number and evolved time
+    respectively of the TimeEvolver `it`.
 """
 function MPSKit.timestep(
-        it::TimeEvolver{<:SimpleUpdate}, psi::InfiniteState, env::SUWeight
+        it::TimeEvolver{<:SimpleUpdate}, psi::InfiniteState, env::SUWeight;
+        iter::Int = it.state.iter, t::Float64 = it.state.t
     )
```
Yue-Zhengyuan, line 186: I added the ability to reset iteration counter in `timestep` for simple update. It should have been there, but I don't know how I missed it.

src/algorithms/time_evolution/ntupdate3site.jl
```diff
@@ -0,0 +1,125 @@
...
+    else
+        @assert stype1 == :middle
+        bond_tensor_midnext(A; trunc = qrtrunc)
+    end
+    b, Y = if stype2 == :last
+        bond_tensor_last(B; trunc = qrtrunc)
+    else
+        @assert stype2 == :middle
+        bond_tensor_midprev(B; trunc = qrtrunc)
+    end
```
Yue-Zhengyuan, line 91: I may better explain why the middle sites need different treatment from the first/last sites when factoring out the reduced bond tensor.

Consider the 3-site gate MPO acting on PEPS. Its virtual bond dimension is $d^2$. For the first site, we know that we can first factor out a reduced bond tensor (carrying one physical leg) which is acted on by the gate.
```
    |
--- X ---  --- a ---
    |            ↘
```
The dimension of the leg connecting `X` and `a` is $Dd$, and number of elements (size) in truncated `a` (which is going to be optimized with ALS) is $D^2 d^2$.

But if we keep the physical leg in `X`,
```
    |
--- X ---  --- a ---
    | ↘
```
then after applying the gate, the leg connecting `X`, `a` has dimension $D d^2$, while the size of truncated `a` remains $D^2 d^2$. This increased the dimension of the bond environment surrounding `a`, thus being less efficient.

The same argument applies to the last site, for which it is also better to transfer the physical leg to the bond tensor `b` to reduce the dimension of the bond environment.

However, things are different for middle sites.
```
     ╱    
--- M --
  ╱ |     
    ↓     
--- g ---
    ↓     
```
After applying the gate, the cluster bond dimension increases to $D d^2$. To truncate the left (previous) bond, if we still transfer the physical leg to the bond tensor
```
                |
--- b ---   --- Y ---
      ↘         |
```
The leg connecting `b`, `Y` have dimension $D d^3$, and truncated `b` has size $D^2 d^4$. But if we keep the physical leg in `Y`,
```
                |
--- b ---   --- Y ---
                | ↘
```
then the leg connecting `b`, `Y` has a smaller dimension $D d^2$, with truncated `b` also having a smaller size $D^2 d^2$. This can make the bond truncation less costly.

There is also an important subtlety: for middle sites, the unitary tensors `X` or `Y` also contains the effect of the gate, in contrast to the first/last tensors, for which `X`, `Y` can be found prior to applying the gate. Thus, it seems that varying the bond tensors alone cannot reach the optimal truncation result except for NN gates.

---

Finally, a technical question on whether it is good to use symbols to distinguish the different cases.

src/algorithms/truncation/bond_tensor.jl
```diff
@@ -113,3 +113,101 @@ function undo_bond_tensor_last(b::MPSTensor, Y::PEPOOrth; gate_ax::Integer = 1)
...
+        | ↘          ↘
+        5   1         (2)
+```
+
+Here the physical leg on `a` is an auxiliary trivial leg.
+"""
+function bond_tensor_midnext(A::PEPSTensor; kwargs...)
+    X, a = left_orth!(permute(A, ((1, 2, 4, 5), (3,)); copy = true); kwargs...)
+    X = permute(X, ((1,), (2, 5, 3, 4)))
+    a = insertrightunit(a, 1)
```
Yue-Zhengyuan, line 143: For middle tensors, the physical leg is not transferred to the reduced bond tensor `a`. A trivial physical leg is added in order to fit the current `bond_truncate` code.

src/algorithms/time_evolution/ntupdate.jl
```diff
@@ -0,0 +1,279 @@
...
+        info0 = (; energy, env)
+        # start evolving
+        energy0, ΔE = energy, 0.0
+        iter0, t0 = it.state.iter, it.state.t
+        for (psi, info) in it
+            iter = it.state.iter
+            showinfo = (iter == 1) || (iter % check_interval == 0) || (iter == it.nstep)
+            !showinfo && continue
+            # bond weight change
+            Δλ = hasproperty(info0, :wts) ? compare_weights(info.wts, info0.wts) : NaN
```
Yue-Zhengyuan, line 236: Here the bond weights are simply the `bond_truncate` SVD spectra. They are not the same as those produced by BP gauge fixing. For the first iteration, since the initialization does not provide information on bond weights, `Δλ` is set to `NaN`.

If we incorporate BP gauge fixing in the future, the bond truncation spectra should be replaced by the diagonalized BP messages. In this case, we can also find `Δλ` for the first step by gauge fixing the initial state.

src/algorithms/time_evolution/ntupdate.jl
```diff
@@ -0,0 +1,279 @@
...
+Internal state of neighbourhood tensor update algorithm
+"""
+struct NTUState{S <: InfiniteState, N <: Number}
+    "number of performed iterations"
+    iter::Int
+    "evolved time"
+    t::N
+    "PEPS/PEPO"
+    psi::S
+end
```
Yue-Zhengyuan, line 41: If we incorporate BP gauge fixing, `NTUState` should also include the BPEnv. It will also be used to construct bond environments. But for now I just leave it out.

Yue-Zhengyuan (commented):
Some comments that may help with reviewing.

Yue-Zhengyuan:
@lkdvos Hope to get merged soon so I can proceed from here.

src/algorithms/contractions/bondenv/benv_ntu.jl
```diff
@@ -0,0 +1,239 @@
...
+"""
+abstract type NeighbourEnv end
+
+"""
+SVD with `truncrank(1)`.
+"""
+function _svd_cut!(t::AbstractTensorMap)
+    t1, s, t2 = svd_trunc!(t; trunc = truncrank(1))
+    return absorb_s(t1, s, t2)
+end
```
lkdvos, line 25: Out of curiosity: do you need the weight to be distributed equally? Otherwise this might be replaced with `left_orth!(t; trunc = truncrank(1))` or `right_orth!(t; trunc = truncrank(1))`, which I think removes some intermediate allocations.

src/algorithms/contractions/bondenv/benv_ntu.jl
```diff
@@ -0,0 +1,239 @@
...
+            ║       ║
+    0   ●===X==   ==Y===●
+            ║       ║
+    1       ●=======●
+        -1  0       1   2
+```
+"""
+function bondenv_ntu(
+        row::Int, col::Int, X, Y, state::S, alg::NNEnv
+    ) where {S <: InfiniteState}
```
lkdvos:
```suggestion
        row::Int, col::Int, X, Y, state::InfiniteState, alg::NNEnv
    )
```
slight cosmetic change, I don't think the `where` clause helps here

src/algorithms/contractions/bondenv/benv_ntu.jl
```diff
@@ -0,0 +1,239 @@
...
+    neighbors = [(-1, 0), (0, -1), (1, 0), (1, 1), (0, 2), (-1, 1)]
+    m = collect_neighbors(state, row, col, neighbors)
+    X, Y = _prepare_site_tensor(X), _prepare_site_tensor(Y)
+    # southwest half
```
lkdvos: Do you need these to be normalized separately? In principle, by putting the entire `benv` contraction into a single `@tensor` call we end up with slightly less intermediate permutations. For example, here the `benv_sw` and `benv_ne` will be permuted for the final contraction, and since these are the biggest objects this might actually matter quite a bit.

src/algorithms/contractions/bondenv/benv_ntu.jl
lkdvos: On a separate note, do we need this to be trivially charged for everything to work? It might reduce quite a bit of the number of legs if we can do something like:

```julia
A, B = left_orth!(t; trunc = truncspace(oneunit(spacetype(t))))
return removeunit(A, numind(A)), removeunit(B, 1)
```

src/algorithms/contractions/bondenv/benv_ntu.jl
```diff
@@ -0,0 +1,239 @@
...
+                1      2
+    =#
+    vecr = enlarge_corner_se(cor_se(m[1, 2]), edge_s(m[1, 1]), edge_e(m[0, 2]), Y)
+    @tensor vecr[:] :=
+        edge_n(m[-1, 1])[D11 D10 D21 D20 -1 -2] *
+        cor_ne(m[-1, 2])[D31 D30 D11 D10] *
+        vecr[D21 D20 D31 D30 -3 -4 -5 -6]
+    normalize!(vecr, Inf)
+    # combine left and right part
+    @tensor benv[-1 -2; -3 -4] := vecl[1 2 -1 -3 3 4] * vecr[1 2 -2 -4 3 4]
```
lkdvos: Same comment here about a single `@tensor` call removing intermediate permutations

src/algorithms/contractions/bondenv/benv_tools.jl
```diff
@@ -11,3 +18,197 @@ _prepare_site_tensor(t::PEPSTensor) = t
...
+    return benv_tensor(ket, bra, open_axs, ntuple(Returns(nothing), 4 - N))
+end
+function benv_tensor(
+        ket::PEPSTensor, bra::PEPSTensor,
+        open_vaxs::NTuple{N, Int}, hairs::NTuple{Nh, Union{Nothing, H}}
+    ) where {N, Nh, H <: Hair}
+    @assert 1 <= N <= 3 && Nh == 4 - N
+    open_axs = open_vaxs .+ 1
+    # axes to be contracted
+    hair_axs = Tuple(ax for ax in 2:5 if ax ∉ open_axs)
```
lkdvos: We might want to inspect the type stability on this: I think you want this to be guaranteed to be a `NTuple{4 - N, Int}`, since we now the `open_vaxs` are all in `1:4`. I'm not sure the compiler is smart enough to figure that out, which presumably affects the type stability of the remainder of this function.

src/algorithms/contractions/bondenv/benv_tools.jl
```diff
@@ -11,3 +18,197 @@ _prepare_site_tensor(t::PEPSTensor) = t
...
+    H_w -   E_w -    - ket -    - E_e   - H_e
+            |           |           |
+
+            |           |           |
+            C_sw -   - E_s -   - C_se
+
+                        |
+                       H_s
+=#
+const open_axs_hair = Dict(:n => (SOUTH,), :e => (WEST,), :s => (NORTH,), :w => (EAST,))
```
lkdvos: It looks slightly strange to me to have these as `const` and then use them (as far as I can tell?) only in the `@eval loops below. Why not immediately loop over a vector of the pairs? That also keeps the combinations together, making it slightly easier as a human to read the code :)
(Vector of pairs should even be slightly more compact than the dictionary, so also for a computer slightly easier to read)

src/algorithms/contractions/bondenv/benv_tools.jl
lkdvos: I assume we can leave this for follow-up optimizations?

src/algorithms/time_evolution/ntupdate3site.jl
lkdvos: I think symbols should be fine, the only other thing I can think of would be to create some `@enum` but I don't think it would matter in the end.

I think I follow the explanation, it seems reasonable that it is not possible to just get rid of the extra legs in the middle and that this doesn't buy us anything there.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -169,17 +170,22 @@ end
...
 Given the `TimeEvolver` iterator `it`, perform one step of time evolution
 on the input state `psi` and its environment `env`.
+
+- Using `iter` and `t` to reset the current iteration number and evolved time
+    respectively of the TimeEvolver `it`.
 """
 function MPSKit.timestep(
-        it::TimeEvolver{<:SimpleUpdate}, psi::InfiniteState, env::SUWeight
+        it::TimeEvolver{<:SimpleUpdate}, psi::InfiniteState, env::SUWeight;
+        iter::Int = it.state.iter, t::Float64 = it.state.t
```
lkdvos: Do you want to restrict to `t::Float64`, or is `t::Number` good enough?

lkdvos (commented):
Sorry for taking so long to get to this!

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: My bad, it should be `Number` as we previously discussed on evolution with complex time step.

src/algorithms/contractions/bondenv/benv_ntu.jl
Yue-Zhengyuan: > do you need the weight to be distributed equally?

Nice observation! Actually no, since this D = 1 leg is in the end contracted, so how we distribute `s` is irrelevant.

> do we need this to be trivially charged for everything to work?

It seems that `_svd_cut` is always applied to a positive map, which then (if my feeling is right) ensures that the leading singular value has trivial charge. I'll double-check this.

In YASTN they don't need to worry about it, since they only support Abelian symmetries and allow a tensor to have nonzero total charge.

src/algorithms/contractions/bondenv/benv_tools.jl
Yue-Zhengyuan: Added a TODO here.

src/algorithms/contractions/bondenv/benv_ntu.jl
Yue-Zhengyuan: For `_svd_cut!` I have updated its docstring, describing a sufficient condition under which the largest singular value in the charge-neutral sector is also the largest in the entire SVD spectrum. The condition is satisfied by tensors constructed from contracting a bra-ket pair of tensors in a twist-free way, so indeed keeping the largest charge-neutral singular value works just as fine as `truncrank(1)`.

Yue-Zhengyuan:
@lkdvos I think all comments in the last review are addressed. Ready for another round of review!

src/algorithms/contractions/bondenv/benv_tools.jl
Yue-Zhengyuan: Introduced a helper function `_hair_axes` for type stability.

Yue-Zhengyuan:
I'm now going to submit a paper for publication which says that NTU is implemented in PEPSKit. Hopefully we can finish this PR soon.

PR #146, opened 2025-03-03
Refactor `absorb_weight` and use `leftorth` in `_qr_bond`

Yue-Zhengyuan:
This PR refactors the function `absorb_weight` and replace the direction labels t, r, b, l (borrowed from YASTN) with the existing constants NORTH = 1, etc. 

In addition, in `_qr_bond` for time evolution, I found that using `leftorth` only appears to be more numerically stable for some reason. So I want to make this change as well. Another possible advantage is that using `leftorth` only can help produce a `bondenv` around `a`, `b` that is a linear map between "un-dualed" spaces, which can save some twists:
```
    ┌-----------------------┐
    |   ┌----┐              |
    └---|    |-← a === b →--┘
        |benv|   ↓     ↓
    ┌---|    |-→ a† == b† ←-┐
    |   └----┘              |
    └-----------------------┘
```

### DIFF
--- a/src/algorithms/time_evolution/simpleupdate.jl
+++ b/src/algorithms/time_evolution/simpleupdate.jl
@@ -114,10 +114,10 @@ end
 
 """
     simpleupdate(peps::InfiniteWeightPEPS, ham::LocalOperator, alg::SimpleUpdate;
-                 bipartite::Bool=false, check_int::Int=500)
+                 bipartite::Bool=false, check_interval::Int=500)
 
 Perform simple update with nearest neighbor Hamiltonian `ham`, where the evolution
-information is printed every `check_int` steps. 
+information is printed every `check_interval` steps. 
 
 If `bipartite == true` (for square lattice), a unit cell size of `(2, 2)` is assumed, 
 as well as tensors and x/y weights which are the same across the diagonals, i.e. at
@@ -128,7 +128,7 @@ function simpleupdate(
     ham::LocalOperator,
     alg::SimpleUpdate;
     bipartite::Bool=false,
-    check_int::Int=500,
+    check_interval::Int=500,
 )
     time_start = time()
     Nr, Nc = size(peps)
@@ -147,7 +147,7 @@ function simpleupdate(
         cancel = (count == alg.maxiter)
         wts0 = deepcopy(peps.weights)
         time1 = time()
-        if ((count == 1) || (count % check_int == 0) || converge || cancel)
+        if ((count == 1) || (count % check_interval == 0) || converge || cancel)
             @info "Space of x-weight at [1, 1] = $(space(peps.weights[1, 1, 1], 1))"
             label = (converge ? "conv" : (cancel ? "cancel" : "iter"))
             message = @sprintf(
--- a/src/algorithms/time_evolution/simpleupdate.jl
+++ b/src/algorithms/time_evolution/simpleupdate.jl
@@ -39,21 +39,16 @@ function _su_bondx!(
     A, B = peps.vertices[row, col], peps.vertices[row, cp1]
     sqrtsA = ntuple(dir -> (dir == EAST), 4)
     sqrtsB = ntuple(dir -> (dir == WEST), 4)
-    _allfalse, _alltrue = ntuple(Returns(false), 4), ntuple(Returns(true), 4)
-    A = _absorb_weights(A, peps.weights, row, col, Tuple(1:4), sqrtsA, _allfalse)
-    B = _absorb_weights(B, peps.weights, row, cp1, Tuple(1:4), sqrtsB, _allfalse)
+    A = _absorb_weights(A, peps.weights, row, col, Tuple(1:4), sqrtsA, false)
+    B = _absorb_weights(B, peps.weights, row, cp1, Tuple(1:4), sqrtsB, false)
     # apply gate
     X, a, b, Y = _qr_bond(A, B)
     a, s, b, ϵ = _apply_gate(a, b, gate, alg.trscheme)
     A, B = _qr_bond_undo(X, a, b, Y)
     # remove environment weights
-    _allfalse, _alltrue = _allfalse[1:3], _alltrue[1:3]
-    A = _absorb_weights(
-        A, peps.weights, row, col, (NORTH, SOUTH, WEST), _allfalse, _alltrue
-    )
-    B = _absorb_weights(
-        B, peps.weights, row, cp1, (NORTH, SOUTH, EAST), _allfalse, _alltrue
-    )
+    _allfalse = ntuple(Returns(false), 3)
+    A = _absorb_weights(A, peps.weights, row, col, (NORTH, SOUTH, WEST), _allfalse, true)
+    B = _absorb_weights(B, peps.weights, row, cp1, (NORTH, SOUTH, EAST), _allfalse, true)
     # update tensor dict and weight on current bond 
     # (max element of weight is normalized to 1)
     peps.vertices[row, col], peps.vertices[row, cp1] = A, B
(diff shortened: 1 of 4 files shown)


src/states/infiniteweightpeps.jl
```diff
@@ -180,33 +228,15 @@ absorb_weight(t, 2, 3, 2, weights; invwt=true)
...
-        wt = weights[1, row, _prev(col, Nc)]
-    end
-    wt2 = sdiag_pow(wt, pow)
-    indices_t = collect(-1:-1:-5)
-    indices_t[ax] = 1
-    indices_wt = (ax in (2, 3) ? [1, -ax] : [-ax, 1])
-    t2 = permute(ncon((t, wt2), (indices_t, indices_wt)), ((1,), Tuple(2:5)))
-    return t2
+)
+    return _absorb_weights(t, weights, row, col, (ax - 1,), (sqrtwt,), (invwt,))
```
lkdvos, line 239: Why do we need `ax - 1` ? I might just be missing something.

src/states/infiniteweightpeps.jl
```diff
@@ -138,9 +138,57 @@ function InfiniteWeightPEPS(
     )
 end
 
+function Base.size(peps::InfiniteWeightPEPS)
+    return size(peps.vertices)
+end
+
+function _absorb_weights(
```
lkdvos, line 145: I think this function feels slightly awkward to me because it has too many responsibilities:
- working with unitcells and directions
- deciding on `sqrt` and `inv`
- contracting the weights.

As a suggestion, do you think it could be more intuitive if these could be handled separately?

Also, is there ever a situation where you need to have both `sqrt` and non-`sqrt`, and both `inv` and non-`inv` together? It might be slightly simpler to have these as a single `Bool`, but again I might be missing something so feel free to correct me if I'm wrong.

lkdvos (commented):
Thanks for the changes, looks more or less good to me. I don't really understand how `leftorth` and `rightorth` can behave that differently, unless we really are missing a twist somewhere, which is definitely possible.

I left some comments with suggested alterations, but feel free to give your opinion about them too, this is definitely not a requirement or anything.

Yue-Zhengyuan:
I found the issue with `rightorth` when performing FU with an initialization obtained earlier with only `leftorth`. If I continue the evolution with `leftorth` only, the energy change log is
```
step      dt   doping    e_site          Δe       |Δλ|  speed/meas(s)
1      1e-02   0.0252  -7.39936         NaN        NaN  286.213/663.429
10     1e-02   0.0252  -7.39936  -9.835e-07  4.210e-06  174.856/509.017
20     1e-02   0.0252  -7.39937  -1.002e-06  4.242e-06  181.070/538.702
30     1e-02   0.0251  -7.39937  -1.004e-06  4.220e-06  182.025/565.015
```
If I use both `leftorth` and `rightorth`, the energy change log is almost the same:
```
step      dt   doping    e_site          Δe       |Δλ|  speed/meas(s)
1      1e-02   0.0252  -7.39936         NaN        NaN  241.480/734.391
10     1e-02   0.0252  -7.39936  -9.835e-07  4.210e-06  130.743/626.449
20     1e-02   0.0252  -7.39937  -1.001e-06  4.244e-06  137.741/655.694
30     1e-02   0.0251  -7.39937  -1.004e-06  4.219e-06  132.872/651.762
```
(ignore the speed difference; variables are not controlled) 
However, I get lots of warnings during the ALS optimization step when using `rightorth`, which is absent when using only `leftorth`:
```
┌ Warning: GMRES linsolve stopped without converging after 100 iterations:
│ * norm of residual = 1.43e-08
│ * number of operations = 2902
└ @ KrylovKit ~/.julia/packages/KrylovKit/i0JjG/src/linsolve/gmres.jl:123
┌ Warning: GMRES linsolve stopped without converging after 100 iterations:
│ * norm of residual = 1.56e-08
│ * number of operations = 2902
└ @ KrylovKit ~/.julia/packages/KrylovKit/i0JjG/src/linsolve/gmres.jl:123
┌ Warning: GMRES linsolve stopped without converging after 100 iterations:
│ * norm of residual = 7.73e-10
│ * number of operations = 2902
└ @ KrylovKit ~/.julia/packages/KrylovKit/i0JjG/src/linsolve/gmres.jl:123

# and a lot more...
```
It may be because the initialization is obtained using only `leftorth`, which is in a different "gauge" that somehow does not work well with `rightorth`, but frankly I don't fully understand why this happens either. Nevertheless, I guess the save on some twists during bond truncation is a good enough reason for me to use `leftorth` only.

lkdvos:
This really seems to indicate that we are missing some twists somewhere though, because otherwise there is no other difference of gauge: `rightorth` really is defined in terms of `leftorth` (see [here](https://github.com/Jutho/TensorKit.jl/blob/f9cddcf97f8d001888a26f4dce7408d5c6e2228f/src/auxiliary/linalg.jl#L187)), so the only difference can be the arrows.
I'm okay with keeping it as `leftorth` for now, but at some point we should really investigate this a bit more.

src/states/infiniteweightpeps.jl
Yue-Zhengyuan: `PEPSTensor` has 5 axes and `ax` takes values from 2 to 5. In the old implementation, if I want to absorb weight on the "north" axis, I set `ax = 2`, instead of `1` (`NORTH`) because the north axis is the 2nd axis of `t`. But in the new `_absorb_weights` I changed my convention to use the constants `NORTH` etc (1 - 4) to specify the axes into which weights are absorbed. The `ax - 1` here is simply for maintaining backward compatibility...

src/states/infiniteweightpeps.jl
Yue-Zhengyuan: The 3 steps always appear together. I find it convenient to just wrap them into one function (that can handle even the most general case), so I don't need to write the same thing over and over again ... 

`sqrt` and non-`sqrt` appear together during simple update before applying the gate (`sqrt` on the internal bond connecting `A`, `B`; while non-`sqrt` on the open ends):
https://github.com/QuantumKitHub/PEPSKit.jl/blob/503099ab293630f30f444ab85d202787e14182f0/src/algorithms/time_evolution/simpleupdate.jl#L40-L44
But indeed `inv` and non-`inv` never appear together in my use cases. I can change this to a single `inv::Bool` variable.

src/states/infiniteweightpeps.jl
lkdvos: Aha I see. Could we maybe add a todo do change this in the next breaking release? It seems somewhat confusing to have both conventions at the same time. I would also be okay to already change it, as the next version will necessarily be breaking (the keyword args PR is very breaking)

src/states/infiniteweightpeps.jl
lkdvos: Ok, makes sense. Thanks for clarifying!

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -37,19 +37,18 @@ function _su_bondx!(
     cp1 = _next(col, Nc)
     # absorb environment weights
     A, B = peps.vertices[row, col], peps.vertices[row, cp1]
-    A = _absorb_weight(A, row, col, "tbl", peps.weights)
-    B = _absorb_weight(B, row, cp1, "trb", peps.weights)
+    sqrtsA = ntuple(dir -> (dir == EAST), 4)
+    sqrtsB = ntuple(dir -> (dir == WEST), 4)
+    A = _absorb_weights(A, peps.weights, row, col, Tuple(1:4), sqrtsA, false)
+    B = _absorb_weights(B, peps.weights, row, cp1, Tuple(1:4), sqrtsB, false)
```
lkdvos, line 43: Just out of curiosity, looking at the difference here it looks like you went from `"tbl"` to `Tuple(1:4)` (different length). This was on purpose right?

lkdvos (dismissed):
Ok to go for me if positive reply on my last comment.

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: In the old version the weight on the internal bond between A, B is absorbed separately. The `"tbl"` only deals with the open axes. In the new version I combined these two steps together.

lkdvos (approved):

PR #150, opened 2025-03-07
Implement TensorKit's `tsvd!` reverse-rule for `:fixed` differentiation mode

pbrehmer:
Here we adapt `tsvd!` and `FixedSVD` such that the `:fixed` differentiation mode can also use TensorKit's SVD adjoint. For context, the `TensorKit.tsvd!` reverse-rule requires the full, untruncated SVD which is currently not passed along the CTMRG algorithm. To change this, `tsvd!` needs to return the untruncated decomposition as well. Hopefully, TensorKit's reverse-rule will eliminate some of the instabilities of KrylovKit's SVD reverse-rule.

### DIFF
--- a/src/algorithms/ctmrg/projectors.jl
+++ b/src/algorithms/ctmrg/projectors.jl
@@ -77,7 +77,7 @@ function compute_projector(enlarged_corners, coordinate, alg::HalfInfiniteProjec
     end
 
     P_left, P_right = contract_projectors(U, S, V, enlarged_corners...)
-    return (P_left, P_right), info
+    return (P_left, P_right), (; U, S, V, info...)
 end
 function compute_projector(enlarged_corners, coordinate, alg::FullInfiniteProjector)
     halfinf_left = half_infinite_environment(enlarged_corners[1], enlarged_corners[2])
@@ -97,5 +97,5 @@ function compute_projector(enlarged_corners, coordinate, alg::FullInfiniteProjec
     end
 
     P_left, P_right = contract_projectors(U, S, V, halfinf_left, halfinf_right)
-    return (P_left, P_right), info
+    return (P_left, P_right), (; U, S, V, info...)
 end
--- a/src/algorithms/ctmrg/projectors.jl
+++ b/src/algorithms/ctmrg/projectors.jl
@@ -56,15 +56,6 @@ Projector algorithm implementing projectors from SVDing the full 4x4 CTMRG envir
     verbosity::Int = 0
 end
 
-# TODO: add `LinearAlgebra.cond` to TensorKit
-# Compute condition number smax / smin for diagonal singular value TensorMap
-function _condition_number(S::AbstractTensorMap)
-    smax = maximum(first ∘ last, blocks(S))
-    smin = maximum(last ∘ last, blocks(S))
-    return smax / smin
-end
-@non_differentiable _condition_number(S::AbstractTensorMap)
-
 """
     compute_projector(enlarged_corners, coordinate, alg::ProjectorAlgorithm)
 
@@ -75,7 +66,7 @@ function compute_projector(enlarged_corners, coordinate, alg::HalfInfiniteProjec
     # SVD half-infinite environment
     halfinf = half_infinite_environment(enlarged_corners...)
     svd_alg = svd_algorithm(alg, coordinate)
-    U, S, V, truncation_error = PEPSKit.tsvd!(halfinf, svd_alg; trunc=alg.trscheme)
+    U, S, V, info = PEPSKit.tsvd!(halfinf, svd_alg; trunc=alg.trscheme)
 
     # Check for degenerate singular values
     Zygote.isderiving() && ignore_derivatives() do
@@ -86,9 +77,7 @@ function compute_projector(enlarged_corners, coordinate, alg::HalfInfiniteProjec
     end
 
     P_left, P_right = contract_projectors(U, S, V, enlarged_corners...)
-    truncation_error /= norm(S)
-    condition_number = @ignore_derivatives(_condition_number(S))
-    return (P_left, P_right), (; truncation_error, condition_number, U, S, V)
+    return (P_left, P_right), info
 end
 function compute_projector(enlarged_corners, coordinate, alg::FullInfiniteProjector)
     halfinf_left = half_infinite_environment(enlarged_corners[1], enlarged_corners[2])
@@ -97,7 +86,7 @@ function compute_projector(enlarged_corners, coordinate, alg::FullInfiniteProjec
     # SVD full-infinite environment
     fullinf = full_infinite_environment(halfinf_left, halfinf_right)
     svd_alg = svd_algorithm(alg, coordinate)
-    U, S, V, truncation_error = PEPSKit.tsvd!(fullinf, svd_alg; trunc=alg.trscheme)
+    U, S, V, info = PEPSKit.tsvd!(fullinf, svd_alg; trunc=alg.trscheme)
 
     # Check for degenerate singular values
     Zygote.isderiving() && ignore_derivatives() do
@@ -108,6 +97,5 @@ function compute_projector(enlarged_corners, coordinate, alg::FullInfiniteProjec
     end
 
     P_left, P_right = contract_projectors(U, S, V, halfinf_left, halfinf_right)
-    condition_number = @ignore_derivatives(_condition_number(S))
-    return (P_left, P_right), (; truncation_error, condition_number, U, S, V)
+    return (P_left, P_right), info
 end
--- a/src/algorithms/optimization/fixed_point_differentiation.jl
+++ b/src/algorithms/optimization/fixed_point_differentiation.jl
@@ -142,7 +142,6 @@ function _rrule(
     state,
     alg::SimultaneousCTMRG,
 )
-    @assert !isnothing(alg.projector_alg.svd_alg.rrule_alg)
     env, = leading_boundary(envinit, state, alg)
     env_conv, info = ctmrg_iteration(InfiniteSquareNetwork(state), env, alg)
     env_fixed, signs = gauge_fix(env, env_conv)
@@ -150,7 +149,7 @@ function _rrule(
     # Fix SVD
     U_fixed, V_fixed = fix_relative_phases(info.U, info.V, signs)
     svd_alg_fixed = SVDAdjoint(;
-        fwd_alg=FixedSVD(U_fixed, info.S, V_fixed),
+        fwd_alg=FixedSVD(U_fixed, info.S, V_fixed, info.U_full, info.S_full, info.V_full),
         rrule_alg=alg.projector_alg.svd_alg.rrule_alg,
     )
     alg_fixed = @set alg.projector_alg.svd_alg = svd_alg_fixed
--- a/src/utility/svd.jl
+++ b/src/utility/svd.jl
@@ -1,9 +1,11 @@
 using TensorKit:
     SectorDict,
+    RealOrComplexFloat,
     _tsvd!,
     _empty_svdtensors,
     _compute_svddata!,
     _create_svdtensors,
+    _compute_truncdim,
     NoTruncation,
     TruncationSpace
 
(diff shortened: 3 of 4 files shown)


src/utility/svd.jl
```diff
@@ -34,20 +36,95 @@ SVD from `KrylovKit.svdsolve` is used.
 """
 PEPSKit.tsvd(t, alg; kwargs...) = PEPSKit.tsvd!(copy(t), alg; kwargs...)
 function PEPSKit.tsvd!(t, alg::SVDAdjoint; trunc::TruncationScheme=notrunc(), p::Real=2)
-    return TensorKit.tsvd!(t; alg=alg.fwd_alg, trunc, p)
+    return TensorKit.tsvd!(t; alg, trunc, p)
```
leburgel, line 39:
```suggestion
    return TensorKit.tsvd!(t; alg.fwd_alg, trunc, p)
```
Probably this still needs to unpack the `fwd_alg` before deferring to `TensorKit.tsvd!`?

src/utility/svd.jl
```diff
@@ -34,20 +36,95 @@ SVD from `KrylovKit.svdsolve` is used.
...
+    smax = maximum(first ∘ last, blocks(S))
+    smin = maximum(last ∘ last, blocks(S))
+    return smax / smin
+end
+
+# Copy code from TensorKit but additionally return full U, S and V to make compatible with :fixed mode
+function TensorKit._tsvd!(
+    t::TensorMap{<:RealOrComplexFloat}, alg::SVDAdjoint, trunc::TruncationScheme, p::Real=2
+)
+    U, Σ, V⁺, truncerr = tsvd(t; trunc=NoTruncation(), p, alg)
```
leburgel, line 54:
```suggestion
    U, Σ, V⁺, truncerr = PEPSKit.tsvd(t; trunc=NoTruncation(), p, alg)
```
I think this as causing the stack overflow in the tests?

src/utility/svd.jl
```diff
@@ -34,20 +36,95 @@ SVD from `KrylovKit.svdsolve` is used.
...
+
+# adjust TensorKit.tsvd! rrule to info NamedTuple return value
+function ChainRulesCore.rrule(
+    ::typeof(TensorKit.tsvd!),
+    t::AbstractTensorMap;
+    trunc::TruncationScheme=TensorKit.NoTruncation(),
+    p::Real=2,
+    alg::SVDAdjoint=SVDAdjoint(),
+)
+    Ũ, Σ̃, Ṽ⁺, info = tsvd(t; trunc, p, alg)
```
leburgel, line 88:
```suggestion
    Ũ, Σ̃, Ṽ⁺, info = PEPSKit.tsvd(t; trunc, p, alg)
```
I think?

Confusio:
> This is still in a rough shape but I think once I figure out how to gauge-fix the full U, S and V, I hope this works! (The gauge transformation only really acts on the truncated singular vectors, so perhaps an identity on the rest is fine?)

I have a naive question regarding gauge fixing in the `:fixed` mode. If we have already fixed U, S, and V, why is additional gauge fixing necessary? My understanding is that gauge fixing merely sets the relative phases of the singular vectors. Once U, S, and V are determined, the corresponding projectors should be fixed, and the environment tensors should converge element-wise. Did I misunderstand what extra role gauge fixing plays in this process?

pbrehmer:
> If we have already fixed U, S, and V, why is additional gauge fixing necessary? My understanding is that gauge fixing merely sets the relative phases of the singular vectors. Once U, S, and V are determined, the corresponding projectors should be fixed, and the environment tensors should converge element-wise.

It is true that the gauge fixing sets the relative (and global) phases of all environment tensors. The point is that two consecutive CTMRG iterations have to be comparable in the sense that the element-wise differences of all environment tensors should be zero at the fixed point. That's also how the gauge fixing works here: based on a given environment, we perform a CTMRG iteration and fix the Us, Ss and Vs of that iteration relative to the given environment. In order to achieve element-wise convergence, we also need to supply the correct, gauge-fixed environment that fits the gauge-fixed SVD.

In case you want to have a look at the code implementing this, see the [`_rrule`](https://github.com/QuantumKitHub/PEPSKit.jl/blob/9039367690202e96f20b9cb5f996a99c89a9dc7d/src/algorithms/optimization/fixed_point_differentiation.jl#L218) for the `:fixed` mode.

For further details on the implemented gauge fixing algorithm itself, you can take a look at Appendix A [here](https://journals.aps.org/prresearch/abstract/10.1103/PhysRevResearch.7.013237).

pbrehmer:
I think this should work now but there are some things where I could use some input @lkdvos @leburgel:

- In order to hook into the `tsvd` rrule with `FixedSVD` the whole SVD wrapper and its rrules got a bit messier. Since that will be refactored at some point anyway it might be fine for now but I'm still happy about suggestions to clean this up a little.
- The TensorKit `svd_pullback!` function (in the ChainRulesCore extension) has a tolerance for gauge dependency warnings and I'm not yet sure what a sensible way would be to control this from the PEPSKit side.

leburgel:
I'll try to have a proper look later today. As for the gauge dependency warnings tolerance, maybe it's good to let it depend on the magnitude of the smallest singular value that is retained (e.g. divide this by a factor 100 or something to set the tol)?

src/algorithms/ctmrg/projectors.jl
```diff
@@ -144,9 +157,7 @@ function compute_projector(enlarged_corners, coordinate, alg::HalfInfiniteProjec
     end
 
     P_left, P_right = contract_projectors(U, S, V, enlarged_corners...)
-    truncation_error /= norm(S)
-    condition_number = @ignore_derivatives(_condition_number(S))
-    return (P_left, P_right), (; truncation_error, condition_number, U, S, V)
+    return (P_left, P_right), (; U, S, V, info...)
```
lkdvos, line 160: No longer dividing truncation error by norm, is this on purpose?

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -249,6 +244,38 @@ function _rrule(
     return (env_fixed, info), leading_boundary_fixed_pullback
 end
 
+function _fix_svd_algorithm(alg::SVDAdjoint, signs, info)
+    # embed gauge signs in larger space to fix gauge of full U and V on truncated subspace
+    signs_full = map(zip(signs, space.(info.S_full))) do (σ, S)
```
lkdvos, line 249:
```suggestion
    signs_full = map(zip(signs, info.S_full)) do (σ, S)
```
Not really necessary to broadcast in argument of map: this allocates intermediate object you don't need.

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -249,6 +244,38 @@ function _rrule(
...
+        extended_σ = zeros(scalartype(σ), S)
```
lkdvos, line 250:
```suggestion
        extended_σ = one(S)
```

src/utility/svd.jl
```diff
@@ -152,20 +218,28 @@ function random_start_vector(t::AbstractMatrix)
...
-    SVDdata, dims = _compute_svddata!(f, alg, trunc)
+    SVDdata, dims = _compute_svddata!(f, alg.fwd_alg, trunc)
     U, S, V = _create_svdtensors(f, SVDdata, dims)
-    truncerr = trunc isa NoTruncation ? abs(zero(scalartype(f))) : norm(U * S * V - f, p)
+    truncation_error =
+        trunc isa NoTruncation ? abs(zero(scalartype(f))) : norm(U * S * V - f, p)
 
-    return U, S, V, truncerr
+    # construct info NamedTuple
+    truncation_error /= norm(S)
```
lkdvos, line 236: I don't think I agree that `tsvd!` should give you a normalized truncation error by default.

lkdvos (commented):
Overall looks quite good to me. Left a couple small remarks

Confusio:
> > If we have already fixed U, S, and V, why is additional gauge fixing necessary? My understanding is that gauge fixing merely sets the relative phases of the singular vectors. Once U, S, and V are determined, the corresponding projectors should be fixed, and the environment tensors should converge element-wise.
> 
> It is true that the gauge fixing sets the relative (and global) phases of all environment tensors. The point is that two consecutive CTMRG iterations have to be comparable in the sense that the element-wise differences of all environment tensors should be zero at the fixed point. That's also how the gauge fixing works here: based on a given environment, we perform a CTMRG iteration and fix the Us, Ss and Vs of that iteration relative to the given environment. In order to achieve element-wise convergence, we also need to supply the correct, gauge-fixed environment that fits the gauge-fixed SVD.
> 
> In case you want to have a look at the code implementing this, see the [`_rrule`](https://github.com/QuantumKitHub/PEPSKit.jl/blob/9039367690202e96f20b9cb5f996a99c89a9dc7d/src/algorithms/optimization/fixed_point_differentiation.jl#L218) for the `:fixed` mode.
> 
> For further details on the implemented gauge fixing algorithm itself, you can take a look at Appendix A [here](https://journals.aps.org/prresearch/abstract/10.1103/PhysRevResearch.7.013237).

Maybe my question here is even naive: as we already got U, S, V fixed, and the projectors fixed. We use these fixed projectors, do the CTMRG iteration several steps (maybe one step is enough?) to obtain the compatible environment. In other words, the current implement is given environment, we try to find the compatible U, S, V. My question is that since we already have the U, S, V, how to find out the compatible environment? One CTMRG step can help us to do this?

Confusio:
> > If we have already fixed U, S, and V, why is additional gauge fixing necessary? My understanding is that gauge fixing merely sets the relative phases of the singular vectors. Once U, S, and V are determined, the corresponding projectors should be fixed, and the environment tensors should converge element-wise.
> 
> It is true that the gauge fixing sets the relative (and global) phases of all environment tensors. The point is that two consecutive CTMRG iterations have to be comparable in the sense that the element-wise differences of all environment tensors should be zero at the fixed point. That's also how the gauge fixing works here: based on a given environment, we perform a CTMRG iteration and fix the Us, Ss and Vs of that iteration relative to the given environment. In order to achieve element-wise convergence, we also need to supply the correct, gauge-fixed environment that fits the gauge-fixed SVD.
> 
> In case you want to have a look at the code implementing this, see the [`_rrule`](https://github.com/QuantumKitHub/PEPSKit.jl/blob/9039367690202e96f20b9cb5f996a99c89a9dc7d/src/algorithms/optimization/fixed_point_differentiation.jl#L218) for the `:fixed` mode.
> 
> For further details on the implemented gauge fixing algorithm itself, you can take a look at Appendix A [here](https://journals.aps.org/prresearch/abstract/10.1103/PhysRevResearch.7.013237).

I understand why we need gauge-fixed U, S, and V when  the environment are given, but my question is even more naive. After one CTMRG step, we have U, S, and V—and consequently the projectors. We keep them (no gauge fixing), and then perform CTMRG iterations. Does this process produce element-wise convergence environments? 

In other words,  in current implementation for gauge fixing, the environment is given first, and we then find the corresponding compatible U, S, and V and then fix it. My question is: since we already have U, S, and V fixed, how exactly do we determine the compatible environment? Is one CTMRG step sufficient for this adjustment, or are multiple iterations required?

src/algorithms/optimization/fixed_point_differentiation.jl
pbrehmer: Unfortunately, this doesn't work since `extended_sigma` needs to have the (complex) `scalartype(sigma)` but `one(S)` will create real entries.

pbrehmer:
I added an adaptive pullback gauge sensitivity tolerance similar to #155 but perhaps by adapting it directly to the minimal singular value the tolerance is too high now - at least it gets rid of the warnings in the (hopefully) unproblematic cases. Let me know whether we should keep it or not and then we can merge!

pbrehmer:
> After one CTMRG step, we have U, S, and V—and consequently the projectors. We keep them (no gauge fixing), and then perform CTMRG iterations. Does this process produce element-wise convergence environments?

This will not work, since $U$, $S$, and $V$ were obtained by applying the CTMRG iteration $f$ to the previous environment $x_{n-1}$ to obtain the current environment $f(A, x_{n-1}) = x_n$ (where $A$ is the PEPS). Even if $x_{n-1}$ and $x_n$ are at a fixed point in terms of singular values of the corners and edges, in general the elements will not have converged. Applying that same $f_{U,S,V}$ with the fixed SVD to $x_n$ again will rotate the gauge degrees of freedom such that $f_{U,S,V}(A, x_n) = \Lambda x_n$ where $\Lambda$ is some gauge transformation. I think I've also tried this in practice and it will not lead to element-wise convergence.

> My question is: since we already have U, S, and V fixed, how exactly do we determine the compatible environment? Is one CTMRG step sufficient for this adjustment, or are multiple iterations required?

The compatible environment is determined by first gauge fixing the environment obtained by an extra CTMRG step (see the `gauge_fix` function) and, with the same gauge signs, fixing the SVD. The extra step is required to obtain the gauge signs.

Conceptually, two steps are required for proper gauge-fixing since two consecutive environments have to live in the same gauge. It is important to remember that the entire fixed-point differentiation is based on the equation $f(A, x_n) = x_n$. Practically, there are also ways where the gauge is fixed based on one step only, e.g. by fixing the sign of SVDs. However, this can break down (I believe) in cases with degeneracy and is not guaranteed to work.

I hope this makes it somewhat clearer! I don't find it very easy to explain the details, so in case you need further context, feel free to check out the reference and the code itself. But of course I'm happy to answer any questions.

leburgel (approved):
Looks good to me.

leburgel:
> Should I first tag the other version by the way?

You mean 0.4.2? Sure, can't hurt.

pbrehmer:
Sure, then let's tag v0.4.2 first and then I'll merge

PR #156, opened 2025-03-12
Bump KrylovKit compat

leburgel:
Bumps the KrylovKit.jl compat to v0.9.5. We haven't actually been compatible with KrylovKit.jl v0.8 since #94, and more recently #126 requires v0.9.5 to work properly.


lkdvos (approved):
Can you immediately bump the PEPSKit version as well?

pbrehmer (approved):

PR #160, opened 2025-03-18
Split off norm-preserving vector retractions

leburgel:
This splits of the core norm-preserving retraction utilities for vectors as `vector_retract` and `vector_transport!`, and updates the PEPS retractions to broadcast these across the unit cell. The motivation behind this is that I would like to reuse the core retraction tools when optimizing over generic vectors that support VectorInterface (e.g. a single `TensorMap` or a plain `Vector{<:Number}`) in settings where the state we're optimizing has additional structure and is therefore not a generic `InfinitePEPS`.

I also noticed that the `utility/diffset.jl` file was still included, but is no longer used anywhere. Since I think we anyway intended to move away from that, I thought it would be appropriate to just remove it.


src/utility/retractions.jl
```diff
@@ -0,0 +1,59 @@
...
+```math
+ξ = \\cos \\left( α \\frac{||η||}{||A||} \\right) η - \\sin \\left( α \\frac{||η||}{||A||} \\right) ||η|| \\frac{A}{||A||},
+```
+such that ``\\langle A', ξ \\rangle = 0`` and ``||A'|| = ||A||``.
+"""
+function vector_retract(A, η, α)
+    n_A = norm(A)
+    n_η = norm(η)
+    cs = cos(α * n_η / n_A)
+    sn = sin(α * n_η / n_A)
```
lkdvos, line 23:
```suggestion
    sn, cs = sincos(α * n_η / n_A)
```

src/utility/retractions.jl
```diff
@@ -0,0 +1,59 @@
...
+```math
+ξ(α) = \\left\\langle \\frac{η}{||η||}, ξ \\right \\rangle \\left( \\cos \\left( α \\frac{||η||}{||A||} \\right) \\frac{η}{||η||} - \\sin( \\left( α \\frac{||η||}{||A||} \\right) \\frac{A}{||A||} \\right) + Δξ
+```
+such that ``||ξ(α)|| = ||ξ||, \\langle A', ξ(α) \\rangle = 0``.
+"""
+function vector_transport!(ξ, A, η, α, A´)
+    n_A = norm(A)
+    n_η = norm(η)
+    cs = cos(α * n_η / n_A)
+    sn = sin(α * n_η / n_A)
```
lkdvos, line 50:
```suggestion
    sn, cs = sincos(α * n_η / n_A)
```

src/utility/retractions.jl
```diff
@@ -0,0 +1,59 @@
...
+
+    A´ = add(A, η, sn * n_A / n_η, cs)
+    ξ = add(A, η, cs, -sn * n_η / n_A)
+
+    return A´, ξ
+end
+
+"""
+    vector_transport!(ξ, A, η, α, A′)
```
lkdvos, line 32: Same comment about the name

src/utility/retractions.jl
```diff
@@ -0,0 +1,59 @@
...
+
+Transports a direction `ξ` at `A` to a valid direction at `A´` corresponding to
+the norm-preserving retraction of `A` along `η` with step size `α`. In particular, starting
+from a direction `η` of the form
+```math
+ξ = \\left\\langle \\frac{η}{||η||}, ξ \\right\\rangle \\frac{η}{||η||} + Δξ
```
lkdvos, line 38: Since this already using unicode, might as well go all the way:
```suggestion
ξ = ⟨ η / ‖η‖, ξ ⟩ η / ‖η‖ + Δξ
```
and similar below? `\langle`, `\rangle` and `\Vert`

lkdvos (commented):
It seems like somehow codecov is saying the retractions are actually never used. Is this true?

src/utility/retractions.jl
leburgel: The only thing is that I much prefer the `\frac` look in a proper math environment, but agreed that it makes the docstring quite unreadable. I'll update throughout.

lkdvos (approved):
Ok for me if tests turn green :)

PR #161, opened 2025-03-19
Documentation update

leburgel:
### To do

- [x] Standardize docstrings using [DocStringExtensions](https://docstringextensions.juliadocs.org/stable/)
- [x] Add a section explaining the use of [PrecompileTools](https://julialang.github.io/PrecompileTools.jl/stable/) to cut down on long precompilation times, see also #129 
- [x] Use [DocumenterCitations](https://juliadocs.org/DocumenterCitations.jl/stable/) to organize literature references
- [x] Add Pluto notebook generation infracstructure for examples that showcase the full PEPSKit feature set on a few different models

### Examples

- [x] A simple Heisenberg example that can serve as a "getting started" example which shows basics of optimization, CTMRG contraction, computing the correlation length and observables
- [x] Example showing how to use fermionic tensors and that optimizing them might require different settings
- [x] Example showcasing $\mathbb{Z}_2$ or $U(1)$ symmetric optimization
- [ ] Example or part of an example which shows how to use `gradient_alg=nothing`
- [x] Boundary MPS example using VUMPS to contract a PEPS
- [x] Example with PEPS-PEPO-PEPS contraction
- [x] Example for contracting a partition function
- [x] A simple update (and in the future NTU/full update) example

https://quantumkithub.github.io/PEPSKit.jl/previews/PR161/


pbrehmer:
Here are some TODOs for the documentation update from our last meeting but also things that came to my mind recently:
- [ ] Standardize docstrings using [DocStringExtensions](https://docstringextensions.juliadocs.org/stable/)
- [ ] Add a section explaining the use of [PrecompileTools](https://julialang.github.io/PrecompileTools.jl/stable/) to cut down on long precompilation times, see also #129 
- [ ] Use [DocumenterCitations](https://juliadocs.org/DocumenterCitations.jl/stable/) to organize literature references
- [ ] Agree on a unicode TN diagram convention and standardize diagrams in docstrings
- [ ] Add examples in Pluto notebook format that (try to) showcase the full PEPSKit feature set on a few different models

In terms of examples that we want to add to the docs, I think we suggested the following (noting that most of these already exist in the examples folder and only need to be converted to a notebook with some additional explanations):
- [ ] A simple Heisenberg example that can serve as a "getting started" example which shows basics of optimization, CTMRG contraction, computing the correlation length and observables
- [ ] Example showing how to use fermionic tensors and that optimizing them might require different settings
- [ ] Example showcasing $\mathbb{Z}_2$ or $U(1)$ symmetric optimization
- [ ] Example or part of an example which shows how to use `gradient_alg=nothing`
- [ ] Boundary MPS example using VUMPS to contract a PEPS
- [ ] Example with PEPS-PEPO-PEPS contraction
- [ ] A simple update (and in the future NTU/full update) example

Feel free to add suggestions and ideas to the above lists!

leburgel:
https://quantumkithub.github.io/PEPSKit.jl/previews/PR161/

lkdvos:
It would be great if we can all pretend I nailed that first attempt

pbrehmer:
I think this PR is starting to approach the review phase: I have converted (almost) all examples from the examples folder to Pluto notebooks and added comments and explanations. The docstrings now also consistently use DocStringExtensions.jl. I've added a couple of empty manual pages where I think we could give a more systematic and detailed rundown of PEPSKit's features. Here's what still needs to be done, I think:

- Read and revise the docstrings and examples, perhaps add a few more explanations.
- Stabilize the `fermi_hubbard.jl` example: This one takes forever and somehow doesn't converge. However, I've also managed to converge it at some point but haven't figured out how to do that consistently.
- Decide if we want to change the unicode diagrams of the contraction docstrings or leave them as is. It would be nice if there was a tool where you could draw unicode TN diagrams in a standardized way.
- Decide whether we want to write up manual pages still in this PR or do that later, or if we want to have them at all.

As a last thing, I have a couple of thoughts on using Pluto notebooks for the examples. I have run into quite a number of problems with them:

- The fact that you can only have a single line of code per cell or else use a `begin...end` block makes the rendered docs look quite awkward, I find. (Also one needs to use semicolons all the time.)
- The rendered notebooks do not display the printed output of the executed code. This would be desirable e.g. when showing a `fixedpoint` optimization process such that the reader can actually see the convergence process or the effect of different verbosities.
- You cannot crossref in the markdown of the notebooks. That means you can't reference functions within the docs and you can't use DocumenterCitations to cite references. Also, within the docs, it seems you can't crossref the example notebooks.
- The Pluto `.jl` files themselves are not very human-readable. I think it would be nice if one could still use the example files themselves just as runnable/legible scripts such that one doesn't need to look at the docs to make sense of what's going on.
- Working with the Pluto notebooks takes a lot of time due to frequent recompilation and automatic execution of cells. Especially in our examples we have code that takes a long time to precompile and some cells (e.g. where we optimize) take a while to execute. I feel that some of Pluto's features like the consistent cell execution order or having a custom package environment per notebook is not really super useful for our use case.

In my opinion, Literate.jl would be better suited for most of these issues. The only problem is that we definitely don't want to execute all the examples when building the docs because that takes forever. @lkdvos What do you think about this? Do you think that perhaps there is another option, where we can cache the Literate.jl output and build that locally (like we do now with Pluto notebooks)?

lkdvos:
I think what you suggest is basically the setup we have in MPSKit, if that is more convenient I'm definitely open to that. I wanted to try out Pluto.jl but I definitely see you are making valid points. It should be somewhat straightforward to switch this over if everyone agrees on this.

leburgel:
Just for the fact that there's only one line of code per cell I would be happy to switch to Literate.jl. I really don't like the look of this, and the semicolons are a real pain as well.

pbrehmer:
Just had a look at the MPSKit Literate setup and I would 100% prefer that over the Pluto setup. That should definitely be easier to work with. Since we all seem to agree, let me transfer the examples to Literate format and copy over the caching system from MPSKit...

lkdvos:
Small warnings:
- be careful with comments in the examples, I think Literate.jl requires them to start with double `##`.
- That setup runs everything twice IIRC, once for the docs pages, once for the notebooks. This might not be something we want here?
- The caching system could use some love, I find myself often just manually deleting entries in that `.toml` file.

Otherwise, definitely okay with changing this, I'm sorry to have put you through this but I'm glad now I know it's not so easy 😌

leburgel:
I'll try to have a decent look later, but this already looks great! My main comment at this point is that the examples section could use an index page, simply because the table of contents looks a bit messy now. Other than that:

- I think we can skip enforcing a unicode diagram standard for now. I don't see the need to have that block this, we can always do that afterwards.
- I think the manual pages that are in place now are already extremely useful. I don't see manual pages as something essential here, we can always add more later, one at a time.
- I'll have a look at the Fermi-Hubbard example. It does take a long time, but when I put it in I did manage to get it to converge reliably if I remember correctly. It also got very close to the target energy in the end. I'll try to see what's going on there as soon as possible.
- I'll add an extra example on actually optimizing the PEPS fixed point of the 3D classical Ising partition function. This wasn't in the examples folder but it's in a test somewhere, would be a nice addition I think.

pbrehmer:
> My main comment at this point is that the examples section could use an index page, simply because the table of contents looks a bit messy now.

Yes, I definitely agree, I'll add that!

> I'll add an extra example on actually optimizing the PEPS fixed point of the 3D classical Ising partition function. This wasn't in the examples folder but it's in a test somewhere, would be a nice addition I think.

I was also thinking about adding that because I really like that application/test, so I'm happy to hear that.

Other than that, I would be done for now, so perhaps we could review and merge soonish :-)

pbrehmer:
Maybe a small addendum: In the `man/models.md` file I wanted to add the docstrings of the models defined in MPSKitModels.jl that we re-export in PEPSKit since I think that would make the page slightly more usable (you would be able to see the function handles and Hamiltonians without clicking on a link). However, after a couple of attempts I still haven't managed to get that working.

@leburgel @lkdvos would you have any idea how to work this out? The problem really boils down to somehow telling Documenter to import docstrings from an external package. But if this turns out to be complicated, it's not that important for now.

lkdvos:
I feel like this is definitely something I set up for MPSKit, but I no longer remember how exactly. The thing should be:
https://juliadocs.org/DocumenterInterLinks.jl/stable/

leburgel:
> I feel like this is definitely something I set up for MPSKit, but I no longer remember how exactly. The thing should be: https://juliadocs.org/DocumenterInterLinks.jl/stable/

We're already using DocumenterInterlinks here, but I think the issue was that MPSKitModels doesn't use Documenter v1 yet so the interlinks didn't work?

pbrehmer:
> We're already using DocumenterInterlinks here, but I think the issue was that MPSKitModels doesn't use Documenter v1 yet so the interlinks didn't work?

Indeed, and while there are hacky ways to linking MPSKitModels anyways (locally generating an inventory file), I don't think it's worth it. The best thing would be to update the docs of MPSKitModels - and I'm trying that right now but it seems like a pain, so I'm just gonna leave it.

lkdvos:
https://github.com/QuantumKitHub/MPSKitModels.jl/pull/52
After I get this merged, should be available

pbrehmer:
Welp, I still can't get Documenter to display external docstrings via `InterLinks`, so I just replaced the markdown links with `@extref`s. It's not perfect but It's certainly okay, so I'm gonna leave it at that!

lkdvos:
I think I fixed it, but I can't seem to make it not check that all docstrings for MPSKitModels are included, so I had to disable the check.

Solution from https://stackoverflow.com/questions/70137119/how-to-include-the-docstring-for-a-function-from-another-package-in-my-julia-doc

Yue-Zhengyuan:
I made a small change to the formula of Hubbard model hopping term in the examples. It should be $c^+_{i\sigma} c^-_{j\sigma} + h.c. = c^+_{i\sigma} c^-_{j\sigma} + c^+_{j\sigma} c^-_{i\sigma} = c^+_{i\sigma} c^-_{j\sigma} - c^-_{i\sigma} c^+_{j\sigma}$. Previously it was `+` between the two terms.

Yue-Zhengyuan:
I only edited files in `examples`. How do I also update `docs/src/examples`? Do I just run `docs/make.jl`?

pbrehmer:
> How do I also update docs/src/examples? Do I just run docs/make.jl?

You have to run `examples/make.jl` but first remove the entries in `examples/Cache.toml` which you want to re-run. You can find more details on this in `examples/README.md`.

Yue-Zhengyuan:
I tried deleting `examples/Cache.toml` and then run `julia examples/make.jl` at the repo root path, but nothing happened: 
```
(base) zhengyuanyue@MacBook-Pro-2021 PEPSKit.jl % julia examples/make.jl
[ Info: AppleAccelerate loaded
  Activating project at `~/GitHub/PEPSKit.jl/examples`
   Resolving package versions...
  No Changes to `~/GitHub/PEPSKit.jl/examples/Project.toml`
  No Changes to `~/GitHub/PEPSKit.jl/examples/Manifest.toml`
  No Changes to `~/GitHub/PEPSKit.jl/examples/Project.toml`
  No Changes to `~/GitHub/PEPSKit.jl/examples/Manifest.toml`
```
Is there anything else I should configure first?

pbrehmer:
Okay, for some weird reason the line in the `examples/make.jl` file

```julia
examples = filter(isdir, readdir(joinpath(@__DIR__, root))) # filter out directories to ignore Cache.toml, etc.
```

returns an empty vector of strings if the make file is executed from the repo root path using `julia examples/make.jl`. However, if you execute it by first running `julia --project` in `/examples/` and then including the file `include("make.jl")` then the same line will return the correct list of example directories and everything runs as it should. The problem seems to be the `isdir` filtering, since the `readdir` does in fact return the same vector of strings in both cases.

I'll try to figure that out but for now you can just activate the `examples` environment manually

```
cd examples/
julia --project
```

and then run `include("make.jl")`. But keep in mind that it will take a while if you rerun all examples.

examples/make.jl
```diff
@@ -0,0 +1,103 @@
...
+            return cp(joinpath(source_dir, f), joinpath(target_dir, f); force=true)
+        end
+        setcached(root, name)
+    end
+end
+
+function build(root="")
+    cd(@__DIR__)
+    examples = filter(isdir, readdir(joinpath(@__DIR__, root))) # filter out directories to ignore Cache.toml, etc.
+    return map(ex -> build_example(root, ex), examples)
```
lkdvos, line 96:
```suggestion
    return cd(@__DIR__) do
        examples = filter(isdir, readdir(joinpath(@__DIR__, root))) # filter out directories to ignore Cache.toml, etc.
        return map(ex -> build_example(root, ex), examples)
    end
```

You'd probably want to change the directory back to its original afterwards, in order to avoid nasty silent bugs in the future. If I'm not mistaken the above does this.

It did actually explain what the problem is: `readdir` returns relative paths, so an alternative to make this work from anywhere would need `readdir(...; join=true)`, and move some of the logic in `build_example` to handle absolute paths instead.

README.md
```diff
@@ -30,35 +30,40 @@ The package can be installed through the Julia general registry, via the package
 pkg> add PEPSKit
 ```
 
+## Key features
+
+- Construction and manipulation of infinite projected entangled-pair states (PEPS)
+- Contraction of infinite PEPS using the corner transfer matrix renormalization group (CTMRG) and variational uniform matrix product states (VUMPS)
```
lkdvos, line 36:
```suggestion
- Contraction of infinite PEPS using the corner transfer matrix renormalization group (CTMRG) and boundary MPS methods
```

pbrehmer:
> Somehow all of my verbosities end up maxed out and the logs are massive compared to what is committed here

Are you using the PEPSKit master version or the latest release version? The latest release version still has buggy verbosities (leading to super verbose output if `verbosity=3` is used in `fixedpoint`) but I recently fixed that so that `verbosity=3` should now only print the optimization iteration information (which is consistent with MPSKit and KrylovKit).

> My motivation for the reordering would be to go in increasing order of complexity and/or experimentality. I'm not too sure about the internal ordering of the simple update ones, I would just put whichever one is the simplest first.

I was also a bit unsure about the ordering and I think the ordering you proposed makes a lot of sense. For the SU examples I'm fine with both having the Heisenberg model first and then Hubbard, or vice versa. Maybe we need to adjust some of the text slightly after reordering. 

> When reordering I wondered why the folders names needed to start with numbers in the first place. This is quite annoying when moving things around since all the other numbers change as well. If we just manually include the examples in a contents block in a specific order, then we should be able to drop the numbers and more easily change the ordering. Or is this something we really don't want to do?

I also found this annoying but I just copied it from MPSKit since I didn't want to break the build mechanism. I'd be happy to drop the number prefixes and just use a `contents` block if that works.

pbrehmer:
I now removed the directory numbering and updated all examples such that they should run through. I'll add the rest of the rendered examples once they have finished running.

lkdvos:
I've slightly reworked the examples so they are properly linked in the sidebar. My first attempt just put them all on the same level, but that seemed a bit much since they then show up as depth 1 subtitles there, so I added some additional sections. Feel free to modify these as you see fit.
The order is now actually determined in the `docs/make.jl` file, rather than in the `@contents` blocks. Unfortunately, I haven't really found a way to also make the subsection order be inherited automatically, so for now we will have to keep these synchronized manually.

docs/src/man/models.md
```diff
@@ -0,0 +1,64 @@
+# Models
+
+PEPSKit implements physical models through the [MPSKitModels.jl](https://quantumkithub.github.io/MPSKitModels.jl/dev/) package as [`LocalOperator`](@ref) structs.
```
leburgel, line 3: I think this statement is a bit misleading, since `PEPSKit.LocalOperator` is (unfortunately) distinct from `MPSKitModels.LocalOperator`. The is quite annoying, especially since both packages export their `LocalOperator` leading to name clashes when both are loaded in the same session. It would be nice to solve this, but for now just explicitly stating that people should use `PEPSKit.LocalOperator` would be good I think.

pbrehmer: Very good point, and I agree that it would be good to resolve the naming clash at some point. For now I changed it to `PEPSKit.LocalOperator`.

leburgel:
This is starting to look really great. I made some edits in the boundary MPS and partition function examples, I'll try to have a closer look at the rest as well soon.

Do we still want an example showcasing `gradient_alg=nothing`, or do we just leave that for later? I would be fine with leaving it out for now, which means it's just a matter of some minor proofreading before we can merge.

examples/hubbard_su/main.jl
```diff
@@ -0,0 +1,121 @@
+using Markdown #hide
+md"""
+# Simple update for the Fermi-Hubbard model at half-filling
+
+Once again, we consider the Hubbard model but this time we obtain the ground-state PEPS by
+imaginary time evoluation. In particular, we'll use the [`SimpleUpdate`](@ref) algorithm.
```
lkdvos, line 6:
```suggestion
imaginary time evolution. In particular, we'll use the [`SimpleUpdate`](@ref) algorithm.
```

examples/hubbard_su/main.jl
```diff
@@ -0,0 +1,121 @@
...
+wpeps = InfiniteWeightPEPS(rand, Float64, physical_space, virtual_space; unitcell=(Nr, Nc));
+
+md"""
+Before starting the simple update routine, we normalize the vertex tensors of `wpeps` by
+dividing with the maximal vertex element (using the infinity norm):
+"""
+
+for ind in CartesianIndices(wpeps.vertices)
+    wpeps.vertices[ind] /= norm(wpeps.vertices[ind], Inf)
+end
```
lkdvos, line 58:
```suggestion
normalize!.(wpeps.vertices, Inf)
```
But is this necessary? since these are initialized as identity, isn't the infinity norm just `1`?

examples/xxz/main.jl
```diff
@@ -0,0 +1,99 @@
...
+
+Here, we want to look at a special case of the Heisenberg model, where the $x$ and $y$
+couplings are equal, called the XXZ model
+
+```math
+H_0 = J \big(\sum_{\langle i, j \rangle} S_i^x S_j^x + S_i^y S_j^y + \Delta S_i^z S_j^z \big) .
+```
+
+For appropriate $\Delta$, the model enters an antiferromagnetic phase (Néel order) which we
+will force by adding staggered magnetic charges to ``H_0``. Furthermore, since the XXZ
```
lkdvos, line 13:
```suggestion
will force by adding staggered magnetic charges to $H_0$. Furthermore, since the XXZ
```

examples/xxz/main.jl
```diff
@@ -0,0 +1,99 @@
...
+of the model, we will use TensorKit's `U1Space`s where we specify dimensions for each
+symmetry sector. From the virtual spaces, we will need to construct a unit cell (a matrix)
+of spaces which will be supplied to the PEPS constructor. The same is true for the physical
+spaces, which can be extracted directly from the Hamiltonian `LocalOperator`:
+"""
+
+V_peps = U1Space(0 => 2, 1 => 1, -1 => 1)
+V_env = U1Space(0 => 6, 1 => 4, -1 => 4, 2 => 2, -2 => 2)
+virtual_spaces = fill(V_peps, size(lattice)...)
+physical_spaces = H.lattice
```
lkdvos, line 65:
```suggestion
physical_spaces = physicalspace(H)
```
This might need to be defined, (imported from MPSKit)

examples/xxz/main.jl
```diff
@@ -0,0 +1,99 @@
...
+and environment dimensions):
+"""
+
+peps, env, E, info = fixedpoint(
+    H, peps₀, env₀; boundary_alg, gradient_alg, optimizer_alg, verbosity=3
+)
+@show E;
+
+md"""
+Note that for the specified parameters $J=\Delta=1$, we simulated the same Hamiltonian as
```
lkdvos, line 94:
```suggestion
Note that for the specified parameters $J = \Delta = 1$, we simulated the same Hamiltonian as
```

src/algorithms/truncation/fullenv_truncation.jl
```diff
@@ -72,9 +92,9 @@ function _fet_message(
 end
 
 """
-    fullenv_truncate(benv::BondEnv{T,S}, b0::AbstractTensorMap{T,S,1,1}, alg::FullEnvTruncation) where {T<:Number,S<:ElementarySpace}
+    fullenv_truncate(benv::BondEnv{T,S}, b0::AbstractTensorMap{T,S,1,1}, alg::FullEnvTruncation) -> U, S, V, info
 
-The full environment truncation algorithm (Physical Review B 98, 085155 (2018)). 
+Perform full environment truncation algorithm on `benv`.
```
lkdvos, line 97: Do we want to keep the citation here?

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -15,7 +15,12 @@ const TRUNCATION_SCHEME_SYMBOLS = IdDict{Symbol,Type{<:TruncationScheme}}(
     :truncspace => TensorKit.TruncationSpace,
     :truncbelow => TensorKit.TruncationCutoff,
 )
-# Should be TruncationScheme but piracy
+
+"""
+    _TruncationScheme(; kwargs...)
+
+Keyword argument parser returning the appropriate `TruncationScheme` algorithm struct.
+"""
```
lkdvos, line 23: While I do think docs are great, this docstring is both for an internal function and also doesn't really give any more information...
On the other hand, the comment that was there does serve a purpose (although could be worded better): really we want this to be a constructor for `TruncationScheme` but that would be type piracy

src/environments/ctmrg_environments.jl
```diff
@@ -20,9 +20,25 @@ Here `P` represents an effective local constituent tensor. This can either be a
...
+    CTMRGEnv(
+        [f=randn, T=ComplexF64], D_north::P, D_east::P, chi_north::S, [chi_east::S], [chi_south::S], [chi_west::S];
+        unitcell::Tuple{Int,Int}=(1, 1),
+    ) where {P<:ProductSpaceLike,S<:ElementarySpaceLike}
+    CTMRGEnv(
+        [f=randn, T=ComplexF64], network::InfiniteSquareNetwork, chis_north::A, [chis_east::A], [chis_south::A], [chis_west::A]
+    ) where {A<:AbstractMatrix{<:ElementarySpaceLike}}}
+    CTMRGEnv(
+        [f=randn, T=ComplexF64,] network::InfiniteSquareNetwork, chi_north::S, [chi_east::S], [chi_south::S], [chi_west::S],
+    ) where {S<:ElementarySpaceLike}
```
lkdvos, line 41: Can we achieve this by `$(SIGNATURES)`? Given that these are all already documented separately, I think `? CTMRGEnv` would pop up both anyways, so we might also not document the constructors here.

src/environments/ctmrg_environments.jl
```diff
@@ -20,9 +20,25 @@ Here `P` represents an effective local constituent tensor. This can either be a
 rank-4 tensor, a pair of PEPS tensors, or a stack of PEPS-PEPO-PEPS tensors depending on the
 network being contracted.
 
-# Fields
-- `corners::Array{C,3}`: Array of corner tensors.
-- `edges::Array{T,3}`: Array of edge tensors.
+## Fields
+
+$(TYPEDFIELDS)
```
lkdvos, line 25: I think we can keep the information about the arrays in the docstrings of the fields, possibly as `"rows x cols x 4 Array of corner tensors, where the last dimension specifies the directions"` or something similar?

src/operators/infinitepepo.jl
```diff
@@ -2,6 +2,17 @@
...
+## Fields
+
+$(TYPEDFIELDS)
+
+## Constructors
+
+    InfinitePEPO(A::AbstractArray{T, 3})
+    InfinitePEPO([f=randn, T=ComplexF64,] Pspaces, Nspaces, Espaces)
+    InfinitePEPO(A::PEPOTensor; unitcell=(1, 1, 1))
+    InfinitePEPO([f=randn, T=ComplexF64,] Pspace::S, Nspace::S, [Espace::S]; unitcell=(1,1,1)) where {S<:ElementarySpace}
```
lkdvos, line 15: Similar comment about constructors here, If they are documented for the individual functions I don't think we should put them here

src/states/infinitepartitionfunction.jl
```diff
@@ -2,6 +2,18 @@
...
+
+$(TYPEDFIELDS)
+
+## Constructors
+
+    InfinitePartitionFunction(A::AbstractMatrix{T})
+    InfinitePartitionFunction(
+        [f=randn, T=ComplexF64,] Pspaces::A, Nspaces::A, [Espaces::A]
+    ) where {A<:AbstractMatrix{<:Union{Int,ElementarySpace}}}
+    InfinitePartitionFunction(A::PartitionFunctionTensor; unitcell=(1, 1))
```
lkdvos, line 16: Similar comment about constructor docstrings

src/states/infinitepeps.jl
```diff
@@ -2,6 +2,17 @@
...
+## Fields
+
+$(TYPEDFIELDS)
+
+## Constructors
+
+    InfinitePEPS(A::AbstractMatrix{T})
+    InfinitePEPS([f=randn, T=ComplexF64,] Pspaces::A, Nspaces::A, [Espaces::A]) where {A<:AbstractMatrix{<:Union{Int,ElementarySpace}}}
+    InfinitePEPS(A::PEPSTensor; unitcell=(1, 1))
+    InfinitePEPS([f=randn, T=ComplexF64,] Pspace, Nspace, [Espace]; unitcell=(1,1))
```
lkdvos, line 15: Similar comment about constructor docstrings

lkdvos (commented):
I went through the entire thing and left some minor comments, overall looks very good, impressive amounts of work!

As a global comment, I am absolutely not consistent with this, but for markdown text and Latex I try to stick to a "one sentence per line" formatting instead of the usual linebreak of 92 characters. This tends to be a lot friendlier on git, since when you change a sentence somewhere in the middle of a paragraph otherwise the entire paragraph has to be reformatted.

I would say that we should try and put a "feature stop" on this PR and do what it takes to get this merged, and then re-evaluate if we need to make additional changes afterwards, mostly to keep everything comprehensible (reading through the entirety already takes a very long time...)

pbrehmer:
> Do we still want an example showcasing `gradient_alg=nothing`, or do we just leave that for later? I would be fine with leaving it out for now, which means it's just a matter of some minor proofreading before we can merge.

I think we should leave it out for now since that gradient mode is anyway not useful in any reasonable example. My idea is to add a manual page on CTMRG gradients at some point (of course in a future PR) where we briefly explain the differences between the different modes. I think that could be useful for people using PEPSKit to just optimize PEPS but also for those that want to explore and develop the method itself (implementing that stuff just from reading papers is a pain as we know).

And thanks for the detailed review! I'll go through it now and apply the suggested changes.

examples/boundary_mps/main.jl
```diff
@@ -0,0 +1,176 @@
...
+mps₀ = initialize_mps(T, [ComplexSpace(20)])
+
+md"""
+
+Note that this will just construct a MPS with random Gaussian entries based on the physical
+spaces of the supplied transfer operator. Of course, one might come up with a better initial
+guess (leading to better convergence) depending on the application. To find the leading
+boundary MPS fixed point, we call [`leading_boundary`](@ref) using the
+[`MPSKit.VUMPS`](@extref) algorithm from MPSKit. Note that, by default, `leading_boundary`
+uses CTMRG where the settings are supplied as keyword arguments, so in the present case we
```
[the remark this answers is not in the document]
pbrehmer: I was wrong about this. I didn't check the MPSKit definitions of `leading_boundary`, so you're right: just supplying an `InfiniteMPO` should use the default `VUMPS` parameters from MPSKit. I'll remove that sentence then.

examples/heisenberg/main.jl
```diff
@@ -0,0 +1,185 @@
...
+Typically, a random environment which we converge on `peps₀` serves as a good starting point.
+To contract a PEPS starting from an environment using CTMRG, we call [`leading_boundary`](@ref):
+"""
+
+env_random = CTMRGEnv(randn, ComplexF64, peps₀, ℂ^χenv);
+env₀, info_ctmrg = leading_boundary(env_random, peps₀; boundary_alg...);
+
+md"""
+Besides the converged environment, `leading_boundary` also returns a `NamedTuple` of
+informational quantities such as the last (maximal) SVD truncation error:
```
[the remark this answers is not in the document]
pbrehmer: What I meant was: it is the truncation error of the last CTMRG iteration and it is maximal with respect to all SVDs in that iteration, i.e., the environments in all spatial directions and unit cell entries.

examples/xxz/main.jl
pbrehmer: I now defined `physicalspace(::LocalOperator)` and used it consistently throughout the code.

src/algorithms/truncation/fullenv_truncation.jl
pbrehmer: I'll add that back in

src/algorithms/truncation/truncationschemes.jl
pbrehmer: Good point, I'll replace it with an informative comment

src/environments/ctmrg_environments.jl
pbrehmer: I was initially unsure whether one should list all constructors once again in the main docstring but in hindsight that's just redundant information since the constructors have their own docstrings already which are anyway being listed. I'll remove the constructor docstring section in these cases.

examples/hubbard_su/main.jl
pbrehmer: Didn't think about this but you're right: this doesn't change anything so I just removed the normalization here

examples/heisenberg/main.jl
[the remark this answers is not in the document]
lkdvos: Ah I see! Maybe we can slightly reword that to make that more clear?

src/environments/ctmrg_environments.jl
pbrehmer: Good idea, I added a small docstring for each of the fields

examples/heisenberg_su/main.jl
```diff
@@ -0,0 +1,146 @@
...
+        S_ops = real.([S_z(symm)]) ## only Sz preserves <Sz>
+    end
+
+    return [
+        collect(
+            expectation_value(
+                peps, LocalOperator(lattice, (CartesianIndex(r, c),) => S), env
+            ) for (r, c) in Iterators.product(1:size(peps, 1), 1:size(peps, 2))
+        ) for S in S_ops
+    ]
```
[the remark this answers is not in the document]
pbrehmer: I rewrote that bit with a `map` now which looks way nicer indeed

examples/xxz/main.jl
pbrehmer: I imported that now from MPSKit and it does work. I hope that's what you mean?

examples/Cache.toml
```diff
@@ -0,0 +1,10 @@
+["./"]
```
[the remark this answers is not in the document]
pbrehmer: I now got rid of the `root` arguments in `examples/make.jl` since all examples are anyway in the same directory. This should be unrelated to the `cd` call but it had to do with the fact that you could supply different root directories which, however, wasn't used.

The way it is right now should be cleaner for our use case but of course it removes the possibility of having different root directories. I'll just have to rerun the examples to generate new hashes now.

pbrehmer:
This is mostly done from my part. I will rerun the examples and update the cache fully once I get access to a more powerful computer again (I'm currently only on my laptop). But if someone else wants to do that, feel free to do so :-)

@leburgel perhaps you could finish the suggestions for the 3D Ising model? Otherwise this PR would be good to go for me! Thanks again for the comprehensive review.

examples/3d_ising_partition_function/main.jl
```diff
@@ -0,0 +1,267 @@
...
+        ## update the environments for reuse
+        PEPSKit.ignore_derivatives() do
+            PEPSKit.update!(env2, env2′)
+            PEPSKit.update!(env3, env3′)
+        end
+        ## compute the network values per site
+        λ3 = network_value(n3, env3)
+        λ2 = network_value(n2, env2)
+        ## use this to compute the actual cost function
+        return -log(real(λ3 / λ2))
```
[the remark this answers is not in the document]
leburgel: I added a comment in the discussion stating that the cost function being minimized corresponds to the free energy per site.

examples/3d_ising_partition_function/main.jl
```diff
@@ -0,0 +1,267 @@
...
+given manifold point to a valid descent direction at a different manifold point according to
+the appropriate metric. In PEPSKit.jl, these two procedures are defined through the
+[`PEPSKit.peps_retract`](@ref) and [`PEPSKit.peps_transport!`](@ref) methods. While it is
+instructive to read the corresponding docstrings in order to understand what these actually
+do, here we can just blindly reuse them where the only difference is that we have to pass
+along an extra environment since our cost function requires two distinct contractions as
+opposed to the setting of Hamiltonian PEPS optimization.
+"""
+
+function pepo_retract(x, η, α)
```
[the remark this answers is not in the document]
leburgel: I expanded the tuple manifold points everywhere like you suggested. For the other symbols, I don't know if more verbose names would help much in understanding what's going on without reading the linked docstrings and the linked OptimKit README. And there the same symbols are used, and they are actually explained. Since I didn't really want to explain more here, I think making it clearer what the manifold point is should be enough here.

examples/Cache.toml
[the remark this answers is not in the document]
leburgel: Is this solved now? I don't see the `["./"]` in the `Cache.toml` when rebuilding examples, so I assume it is.

leburgel:
I addressed the remaining comments for the 3D classical Ising example I think. As far as I'm concerned, we can update the cache again and merge.

examples/Cache.toml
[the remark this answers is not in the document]
pbrehmer: Yes, this should be solved now

pbrehmer:
I'm just rerunning the examples right now, so I'll push the updated rendered examples and cache, and then I'd be fine with merging as well.

pbrehmer:
Alright, good to go for me once the tests run through :-) @lkdvos What do you think? Also, shall we tag a new release after merging this?

lkdvos:
I'll have breakfast and quickly go through it one last time, but I think I gave my comments in the last round so should be good to go. Agreed with the version update as well, but let's keep that as a separate commit so I can also double check what changed.

lkdvos (approved):

PR #165, opened 2025-03-26
Remove unnecessary type restriction on `_safe_pow`

pbrehmer:
Here we loosen the restriction of the number type in `_safe_pow`. This makes sense since `sdiag_pow` should also work on complex diagonal tensor maps. (I need this for a niche application related to differentiating the SVD which has to do with extra phase freedom for complex SVDs.)

### DIFF
--- a/src/utility/util.jl
+++ b/src/utility/util.jl
@@ -45,7 +45,7 @@ function _elementwise_mult(a₁::AbstractTensorMap, a₂::AbstractTensorMap)
     return dst
 end
 
-_safe_pow(a::Real, pow::Real, tol::Real) = (pow < 0 && abs(a) < tol) ? zero(a) : a^pow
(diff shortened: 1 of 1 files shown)


lkdvos:
Somehow the tests all ran out of time again, while this really shouldn't have changed anything. Any clue?

pbrehmer:
Whoops that is really weird, I really have no idea why that should change anything. Perhaps that changed precompilation times? In any case that shouldn't make such a big difference. I'll try to find some time tomorrow...

lkdvos:
Seems like this is unrelated to this change, so I'm okay with just merging this and trying to fix the time-outs in a different PR

pbrehmer:
Alright, that's certainly fine by me. Then I would think the time-outs are related to different package versions.

lkdvos (approved):

leburgel:
Seems like a Zygote thing, CI was running Zugote v0.7.5 here versus v0.7.4 in the latest run where all the tests passed.

Issue #166, opened 2025-03-27
Performance(?) regressions after update to Zygote.jl v0.7.5


leburgel:
It seems the gradient and example tests are timing out for an unknown reason, as first noticed in #165. Just from looking at the test Project.toml, it seems this might be due to tests now running on Zygote.jl v0.7.5 versus v0.7.4 for the most recent successful test run.

[After #160](https://github.com/QuantumKitHub/PEPSKit.jl/actions/runs/13933102532/job/38994802901):

```julia-repl
     Testing PEPSKit
      Status `/tmp/jl_kXLtOm/Project.toml`
  [7d9f7c33] Accessors v0.1.42
  [d360d2e6] ChainRulesCore v1.25.1
  [cdddcdb0] ChainRulesTestUtils v1.13.0
  [34da2185] Compat v4.16.0
  [26cc04aa] FiniteDifferences v0.12.32
  [0b1a1467] KrylovKit v0.9.5
⌅ [e6f89c97] LoggingExtras v1.0.3
  [bb1c41ca] MPSKit v0.12.6
  [ca635005] MPSKitModels v0.4.0
⌅ [67456a42] OhMyThreads v0.7.0
  [77e91f04] OptimKit v0.4.0
  [52969e89] PEPSKit v0.4.2 `~/work/PEPSKit.jl/PEPSKit.jl`
  [1fd47b50] QuadGK v2.11.2
  [1bc83da4] SafeTestsets v0.1.0
  [07d1fe3e] TensorKit v0.14.5
  [6aa20fa7] TensorOperations v5.1.4
  [409d34a3] VectorInterface v0.5.0
  [e88e6eb3] Zygote v0.7.4
  [37e2e46d] LinearAlgebra
  [de0858da] Printf
  [9a3f8284] Random
  [10745b16] Statistics v1.10.0
  [8dfed614] Test
```

Giving things like:

```julia-repl
Test Summary:                | Pass  Total     Time
Transverse Field Ising model |    5      5  8m31.5s
511.808261 seconds (958.20 M allocations: 93.805 GiB, 2.60% gc time, 87.28% compilation time: <1% of which was recompilation)
```

[After #165](https://github.com/QuantumKitHub/PEPSKit.jl/actions/runs/14111290881/job/39530398280):

```julia-repl
     Testing PEPSKit
      Status `/tmp/jl_RW27Ks/Project.toml`
  [7d9f7c33] Accessors v0.1.42
[...]

PR #168, opened 2025-04-07
Define `Base.show` for `SUWeight`

Yue-Zhengyuan:
This PR defines `Base.show` for `SUWeight`, so we can use `display(peps.weights)` to see the diagonal elements in the weight matrices in an `InfiniteWeightPEPS`. It's a simple IO function so I don't have tests for it. A possible future improvement is to define a more compact output for `println` and `print`. 

Example output for `peps` with `U1Irrep` symmetry: 
```
SUWeight{TensorMap{Float64, GradedSpace{U1Irrep, TensorKit.SortedVectorDict{U1Irrep, Int64}}, 1, 1, Vector{Float64}}}
(1, 1, 1):
Irrep[U₁](0) = [1.0, 0.013284183529494106, 0.01025508746347934]
Irrep[U₁](-1) = [0.155297993782553]
(2, 1, 1):
Irrep[U₁](1/2) = [1.0, 0.013284183529494106]
Irrep[U₁](-1/2) = [0.155297993782553]
Irrep[U₁](3/2) = [0.01025508746347934]
(1, 2, 1):
Irrep[U₁](0) = [1.0, 0.013284183529494106]
Irrep[U₁](1) = [0.155297993782553]
Irrep[U₁](-1) = [0.01025508746347934]
(2, 2, 1):
Irrep[U₁](0) = [1.0, 0.013284183529494106, 0.01025508746347934]
Irrep[U₁](1) = [0.155297993782553]
(1, 1, 2):
Irrep[U₁](0) = [1.0, 0.013284183529494106]
Irrep[U₁](1) = [0.155297993782553]
Irrep[U₁](-1) = [0.01025508746347934]
(2, 1, 2):
Irrep[U₁](0) = [1.0, 0.013284183529494106, 0.01025508746347934]
Irrep[U₁](1) = [0.155297993782553]
(1, 2, 2):
Irrep[U₁](0) = [1.0, 0.013284183529494106, 0.01025508746347934]
Irrep[U₁](-1) = [0.155297993782553]
(2, 2, 2):
Irrep[U₁](1/2) = [1.0, 0.013284183529494106]
Irrep[U₁](-1/2) = [0.155297993782553]
Irrep[U₁](3/2) = [0.01025508746347934]
```

### DIFF
--- a/src/states/infiniteweightpeps.jl
+++ b/src/states/infiniteweightpeps.jl
@@ -49,7 +49,7 @@ function compare_weights(wts1::SUWeight, wts2::SUWeight)
     return sum(_singular_value_distance, zip(wts1.data, wts2.data)) / length(wts1)
 end
 
-function Base.show(io::IO, wts::SUWeight)
+function Base.show(io::IO, ::MIME"text/plain", wts::SUWeight)
     println(io, typeof(wts))
     for idx in CartesianIndices(wts.data)
         println(io, Tuple(idx), ":")
--- a/src/states/infiniteweightpeps.jl
+++ b/src/states/infiniteweightpeps.jl
@@ -49,6 +49,17 @@ function compare_weights(wts1::SUWeight, wts2::SUWeight)
     return sum(_singular_value_distance, zip(wts1.data, wts2.data)) / length(wts1)
 end
 
+function Base.show(io::IO, wts::SUWeight)
+    println(io, typeof(wts))
+    for idx in CartesianIndices(wts.data)
+        println(io, Tuple(idx), ":")
+        for (k, b) in blocks(wts.data[idx])
+            println(io, k, " = ", diag(b))
+        end
+    end
+    return nothing
+end
+
 """
     struct InfiniteWeightPEPS{T<:PEPSTensor,E<:PEPSWeight}
 


pbrehmer:
Looks good to me, I'd be fine with merging this! Some tests sadly still time out but that's unrelated to this PR, see #166. We'll have to figure that out separately.

pbrehmer (approved):

src/states/infiniteweightpeps.jl
```diff
@@ -49,6 +49,17 @@ function compare_weights(wts1::SUWeight, wts2::SUWeight)
     return sum(_singular_value_distance, zip(wts1.data, wts2.data)) / length(wts1)
 end
 
+function Base.show(io::IO, wts::SUWeight)
```
lkdvos, line 52:
```suggestion
function Base.show(io::IO, ::MIME"text/plain", wts::SUWeight)
```

I think the usual convention is to have the 2-argument `show` be something that is Julia-parsable, while the 3-argument is human-readable.
This is definitely a convention that is not always respected, and honestly the entire `show` ecosystem in Julia could do with some improvements, but here I think we can just follow that.

To see the difference, note that by default, displaying something calls the 3-arg show, which falls back to the 2-arg one. However, for example in `Vector` contexts you probably don't want to have this multiline printing, so you can compare these two cases.

Yue-Zhengyuan: Currently I'm content with defining the human-readable `display` only.

lkdvos (approved):

Issue #169, opened 2025-04-07
Space mismatch for simple update with trivial unit cell


lkdvos:
The following minimal example leads to a space mismatch:

```julia
using TensorKit
using MPSKitModels: c_plusmin, c_minplus, c_number
using PEPSKit

# hamiltonian
unitcell = (1, 1)
physical_space = Vect[FermionParity](0 => 1, 1 => 1)
spaces = fill(physical_space, unitcell)
t = 1.0
V = 0.1

hopping = c_plusmin() + c_minplus()
interaction = c_number() ⊗ c_number()
local_h = -t * hopping + V * interaction

H = PEPSKit.nearest_neighbour_hamiltonian(spaces, local_h);

Dbond = 4
virtual_space = Vect[FermionParity](0 => Dbond ÷ 2, 1 => Dbond ÷ 2)
peps = InfiniteWeightPEPS(rand, ComplexF64, physical_space, virtual_space; unitcell);

dt = 1e-2
tol = 1e-6
trscheme = truncerr(1e-10) & truncdim(Dbond)
alg = SimpleUpdate(dt, tol, 1000, trscheme)

peps, = simpleupdate(peps, H, alg; bipartite=false)
[ Info: Space of x-weight at [1, 1] = Vect[FermionParity](0=>2, 1=>2)
[ Info: SU iter 1      :  dt = 1e-02,  weight diff = 1.260e+00,  time = 0.003 sec
ERROR: SpaceMismatch("South space of bond weight y(1, 1) does not match.")
Stacktrace:
```

Interestingly, `unitcell = (2, 2)` works fine. Am I doing something wrong or is this a bug?

Yue-Zhengyuan:
I think the minimal unit cell size for SU is 2 x 2 (if bipartite, then it can be reduced to 1 x 2). When updating a nearest neighbor bond using SVD, I cannot guarantee the two tensors connected by the bond are the same. I can look into this more carefully and raise errors to prevent unintended usage.

lkdvos:
Thanks for the explanation, that does make sense. I never really thought about it too carefully, but I guess raising an error to prevent unintended usage is a good idea.

Yue-Zhengyuan:
The space mismatch here is because the x-bond originally with dimension 2+2 is changed to 1+3. Without the unit cell size restriction, SU only changes the west space to 1+3, but the east space is unchanged. One can force a further truncation on the east space as well, but that violates how SU minimizes its cost function.

PR #170, opened 2025-04-08
Add unit cell size check for simple update

Yue-Zhengyuan:
This PR addresses issue #169 by adding a iPEPS unit cell size check for simple update. It ensure that the unit cell size should be no smaller than (2, 2), because when updating a nearest neighbor bond, SU generally produces two different tensors connected by this bond.


lkdvos (approved):

Yue-Zhengyuan:
Since some Zygote related tests are cancelled, I cannot merge the PR. Could you please help?

Issue #164, opened 2025-03-22
Time evolution for next-nearest neighbor Hamiltonian


Yue-Zhengyuan:
I plan to start adding the ability to handle Hamiltonians with next-nearest neighbor (NNN) terms (e.g. the J1-J2 model) in the iPEPS time evolution algorithms. There are a number of choices:

1. Simple update: [PhysRevB.103.235155](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.103.235155) (Fig. 8)
2. Full update: [PhysRevB.97.174408](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.97.174408)

This two methods constructs an environment around the 3 sites involved in the update. There is another class of "loop update" methods, which treats the 2 x 2 sites around the NNN bond as a PBC-MPS (there will be 2 physical axes at each site):

3. [PhysRevB.94.075143](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.94.075143) This method uses QR/LQ throughout the loop to obtain R, L matrices on the bond to be truncated, and construct projectors on the bond to truncate. 
4. [PhysRevB.102.075147](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.102.075147). This method uses FET to iteratively optimize the tensor at each site on the 2 x 2 loop. 

I plan to add the "loop update" methods first. Except the SU method 1, the NNN term is expressed as a 3-site MPO. The loop update can also involve a 4-site MPO in the 2 x 2 loop.  But in our current way to store the Hamiltonian as a `LocalOperator`, each term is a whole gate. Should we add the ability for `LocalOperator` to store a multi-site term as an MPO, or just do a manual conversion specifically for the time evolution methods?

In addition, could you please also inform me if there are existing methods from MPSKit that can be used to optimize a PBC-MPS?

lkdvos:
Hi Yue,

For the `LocalOperator`, I'm definitely open to having a version that stores an MPO instead of the full gates. I am wondering however if we don't want to just have the option to have both: we could simply also add `LocalFactorizedOperator`, (maybe pick a better name though...) and we can then still decide to remove/replcace the previous `LocalOperator` at some point if this turns out to be better. This has the added benefit that we can easily test the two against eachother for correctness as well.

[...]

PR #171, opened 2025-04-09
Add 3-site simple update (aka 3-site cluster update)

Yue-Zhengyuan:
This PR adds the 3-site simple update (aka 3-site cluster update; see discussion in Issue #164), which can handle Hamiltonians containing up to next nearest neighbor terms (e.g. the J1-J2 model). 

- It convert the Hamiltonian from `LocalOperator` format to a bunch of 3-site MPOs acting on the clusters. Relevant functions are added in `src/algorithms/time_evolution/evoltools.jl`. 
- Truncation of bonds in the 3-site cluster after applying the time evolution gate is done using some "projectors" (I didn't bother to customize OBC MPS algorithms from MPSKit to suit my needs). Their construction is explained in detail in the beginning of the file `src/algorithms/time_evolution/simpleupdate3site.jl`. It is tested in the new test file `test/timeevol/cluster_projectors.jl`. 
- Added an J1-J2 model SU example at `examples/heisenberg_evol/j1j2.jl`. We may also consider using the 3-site SU to provide initialization in the J1-J2 AD test. But note that the 3-site SU also requires a minimal unit cell size (2, 2) (in addition, bipartite structure is now not allowed), while the AD test currently uses 1-site unit cell with C4v symmetry. 

This is kind of a large unseparated PR. Suggestions are welcome. I'm still testing fermion models, so the code is not finalized yet.


Yue-Zhengyuan:
I further added a test to compare the SU and 3-site SU energy for the Hubbard model without next-nearest neighbor terms (they should be close). I think the code is now ready for fermions.

examples/heisenberg_evol/heis_tools.jl
```diff
@@ -1,9 +1,13 @@
+@static if Sys.isapple()
+    using AppleAccelerate
+end
```
lkdvos, line 3: I would be careful with putting this in examples. I agree it's a very relevant performance improvement, but it isn't really tied to the example, and might distract from what we actually want to show.

As a sidenote, I have this in my Julia startup file, so I never have to worry about that and always have it loaded :)

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,523 @@
...
+            | ╱         | ╱
+    --L0-*--Q---  ←  ---M--*-L1--
+           ╱           ╱
+```
+"""
+function lq_through(
+    M::AbstractTensorMap{T,S,1,4}, L1::AbstractTensorMap{T,S,1,1}; normalize::Bool=true
+) where {T<:Number,S<:ElementarySpace}
+    @plansor A[-1; -2 -3 -4 -5] := M[-1; -2 -3 -4 1] * L1[1; -5]
+    l, q = rightorth(A; alg=LQpos())
```
lkdvos, line 166:
```suggestion
    l, q = rightorth!(A; alg=LQpos())
```

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,523 @@
...
+        V = space(Pb, 1)
+        f = isomorphism(flip(V), V)
+        f2 = twist(f, 1)
+        # Pa ← f†) → (f ← s ← f†) → (f ← Pb
+        @tensor begin
+            Pa[-1; -2] := Pa[-1; 1] * f'[1; -2]
+            Pb[-1; -2] := f2[-1; 1] * Pb[1; -2]
+            s[-1; -2] := f2[-1; 1] * s[1; 2] * f'[2; -2]
+        end
+        s = DiagonalTensorMap(permute(s, ((2,), (1,))))
```
lkdvos, line 231: I think there is functionality in TensorKit to "flip" the arrow of a tensor: `flip(t, i)` should do this. This might be more efficient and readable and avoid having to reconstruct the diagonal tensor.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,523 @@
...
+    # local truncation error on each bond
+    ϵs = zeros(N - 1)
+    for (i, (R, L, rev)) in enumerate(zip(Rs, Ls, revs))
+        trunc2 = if isa(trunc, FixedSpaceTruncation)
+            truncspace(space(Ms[i + 1], 1))
+        else
+            trunc
+        end
+        Pas[i], wts[i], Pbs[i], ϵs[i] = _proj_from_RL(R, L; trunc=trunc2, rev)
+    end
```
lkdvos, line 255: This kind of construction is actually type-unstable: you instantiate the vectors with abstract types so the compiler no longer has the information about the concrete types of the elements.
Something worth trying is:

```suggestion
    Pas = Vector{AbstractTensorMap}(undef, N - 1)
    wts = Vector{DiagonalTensorMap}(undef, N - 1)
    Pbs = Vector{AbstractTensorMap}(undef, N - 1)
    # local truncation error on each bond
    ϵs = zeros(N - 1)
    Pas_wts_Pbs = map(1:(N-1)) do i
        trunc2 = if isa(trunc, FixedSpaceTruncation)
            truncspace(space(Ms[i + 1], 1))
        else
            trunc
        end
        Pa, wt, Pb, ϵs[i] = _proj_from_RL(Rs[i], Ls[i]; trunc=trunc2, rev)
        return Pa, wt, Pb
    end
    Pas = map(Base.Fix2(getindex, 1), Pas_wts_Pbs)
    wts = map(Base.Fix2(getindex, 2), Pas_wts_Pbs)
    Pbs = map(Base.Fix2(getindex, 3), Pas_wts_Pbs)
```
The alternative would be to try and compute the concrete types of the results

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,523 @@
...
+    return nothing
+end
+
+"""
+    su3site_iter(gatempos::Dict{Symbol}, peps::InfiniteWeightPEPS, alg::SimpleUpdate)
+
+One round of 3-site simple update for Hamiltonian with 2nd neighbor terms. 
+"""
+function su3site_iter(gatempos::Dict{Symbol}, peps::InfiniteWeightPEPS, alg::SimpleUpdate)
```
lkdvos, line 458: Is there a reason you chose the `gatempos` to be a `Dict{Symbol}` here? Are they ever not `:sw` and `:se`? Otherwise a `Tuple` or `NamedTuple` might be just as descriptive, and would avoid an allocation.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,523 @@
...
+    return nothing
+end
+
+"""
+    simpleupdate3site(peps::InfiniteWeightPEPS, ham::LocalOperator, alg::SimpleUpdate; check_interval::Int=500)
+
+Perform simple update for next-nearest neighbor Hamiltonian `ham`, 
+where the evolution information is printed every `check_interval` steps. 
+"""
+function simpleupdate3site(
```
lkdvos, line 484: Is there a way to detect the kind of hamiltonian that is supplied, and automatically map this to either the nearest-neighbour or the `3-site` version?

src/states/infinitepeps.jl
```diff
@@ -188,6 +188,18 @@ Base.rotl90(A::InfinitePEPS) = InfinitePEPS(rotl90(rotl90.(unitcell(A))))
 Base.rotr90(A::InfinitePEPS) = InfinitePEPS(rotr90(rotr90.(unitcell(A))))
 Base.rot180(A::InfinitePEPS) = InfinitePEPS(rot180(rot180.(unitcell(A))))
 
+## Tensor-wise normalization
+
+function LinearAlgebra.normalize!(peps::InfinitePEPS, p::Real=2)
+    for idx in CartesianIndices(peps.A)
+        peps.A[idx] /= LinearAlgebra.norm(peps.A[idx], p)
+    end
```
lkdvos, line 196:
```suggestion
    normalize!.(unitcell(peps), p)
```
TensorKit should support `normalize` directly.

src/states/infinitepeps.jl
```diff
@@ -188,6 +188,18 @@ Base.rotl90(A::InfinitePEPS) = InfinitePEPS(rotl90(rotl90.(unitcell(A))))
...
+    return peps
+end
+function LinearAlgebra.normalize(peps::InfinitePEPS, p::Real=2)
+    return LinearAlgebra.normalize!(deepcopy(peps))
```
lkdvos, line 200:
```suggestion
    return LinearAlgebra.normalize!(deepcopy(peps), p)
```

test/timeevol/cluster_projectors.jl
```diff
@@ -0,0 +1,126 @@
...
+    Random.seed!(0)
+    N, n = 5, 2
+    for (Vphy, Vns, V) in Vspaces
+        Vvirs = fill(Vns, N + 1)
+        Vvirs[n + 1] = V
+        Ms1 = Vector{AbstractTensorMap}(undef, N)
+        for i in 1:N
+            Vw, Ve = Vvirs[i], Vvirs[i + 1]
+            Ms1[i] = rand(Vw ← Vphy' ⊗ Vns ⊗ Vns' ⊗ Ve) / nrm
+        end
```
lkdvos, line 30: This is again not entirely type stable. I don't mind it as much here in the tests, but in general a `map` construction is probably better

test/timeevol/cluster_projectors.jl
```diff
@@ -0,0 +1,126 @@
...
+    Random.seed!(100)
+    # with U(1) spin rotation symmetry
+    Pspace = hubbard_space(Trivial, U1Irrep)
+    Vspace = Vect[FermionParity ⊠ U1Irrep]((0, 0) => 2, (1, 1//2) => 1, (1, -1//2) => 1)
+    Espace = Vect[FermionParity ⊠ U1Irrep]((0, 0) => 8, (1, 1//2) => 4, (1, -1//2) => 4)
+    trscheme_env = truncerr(1e-12) & truncdim(16)
+    wpeps = InfiniteWeightPEPS(rand, Float64, Pspace, Vspace; unitcell=(Nr, Nc))
+    ham = hubbard_model(ComplexF64, Trivial, U1Irrep, InfiniteSquare(Nr, Nc); t=1.0, U=8.0)
+    # convert to real tensors
+    ham = LocalOperator(ham.lattice, Tuple(ind => real(op) for (ind, op) in ham.terms)...)
```
lkdvos, line 94: We could consider adding this to the `LocalOperator` directly and define `real(ham)` like this.

lkdvos (commented):
I left some general comments throughout, keep in mind that I do not have much experience with the algorithm itself so I definitely might be missing some things or getting things wrong.
As always though, this is an impressive amount of work, and a great addition.

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: This can surely be done: just check that for each term in the Hamiltonian, the distance between the sites of this term is 1. But this can be added in another PR. For test purposes I may deliberately to pass a Hamiltonian with only nearest neighbor terms to the 3-site SU.

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: I can change to `NamedTuple`. Indeed we only use `:sw` and `:se` 3-site clusters.

examples/heisenberg_evol/heis_tools.jl
Yue-Zhengyuan: I didn't know about the `startup.jl` trick. Thanks!

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,469 @@
...
+    @plansor A[-1; -2 -3 -4 -5] := R0[-1; 1] * M[1; -2 -3 -4 -5]
+    q, r = leftorth(A, ((1, 2, 3, 4), (5,)); alg=QRpos())
+    normalize && (r /= norm(r, Inf))
+    return q, r
+end
+function qr_through(
+    ::Nothing, M::AbstractTensorMap{T,S,1,4}; normalize::Bool=true
+) where {T<:Number,S<:ElementarySpace}
+    q, r = leftorth(M, ((1, 2, 3, 4), (5,)); alg=QRpos())
+    normalize && (r /= norm(r, Inf))
```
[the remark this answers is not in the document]
Yue-Zhengyuan: Actually which norm is used is not quite important. 2-norm works just as fine. One just need to prevent `r` explode as QR is successively done through the MPS. I think such "canonicalnization" of an MPS is also done in MPSKit; which norm is being used there?

test/timeevol/cluster_projectors.jl
Yue-Zhengyuan: Good suggestion! I can open a separate PR to implement this.

src/algorithms/time_evolution/simpleupdate3site.jl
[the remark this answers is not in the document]
lkdvos: In general the states are normalized or we want to keep that norm, so we don't explicitly normalize. In canonical form though, the norm of the state would be the 2-norm of the center gauge matrix, so that would be the equivalent thing to normalize (which is also what we do in the infinite case)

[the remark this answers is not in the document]
Yue-Zhengyuan: Thanks for the information. In 3-site SU, the cluster being used is not physically a state, so the exact way to normalize it is not important. I choose to use the `Inf` norm in line with how I normalize the bond weights. 

On the other hand, the bond weights are Schmidt entanglement spectrum and should be better normalized by the 2-norm, so I agree 2-norm may look better, but I just don't want to change my old habit...😅

src/algorithms/time_evolution/evoltools.jl
```diff
@@ -122,3 +129,90 @@ function _apply_gate(
...
+) where {T<:Number,S<:ElementarySpace}
+    g1, s, g23 = tsvd(gate, ((1, 4), (2, 5, 3, 6)); trunc=truncerr(svderr))
+    g1, g23 = PEPSKit.absorb_s(g1, s, g23)
+    g2, s, g3 = tsvd(g23, ((1, 2, 3), (4, 5)); trunc=truncerr(svderr))
+    g2, g3 = PEPSKit.absorb_s(g2, s, g3)
+    # for possibly better numerical stability
+    g1.data[:] = round.(g1.data; digits=14)
+    g2.data[:] = round.(g2.data; digits=14)
+    g3.data[:] = round.(g3.data; digits=14)
+    return [g1, g2, g3]
```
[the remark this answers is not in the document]
Yue-Zhengyuan: Is `decompose_localmpo` using the same axis order?

The reason I add the rounding is that I find many `< 1e-16` numbers in the decomposed MPO from some random gate. For physical gates this does not seem to happen. I can remove the roundings.

src/algorithms/time_evolution/simpleupdate3site.jl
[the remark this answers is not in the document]
lkdvos: I'm fine with either, I was just interested :)

src/algorithms/time_evolution/evoltools.jl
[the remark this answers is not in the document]
lkdvos: `decompose_localmpo` yields `MPOTensor` objects, which use `(W x S <- N x E)` because of planarity reasons

test/timeevol/cluster_projectors.jl
Yue-Zhengyuan: See #177.

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: Thanks for informing me! But I find that `flip` acting on `s` will return a ordinary `TensorMap`, so I still need to add the conversion to `DiagonalTensorMap`.

src/states/infinitepeps.jl
Yue-Zhengyuan: After some thinking I decide not to define `LinearAlgebra.normalize`, but use `normalize!.(unitcell(peps), p)` directly. The former gives people the impression that one is normalizing the entire `peps` such that `<peps|peps> = 1`, but actually I'm just performing tensorwise normalization.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,525 @@
...
+    return l, q
+end
+
+"""
+Given a cluster `Ms`, find all `R`, `L` matrices on each internal bond
+"""
+function _get_allRLs(Ms::Vector{T}) where {T<:PEPSTensor}
+    # M1 -- (R1,L1) -- M2 -- (R2,L2) -- M3
+    N = length(Ms)
+    Rs = Vector{AbstractTensorMap}(undef, N - 1)
```
Yue-Zhengyuan, line 186: @lkdvos The `Rs`, `Ls` here is also not type stable. Here a complication is that each `R`, `L` matrix is obtained successively, and depend on the previous/next element, so I cannot directly use `map` or `collect`. You mentioned that I can compute the type; how can I do this here?

pbrehmer: Perhaps `Base.infer_return_type` does what you need here?

Yue-Zhengyuan: I have now taken a more naive approach. See my comment below.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,532 @@
...
+end
+
+"""
+Given a cluster `Ms`, find all `R`, `L` matrices on each internal bond
+"""
+function _get_allRLs(Ms::Vector{T}) where {T<:PEPSTensor}
+    # M1 -- (R1,L1) -- M2 -- (R2,L2) -- M3
+    N = length(Ms)
+    # get the first R and the last L
+    R_first = qr_through(nothing, Ms[1]; normalize=true)[2]
```
Yue-Zhengyuan, line 187: I now make `Rs`, `Ls` have a more concrete type by first calculating one of their elements, and then create these two lists.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,532 @@
...
+    L_last = lq_through(Ms[N], nothing; normalize=true)[1]
+    Rs = Vector{typeof(R_first)}(undef, N - 1)
+    Ls = Vector{typeof(L_last)}(undef, N - 1)
+    Rs[1], Ls[end] = R_first, L_last
```
Yue-Zhengyuan, line 191: I now make `Rs`, `Ls` have a more concrete type by first calculating one of their elements, and then create these two lists.

src/algorithms/time_evolution/simpleupdate3site.jl
lkdvos: @pbrehmer if possible it is definitely better to avoid depending on these constructions. I was thinking more along the lines of since these are matrices, you kind of know that they will have `tensormaptype(spacetype(T), 1, 1, storagetype(T))` as a type.

@Yue-Zhengyuan this is definitely good for me, but the dependence can actually be resolved by something like `accumulate`, but this typically takes some hoops to make it work cleanly, so if the current approach works let's keep that.

src/algorithms/time_evolution/simpleupdate3site.jl
lkdvos: The reason I'm asking is mostly since I was wondering if we really want to export `simpleupdate3site` as an external-facing function. I would prefer it if you could just use `simpleupdate`, and this would then either tell you that it is not supported or dispatch to a lower level function to implement the specific case at hand.
This has the benefit that if we ever extend or change this, we don't necessarily have to perform breaking changes for this.

test/timeevol/cluster_projectors.jl
lkdvos: This still needs to be adapted here :)

src/algorithms/time_evolution/simpleupdate3site.jl
lkdvos: So I didn't really think this true: defining `flip(::DiagonalTensorMap)` is a little bit more involved... In this case, I think you can avoid the entire flip by doing the svd in the other direction, ie `tsvd!(transpose(rl); trunc)` and then transposing the results. Why does this show up though, i.e. why do we need the `rev` option?

test/timeevol/cluster_projectors.jl
Yue-Zhengyuan: Done:
https://github.com/QuantumKitHub/PEPSKit.jl/blob/52458e97cd71b4f12e13d2b59300a3071a8adabf/test/timeevol/cluster_projectors.jl#L90-L92

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: I did consider doing `tsvd` in the other direction, but didn't use this approach since I'm exhausted getting the twists right... I want to keep the current implementation and wait for the upstream improvement in TensorKit. 

I need `rev` because the arrows in the cluster do not always point in the same direction. Under the usual choice of arrows in an `InfiniteWeightPEPS`, each `:sw` cluster looks like this:
```
    M1
    ↓
    y       
    ↓
    M2 ← x ← M3
```
Thus if I don't transpose `rl` for `tsvd`, I need to reverse the arrow direction between `M1` and `M2` afterwards. 

In addition, After I apply the 3-site gate MPO, I fuse the internal virtual legs of the gate MPO and the cluster such that the result is `M1 ← M2 ← M3`. So I need to convert the result back to `M1 → M2 ← M3`. In principle I could construct the fusers that preserve the arrow direction in the cluster, so I don't need to use an additional `rev` argument to specify on which bonds should I reverse the arrows. But I want to be more flexible and enable manual tweaking of the arrows on each bond. 

BTW, I found that `transpose(rl)` behave differently from `permute(rl, ((2,), (1,)))`. Is it because `transpose` will not produce fermion parity signs?
```julia
using TensorKit
V = Vect[FermionParity](0 => 2, 1 => 2)
t = rand(V ← V);
transpose(t) ≈ permute(t, ((2,), (1,))) # false
transpose(t) ≈ twist(permute(t, ((2,), (1,))), 1) # true
```

lkdvos: The difference between the two is that `transpose` is planar, while `permute` is not. In other words, `transpose` will bend the legs either clockwise or anticlockwise (does not alter the result), while `permute` will cross them.

I understand the feeling about the twists, it really is not the nicest thing to work around...

Yue-Zhengyuan:
I recently did a more thorough test on $J_1$-$J_2$ model, and found that for $D = 4$ and $J_2/J_1 = 0.5$, the spin correlation $\langle S_i \cdot S_j \rangle$ on vertical bonds is always a little bit smaller (magnitude) than horizontal bonds (-0.2917 vs -0.3365). I need some time to think about how to improve on this.

lkdvos:
Is this the same if you "rotate" the hamiltonian and peps every couple iterations, to have a more uniform treatment of the directions?

Yue-Zhengyuan:
Rotating the `InfiniteWeightPEPS` is a bit painful because I also need to move the weights around, and transpose and flip them. I may try the diagonal mirroring though.

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: Now `simpleupdate` can automatically decide whether to use the 2-site or the 3-site version depending on whether the input `ham` only has nearest neighbor terms. But I also added a keyword argument `force_3site` to use the 3-site version for any `ham`.

Yue-Zhengyuan:
I think successive rotations every few steps might be better than mirroring in reducing the artificial anisotropy, but this involves many changes about `InfiniteWeightPEPS`. Can I leave this improvement to a future PR that implements `InfiniteWeightPEPS` rotation?

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: OK, I'll keep the current naive solution.

Yue-Zhengyuan:
I now realize that implementing the rotation for `InfiniteWeightPEPS` has the benefit that it can map one diagonal bond to the other, so I only need to write code for the `:sw` cluster. Let me put this PR on hold for a while and work on the rotations (even though it's really annoying to deal with the twists, it may be quite worthwhile in the long term). After that we may finally relax the constraint on "dualness" in simple update.

Yue-Zhengyuan:
I have tried the 3-site SU using rotations to treat the x and y directions equally. Now although I can make the spin correlation more uniform, for D = 4 I'm easily stuck in a higher energy state, with a large magnetization:
```
  "corH" => [-0.319716 -0.319713; -0.319702 -0.319725]
  "mag" => [[0.245297 -0.245365; -0.245305 0.245373]]
  "corV" => [-0.31973 -0.319712; -0.319718 -0.319698]
  "e_site" => -0.489467
```
In comparison, previously I can get 
```
  "corH" => [-0.337955 -0.338183; -0.337956 -0.338183]
  "mag" => [[0.113206 -0.112297; -0.113216 0.112307]]
  "corV" => [-0.291412 -0.291501; -0.291644 -0.291785]
  "e_site" => -0.49124
```
and Juraj Hasik's [benchmark data](https://github.com/jurajHasik/j1j2_ipeps_states/blob/ea4140fbd7da0fc1b75fac2871f75bda125189a8/single-site_pg-C4v-A1_internal-U1/j20.5/state_1s_A1_U1B_j20.5_D4_chi_opt96.dat) for energy and magnetization (with spin U(1) symmetry) is
```
-0.49424842747070724 0.1423604276676978
```
However, if I use my previous state as initialization for my new code, I get
```
  "corH" => [-0.318614 -0.318632; -0.318614 -0.318632]
  "mag" => [[0.103704 -0.103703; -0.103703 0.103703]]
  "corV" => [-0.318632 -0.318632; -0.318614 -0.318614]
  "e_site" => -0.490849
```
The correlation can become more uniform, although the energy is a little bit higher. Again this demonstrates that time evolution algorithms is extremely sensitive to the initialization...

Yue-Zhengyuan:
I just found a workaround... J2 = 0.5 is quite close to a (possible) spin liquid phase, which is challenging for SU when starting from a random state. Therefore, I can start from a smaller J and gradually increase it. Then I can finally reach a state that is good enough. I'll push the new code and the example after #182 is merged.

Yue-Zhengyuan:
@pbrehmer I have added an example to do SU for J1-J2 model (with U(1) spin symmetry), but haven't generate the docs yet. Could you please try if the obtained SU state can be a good starting point for AD optimization?

pbrehmer:
I'll take a look tomorrow and try out the SU J1-J2 state. I you want, I can then add a section to the example which uses the SU state to initialize an AD optimization (if it works well).

Yue-Zhengyuan:
@pbrehmer I'm glad to see that with SU initialization the AD energy is lower than YASTN result, so maybe a good initialization does help. How does the energy compare with the AD optimization with just 1x1 unit cell starting from random initialization?

pbrehmer:
I was also happy to see that! Actually, on a 1x1 unit cell from random initialization, I'm having a lot of trouble to optimize the PEPS. There, I'm encountering degenerate singular values sometimes and CTMRG sometimes struggles to converge during optimization such that the linesearch fails. I also tried Lander's backtracking linesearch algorithm and also there I'm having problems.

So my feeling is that the SU-evolved PEPS is actually quite a nice starting point for the AD optimization. I'll try to optimize that fully (up to higher accuracy) but I suspect that the environment dimension might be a little low for that - for small chi, PEPS optimization runs tend to get stuck at some point.

Yue-Zhengyuan:
Looking at [YASTN data](https://github.com/jurajHasik/j1j2_ipeps_states/blob/ea4140fbd7da0fc1b75fac2871f75bda125189a8/single-site_pg-C4v-A1_internal-U1/j20.5/state_1s_A1_U1B_j20.5_D4_chi_opt96.dat), the first line is 
```
# uuid CTM_BFGS100LS_U1B_D4-chi96-j20.5-run0-c2U1BD4j2045chi96n0
```
So he is already using $\chi = 96$. Then he measure the optmized state with various $\chi$'s, and the energy converges to 1e-6 precision for $\chi \gtrsim 128$. Indeed a large $\chi$ is necessary. Maybe now `IterSVD` can be used to speed up the optimization?

In the calculation he also imposes the geometric $C_{4v}$ symmetry to the 1x1 unit cell in addition to a sub-lattice spin rotation, which further reduces the number of parameters. Did you do this for the 1x1 optimization before?

pbrehmer:
Indeed! I'm not yet sure if `IterSVD` already pays off here with `D=4` but in any case we can push the environment dimension quite a bit. I didn't impose $C_{4v}$ symmetry but I'll try that next. Of course we don't have proper $C_{4v}$ CTMRG yet because the environment should consist only of one corner and edge tensor (per unit cell entry). But I'm thinking that we should implement that anyway at some point (also if we want to run proper benchmarks).

examples/j1j2_su/main.jl
```diff
@@ -0,0 +1,121 @@
...
+using Random
+using TensorKit, PEPSKit
+Random.seed!(2025);
+
+md"""
+## Simple updating a challenging phase
+
+Let's start by initializing an `InfiniteWeightPEPS` for which we set the required
+parameters as well as physical and virtual vector spaces. Since the $J_1$-$J_2$ model has
+*next*-neighbor interactions, the simple update algorithm requires a $2 \times 2$ unit cell:
```
lkdvos, line 30:
```suggestion
*next*-nearest-neighbor interactions, the simple update algorithm requires a $2 \times 2$ unit cell:
```

lkdvos: Also, I'm slightly confused by this explanation: for me this sounds like next-nearest neighbor hamiltonians have a 2x2 unitcell ground state.
In principle the Hamiltonian still has a translation symmetry for a single site, so I don't think it follows automatically that you would need a $2x2$ unit cell for a next-nearest-neighbour hamiltonian.
From what I understand, this is mostly an algorithmic restriction, and is a result of using simple update, so maybe we can slightly reword this to make that more obvious?

examples/j1j2_su/main.jl
```diff
@@ -0,0 +1,121 @@
...
+"""
+
+dt, tol, maxiter = 1e-2, 1e-8, 30000
+check_interval = 4000
+trscheme_peps = truncerr(1e-10) & truncdim(Dbond)
+alg = SimpleUpdate(dt, tol, maxiter, trscheme_peps)
+for J2 in 0.1:0.1:0.5
+    H = real( ## convert Hamiltonian `LocalOperator` to real floats
+        j1_j2_model(ComplexF64, symm, InfiniteSquare(Nr, Nc); J1, J2, sublattice=false),
+    )
```
lkdvos, line 57:
```suggestion
    H = j1_j2_model(Float64, symm, InfiniteSquare(Nr, Nc); J1, J2, sublattice=false)
```
I know this used to fail since we tried to construct `Sy`, was this not fixed in the meantime?

examples/j1j2_su/main.jl
```diff
@@ -0,0 +1,121 @@
...
+
+md"""
+After we reach $J_2 / J_1 = 0.5$, we gradually decrease the evolution time step to obtain 
+a more accurately evolved PEPS:
+"""
+
+dts = [1e-3, 1e-4]
+tols = [1e-9, 1e-9]
+J2 = 0.5
+H = real(j1_j2_model(ComplexF64, symm, InfiniteSquare(Nr, Nc); J1, J2, sublattice=false))
```
lkdvos, line 70: Same comment about `real` here.

src/algorithms/time_evolution/evoltools.jl
```diff
@@ -1,28 +1,28 @@
 """
-    get_gate(dt::Float64, H::LocalOperator)
+    get_expham(dt::Float64, H::LocalOperator)
 
-Compute `exp(-dt * H)` from the nearest neighbor Hamiltonian `H`.
+Compute `exp(-dt * H)` from Hamiltonian `H`.
+Each term in `H` must be a single `TensorMap`.
```
lkdvos, line 5: This is a bit misleading, given that this is not actually the full exponential of the hamiltonian, but rather a weird intermediate struct we use that represents `sum(exp(-h * dt) for h in H)` but is not used that way. (in particular, `exp(sum)` would be `prod(exp)`, not `sum(exp)`). I don't have any issues with reusing `LocalOperator` that way, I would just reword the docstring.

src/algorithms/time_evolution/evoltools.jl
```diff
@@ -1,28 +1,28 @@
...
 
 """
-    is_equivalent(
-        bond1::NTuple{2,CartesianIndex{2}},
-        bond2::NTuple{2,CartesianIndex{2}},
-        (Nrow, Ncol)::NTuple{2,Int},
-    )
+Check if a Hamiltonian contains only nearest neighbor terms
+"""
+function is_nnham(H::LocalOperator)
```
lkdvos, line 14:
```suggestion
    is_nearest_neighbour(H)

Check if an operator contains only nearest neighbor terms
"""
function is_nearest_neighbour(H::LocalOperator)
```

In general I think using longer names might be easier to read here, but I don't feel that strongly about it.
Adding the signature in the docstring is a good idea though

src/algorithms/time_evolution/evoltools.jl
```diff
@@ -1,28 +1,28 @@
...
+    return all([
+        numin(op) == 2 && norm(Tuple(terms[2] - terms[1])) == 1.0 for (terms, op) in H.terms
+    ])
```
lkdvos, line 17:
```suggestion
    return all(H.terms) do (terms, op)
        return numin(op) == 2 && sum(abs, Tuple(terms[2] - terms[1])) == 1
    end
```
The anonymous function doesn't have to allocate this way, and I think it is safer to use `sum(abs, ...)` to avoid having to deal with floating point equalities.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,465 @@
...
+             ╱           ╱
+    --R0----M---  →  ---Q--*-R1--
+          ╱ |         ╱ |
+```
+"""
+function qr_through(
+    R0::AbstractTensorMap{T,S,1,1}, M::AbstractTensorMap{T,S,1,4}; normalize::Bool=true
+) where {T<:Number,S<:ElementarySpace}
+    @tensor A[-1; -2 -3 -4 -5] := R0[-1; 1] * M[1; -2 -3 -4 -5]
+    q, r = leftorth(A, ((1, 2, 3, 4), (5,)); alg=QRpos())
```
lkdvos, line 140:
```suggestion
    @tensor A[-1 -2 -3 -4; -5] := R0[-1; 1] * M[1; -2 -3 -4 -5]
    q, r = leftorth!(A; alg=QRpos())
```
This might not make a huge difference (yet), but in the long run it's easier to reuse intermediate allocations if they are bundled in `@tensor` calls

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,465 @@
...
+    @assert isdual(domain(r, 1)) == isdual(codomain(r, 1))
+    normalize && (r /= norm(r, Inf))
```
lkdvos, line 142:
```suggestion
    normalize && normalize!(r, Inf)
```

src/algorithms/time_evolution/simpleupdate3site.jl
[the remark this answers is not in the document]
lkdvos: reviving this suggestion since it is an in-place normalization instead, so slightly more efficient

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,465 @@
...
+          ╱ |         ╱ |
+```
+"""
+function lq_through(
+    M::AbstractTensorMap{T,S,1,4}, L1::AbstractTensorMap{T,S,1,1}; normalize::Bool=true
+) where {T<:Number,S<:ElementarySpace}
+    @plansor A[-1; -2 -3 -4 -5] := M[-1; -2 -3 -4 1] * L1[1; -5]
+    l, q = rightorth!(A; alg=LQpos())
+    @assert isdual(domain(l, 1)) == isdual(codomain(l, 1))
+    normalize && (l /= norm(l, Inf))
```
lkdvos, line 168:
```suggestion
    normalize && normalize!(l, Inf)
```

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,465 @@
...
+    return l, q
+end
+function lq_through(
+    M::AbstractTensorMap{T,S,1,4}, ::Nothing; normalize::Bool=true
+) where {T<:Number,S<:ElementarySpace}
+    l, q = rightorth(M; alg=LQpos())
+    @assert isdual(domain(l, 1)) == isdual(codomain(l, 1))
+    normalize && (l /= norm(l, Inf))
```
lkdvos, line 176:
```suggestion
    normalize && normalize!l, Inf)
```

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -0,0 +1,465 @@
...
+    @assert length(Ms) == length(gs)
+    @assert all(!isdual(space(g, 1)) for g in gs[2:end])
+    # fusers to merge axes on bonds in the gate-cluster product
+    # M1 == f1† -- f1 == M2 == f2† -- f2 == M3
+    fusers = collect(
+        begin
+            V1, V2 = space(M, 1), space(g, 1)
+            isomorphism(fuse(V1, V2) ← V1 ⊗ V2)
+        end for (M, g) in zip(Ms[2:end], gs[2:end])
+    )
```
lkdvos, line 285:
```suggestion
    fusers = map(Ms[2:end], gs[2:end]) do M, g
        V1, V2 = space(M, 1), space(g, 1)
        return isomorphism(fuse(V1, V2) ← V1 ⊗ V2)
    end
```

lkdvos (commented):
Left some small comments, but in general this looks almost good to go for me!

src/algorithms/time_evolution/simpleupdate3site.jl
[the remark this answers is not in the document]
Yue-Zhengyuan: I just searched in the repo and find a few `/= norm` elsewhere that can be replaced by `normalize!`. We can improve this on a separate PR.

examples/j1j2_su/main.jl
Yue-Zhengyuan: No, I still cannot directly set `T = Float64` because of `S_y` in `MPSKitModels` v0.4.2.

examples/j1j2_su/main.jl
Yue-Zhengyuan: Indeed it's an algorithm limitation. I'll make this point more clear.

src/algorithms/time_evolution/evoltools.jl
Yue-Zhengyuan: Are we following the British spelling in the repo? (`neighbour` instead of `neighbor`) But I also see that `center` instead of `centre` is used...

Yue-Zhengyuan:
@pbrehmer I tried to regenerate the J1-J2 SU example by deleting the entry in `examples/Cache.toml` and run `julia examples/make.jl`, but it still tries to generate all examples. Can you help look at this?

src/algorithms/time_evolution/evoltools.jl
pbrehmer: I think British and American spelling are mixed throughout the repo, so at this point you can use what you prefer. (Perhaps we can unify at some point but I dont think that's a super high-priority issue right now.)

pbrehmer:
Alright this seems to be good to go now!

Regarding the example cache: I think that the caching is somehow broken when running the examples from a forked repository. This might be because `@__DIR__` returns an absolute path which is then used in the checksum. (The absolute paths are of course different for different users.) I need to clean up this caching mechanism at some point and then I'll take care of that. I'll do that in a separate PR once I find the time/need!

lkdvos (approved):

PR #172, opened 2025-04-10
Add InfinitePEPS to InfiniteWeightPEPS conversion

sanderdemeyer:
Add function to convert an InfinitePEPS to an InfiniteWeightPEPS. This would be useful to have to be able to alternate AD-CTMRG with (trivial) SU for optimization. I'm open to suggestions if someone would want a more general conversion method.

### DIFF
--- a/src/states/infiniteweightpeps.jl
+++ b/src/states/infiniteweightpeps.jl
@@ -268,6 +268,20 @@ function InfinitePEPS(peps::InfiniteWeightPEPS)
 end
 
 """
+    InfiniteWeightPEPS(peps::InfinitePEPS)
+
+Create `InfiniteWeightPEPS` from `InfinitePEPS` by initializing the bond weights as identity matrices of element type `Float64`.
+"""
+function InfiniteWeightPEPS(peps::InfinitePEPS)
+    Nr, Nc = size(peps)
+    weights = map(Iterators.product(1:2, 1:Nr, 1:Nc)) do (d, r, c)
+        V = (d == 1 ? domain(peps[r, c])[2] : domain(peps[r, c])[1])
+        DiagonalTensorMap(ones(reduceddim(V)), V)
+    end
+    return InfiniteWeightPEPS(peps.A, SUWeight(weights))
+end
+
+"""
     mirror_antidiag(peps::InfiniteWeightPEPS)
 
 Mirror the unit cell of an iPEPS with weights by its anti-diagonal line.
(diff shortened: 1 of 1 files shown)


src/states/infiniteweightpeps.jl
```diff
@@ -267,6 +267,20 @@ function InfinitePEPS(peps::InfiniteWeightPEPS)
...
+"""
+    InfiniteWeightPEPS(peps::InfinitePEPS)
+
+Create `InfiniteWeightPEPS` from `InfinitePEPS` by initializing the bond weights as identity matrices of element type `Float64`.
+"""
+function InfiniteWeightPEPS(peps::InfinitePEPS)
+    Nr, Nc = size(peps)
+    weights = map(Iterators.product(1:2, 1:Nr, 1:Nc)) do (d, r, c)
+        V = (d == 1 ? domain(peps[r, c])[2] : domain(peps[r, c])[1])
+        DiagonalTensorMap(ones(reduceddim(V)), V)
```
Yue-Zhengyuan, line 279: To play it safe, shall we add `@assert !isdual(V)` here? Currently whenever `InfiniteWeightPEPS` is used, it is required that the north and east in-space is not dual-ed.

sanderdemeyer: Good idea. I will add this.

Yue-Zhengyuan (approved):
OK to go for me if no further suggestions from others.

Yue-Zhengyuan:
Now I come up with an improvement: we can define `InfiniteWeightPEPS(peps::InfinitePEPS, wts::SUWeight)` to provide an initial "guess" of the weights of the InfiniteWeightPEPS (e.g. from the last AD-SU iteration), so the following trivial SU does not need to start from identity weights all over again. The vertex tensors can be found using `absorb_weights` with `sqrtwt=true, inv=true`.

sanderdemeyer:
This would indeed be a valuable improvement, but I'm still not sure how we could provide a good initial guess of the weights from an InfinitePEPS. 

If we think of the weights as a mean-field approximation of the environment, we could calculate this mean-field environment using CTMRG with environment bond dimension 1, and use the square root of the edges as bond tensors, but I'm not sure whether this is necessarily an improvement over identity weights. I also doubt this would be more efficient than starting from identity weights and just performing SU a few steps more.

Anyway, this seems like a good feature to have, regardless of our way of finding an initial guess of the weights.

PR #174, opened 2025-04-11
Fermionic fixes

lkdvos:
Some missing parity matrices

### DIFF
--- a/src/algorithms/contractions/vumps_contractions.jl
+++ b/src/algorithms/contractions/vumps_contractions.jl
@@ -6,7 +6,7 @@ using MPSKit: GenericMPSTensor, MPSBondTensor
 
 function MPSKit.transfer_left(
     GL::GenericMPSTensor{S,N},
-    O::NonTrivialLocalSandwich,
+    O::Union{PEPSSandwich,PEPOSandwich},
     A::GenericMPSTensor{S,N},
     Ā::GenericMPSTensor{S,N},
 ) where {S,N}
@@ -16,7 +16,7 @@ end
 
 function MPSKit.transfer_right(
     GR::GenericMPSTensor{S,N},
-    O::NonTrivialLocalSandwich,
+    O::Union{PEPSSandwich,PEPOSandwich},
     A::GenericMPSTensor{S,N},
     Ā::GenericMPSTensor{S,N},
 ) where {S,N}
@@ -126,7 +126,7 @@ end
 function MPSKit.contract_mpo_expval(
     AC::GenericMPSTensor{S,N},
     GL::GenericMPSTensor{S,N},
-    O::NonTrivialLocalSandwich,
+    O::Union{PEPSSandwich,PEPOSandwich},
     GR::GenericMPSTensor{S,N},
     ACbar::GenericMPSTensor{S,N}=AC,
 ) where {S,N}
@@ -157,7 +157,7 @@ end
 
 function MPSKit.∂AC(
     AC::GenericMPSTensor{S,N},
-    O::NonTrivialLocalSandwich,
+    O::Union{PEPSSandwich,PEPOSandwich},
     GL::GenericMPSTensor{S,N},
     GR::GenericMPSTensor{S,N},
 ) where {S,N}
--- a/src/algorithms/contractions/vumps_contractions.jl
+++ b/src/algorithms/contractions/vumps_contractions.jl
@@ -5,14 +5,20 @@ using MPSKit: GenericMPSTensor, MPSBondTensor
 #
 
 function MPSKit.transfer_left(
-    GL::GenericMPSTensor{S,N}, O, A::GenericMPSTensor{S,N}, Ā::GenericMPSTensor{S,N}
+    GL::GenericMPSTensor{S,N},
+    O::NonTrivialLocalSandwich,
+    A::GenericMPSTensor{S,N},
+    Ā::GenericMPSTensor{S,N},
 ) where {S,N}
     Ā = twistdual(Ā, 2:N)
     return _transfer_left(GL, O, A, Ā)
 end
 
 function MPSKit.transfer_right(
-    GR::GenericMPSTensor{S,N}, O, A::GenericMPSTensor{S,N}, Ā::GenericMPSTensor{S,N}
+    GR::GenericMPSTensor{S,N},
+    O::NonTrivialLocalSandwich,
+    A::GenericMPSTensor{S,N},
+    Ā::GenericMPSTensor{S,N},
 ) where {S,N}
     Ā = twistdual(Ā, 2:N)
     return _transfer_right(GR, O, A, Ā)
@@ -120,7 +126,7 @@ end
 function MPSKit.contract_mpo_expval(
     AC::GenericMPSTensor{S,N},
     GL::GenericMPSTensor{S,N},
-    O::Union{PEPSSandwich,PEPOSandwich},
+    O::NonTrivialLocalSandwich,
     GR::GenericMPSTensor{S,N},
     ACbar::GenericMPSTensor{S,N}=AC,
 ) where {S,N}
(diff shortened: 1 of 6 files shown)


src/algorithms/toolbox.jl
```diff
@@ -142,12 +142,11 @@ environment `env`.
...
-        env.corners[NORTHWEST, _prev(r, end), _prev(c, end)] *
-        env.corners[NORTHEAST, _prev(r, end), c] *
-        env.corners[SOUTHEAST, r, c] *
-        env.corners[SOUTHWEST, r, _prev(c, end)],
-    )
+    C_NW = env.corners[NORTHWEST, _prev(r, end), _prev(c, end)]
+    C_NE = env.corners[NORTHEAST, _prev(r, end), c]
+    C_SE = env.corners[SOUTHEAST, r, c]
+    C_SW = env.corners[SOUTHWEST, r, _prev(c, end)]
+    return @plansor C_NW[1; 2] * C_NE[2; 3] * C_SE[3; 4] * C_SW[4; 1]
```
Yue-Zhengyuan, line 149: Could you please explain why `@plansor` is used instead of `@tensor`? I think in CTMRG the only place where the fermion sign in contraction should be cancelled is in finding the projectors.

lkdvos: I'm just getting confused here, this is my bad. (I wanted to add some tests before posting this PR, because it's so easy to mess up without, but since I saw the issue I thought I'd go ahead and already post it)

I thought `tr` was secretely being supertrace, and found the negative norms, so I assumed this was wrong and thought we needed the regular trace instead, hence the original twist to compensate.
Then I replaced this with `@plansor`, since that wouldn't give any parity factors, so would be the regular trace.
I now checked, and indeed `tr` is already the regular trace, and not the supertrace, so you are right, this should just be `@tensor`.

src/networks/local_sandwich.jl
```diff
@@ -71,3 +71,9 @@ function virtualspace(O::PEPOSandwich, dir)
         virtualspace(bra(O), dir)',
     ])
 end
+
+## LocalSandwich type union
+
+# purely for type-piracy-avoiding dispatch purposes
+
+const NonTrivialLocalSandwich = Union{PEPOSandwich,PEPSSandwich}
```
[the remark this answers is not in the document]
lkdvos: What if we just don't give it a name and hardcode the `Union`? I want to avoid people using this alias 😁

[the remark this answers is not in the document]
leburgel: My only motivation was not typing the union four times, but the character savings are negligible so I'm fine either way :)

src/algorithms/contractions/vumps_contractions.jl
```diff
@@ -4,14 +4,35 @@ using MPSKit: GenericMPSTensor, MPSBondTensor
...
 ## PEPS
 
-function MPSKit.transfer_left(
+function _transfer_left(
     GL::GenericMPSTensor{S,3},
     O::PEPSSandwich,
     A::GenericMPSTensor{S,3},
     Ā::GenericMPSTensor{S,3},
 ) where {S}
+    Ā = twistdual(Ā, [2, 3])
```
leburgel, line 35:
```suggestion
```

src/algorithms/contractions/vumps_contractions.jl
```diff
@@ -20,12 +41,13 @@ function MPSKit.transfer_left(
...
 end
 
-function MPSKit.transfer_right(
+function _transfer_right(
     GR::GenericMPSTensor{S,3},
     O::PEPSSandwich,
     A::GenericMPSTensor{S,3},
     Ā::GenericMPSTensor{S,3},
 ) where {S}
+    Ā = twistdual(Ā, [2, 3])
```
leburgel, line 50:
```suggestion
```

src/algorithms/contractions/vumps_contractions.jl
```diff
@@ -128,14 +157,27 @@ end
...
 
-function MPSKit.∂AC(
+function _∂AC(
     AC::GenericMPSTensor{S,3},
     O::PEPSSandwich,
     GL::GenericMPSTensor{S,3},
     GR::GenericMPSTensor{S,3},
 ) where {S}
+    GL = twistdual(GL, 1)
+    GR = twistdual(GR, numind(GR))
```
leburgel, line 180:
```suggestion
```

src/networks/tensors.jl
```diff
@@ -95,6 +95,29 @@ Base.rot180(t::PEPSTensor) = permute(t, ((1,), (4, 5, 2, 3)))
...
+    )
+    @tensor O[west south; north east] :=
+        top[phys; top_north top_east top_south top_west] *
+        conj(bot[phys; bot_north bot_east bot_south bot_west]) *
+        twist(F_west, 3)[west; top_west bot_west] *
+        twist(F_south, 3)[south; top_south bot_south] *
+        conj(F_west[east; top_east bot_east]) *
+        conj(F_south[north; top_north bot_north])
+    return O
+end
```
leburgel, line 119:
```suggestion
mpotensor(top::PEPSTensor, bot::PEPSTensor=top) = mpotensor((top, bot))
function mpotensor(P::PEPSSandwich)
    F_west = isomorphism(
        storagetype(ket(P)),
        fuse(west_virtualspace(ket(P)), west_virtualspace(bra(P))'),
        west_virtualspace(ket(P)) ⊗ west_virtualspace(bra(P))',
    )
    F_south = isomorphism(
        storagetype(ket(P)),
        fuse(south_virtualspace(ket(P)), south_virtualspace(bra(P))'),
        south_virtualspace(ket(P)) ⊗ south_virtualspace(bra(P))',
    )
    @tensor O[west south; north east] :=
        ket(P)[phys; top_north top_east top_south top_west] *
        conj(bra(P)[phys; bot_north bot_east bot_south bot_west]) *
        twist(F_west, 3)[west; top_west bot_west] *
        twist(F_south, 3)[south; top_south bot_south] *
        conj(F_west[east; top_east bot_east]) *
        conj(F_south[north; top_north bot_north])
    return O
end
```
Or the other way around, but I would like to have this work on the sandwich type directly as well.

src/utility/util.jl
```diff
@@ -102,6 +102,22 @@ function absorb_s(u::AbstractTensorMap, s::DiagonalTensorMap, vh::AbstractTensor
...
+    twistdual!(t::AbstractTensorMap, i)
+
+Twist the i-th leg of a tensor `t` if it represents a dual space.
+"""
+function twistdual!(t::AbstractTensorMap, i::Int)
+    isdual(space(t, i)) || return t
+    return twist!(t, i)
+end
+function twistdual!(t::AbstractTensorMap, is)
+    is′ = filter(i -> isdual(space(t, i)), is)
```
leburgel, line 116: What happens if there's not dual spaces here? Will this be handled properly?

leburgel (commented):
There's some accidental double twists due to the merge I think, and I would prefer the `mpotensor` converter to be a bit more in line with how we handle local sandwiches elsewhere, but looks great otherwise.

src/networks/tensors.jl
leburgel: Also, this will not work for PEPS tensors for which the north and east space are not the dual of the south and west space respectively, which does happen in larger unit cells. But probably the twists here are specific to the south and west spaces not being dual and the north and east being the duals of these.

So I'm fine with this as long as we assert that the virtual spaces actually satisfy the assumptions here.

src/utility/util.jl
lkdvos: Yes, `twist` handles that case correctly

src/networks/local_sandwich.jl
```diff
@@ -55,6 +55,39 @@ function virtualspace(O::PEPSSandwich, dir)
     return virtualspace(ket(O), dir) ⊗ virtualspace(bra(O), dir)'
 end
 
+# not overloading MPOTensor because that defines AbstractTensorMap{<:Any,S,2,2}(::PEPSTensor, ::PEPSTensor)
+# ie type piracy
+mpotensor(top::PEPSTensor) = mpotensor((top, top))
+function mpotensor((top, bot)::PEPSSandwich)
```
leburgel, line 61: Can you add the MPO tensor conversion to the the fermionic PEPS test you added as well to have some coverage for this?

leburgel (approved):
I added the `mpotensor` converter to the tests, good to go now for me.

PR #175, opened 2025-04-11
Fix optimizer verbosity in `fixedpoint` algorithm selector

pbrehmer:
Small fix for wrong relative verbosity in `select_algorithm(fixedpoint, ...)`: we want to have optimizer output for each iteration if the global verbosity is 1.


lkdvos:
Just out of curiosity, and to check that we are consistent:
In MPSKit and KrylovKit, I think we use:
0. no output
1. warnings
2. init and convergence
3. iteration
4. detailed iteration (subalgorithms output)
5. ...

pbrehmer:
In that case, PEPSKit is not really consistent with MPSKit and KrylovKit. Here, `verbosity=1` will print optimization output at every iteration, `verbosity=2` will additionally print CTMRG information. I guess then it would make more sense to make the verbosity consistent with the rest... Thanks for pointing that out!

lkdvos (approved):

PR #176, opened 2025-04-11
Fix `LocalOperator` constructor for indices of type `Tuple{Int,Int}`

pbrehmer:
Previously, tuple-typed indices would lead to an error since the inner constructor uses a `PeriodicArray` (which apparently can only be indexed using `CartesianIndices`). Now, I convert tuple indices to `CartesianIndices` such that internally always the same type is used but users can still use tuple indices (which we also do in the docstring).

### DIFF
--- a/src/algorithms/select_algorithm.jl
+++ b/src/algorithms/select_algorithm.jl
@@ -43,7 +43,7 @@ function select_algorithm(
 
     # adjust optimizer tol and verbosity
     if optimizer_alg isa NamedTuple
-        defaults = (; tol, verbosity=verbosity ≤ 1 ? -1 : 3)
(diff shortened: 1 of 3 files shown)


lkdvos:
This is not something specific to `PeriodicArray` actually: if you use a tuple of `Int` you need to splat them to make the indexing work. What was actually breaking before?

pbrehmer:
The thing that made it break was [this line](https://github.com/QuantumKitHub/PEPSKit.jl/blob/31e28f7b5cf5cc5062b15ce12b382439d97bf488/src/operators/localoperator.jl#L37-L39):

```julia
for i in 1:length(inds)
    @assert space(operator, i) == plattice[inds[i]]
end
```

since the tuples weren't splatted.

lkdvos (approved):

PR #177, opened 2025-04-13
Define `real`, `imag` for `LocalOperator`

Yue-Zhengyuan:
This PR defines `Base.real` and `Base.imag` for a `LocalOperator` to extract its real and imaginary parts, respectively. Relevant test/example code is updated.

### DIFF
--- a/src/operators/localoperator.jl
+++ b/src/operators/localoperator.jl
@@ -96,6 +96,15 @@ function Base.repeat(O::LocalOperator, m::Int, n::Int)
     return LocalOperator(lattice, terms...)
 end
 
+# Real and imaginary part
+# -----------------------
+function Base.real(O::LocalOperator)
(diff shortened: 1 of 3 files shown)


src/operators/localoperator.jl
```diff
@@ -96,6 +96,15 @@ function Base.repeat(O::LocalOperator, m::Int, n::Int)
     return LocalOperator(lattice, terms...)
 end
 
+# Real and imaginary part
+# -----------------------
+function Base.real(O::LocalOperator)
+    return LocalOperator(O.lattice, Tuple(ind => real(op) for (ind, op) in O.terms)...)
```
lkdvos, line 102:
```suggestion
    return LocalOperator(O.lattice, (ind => real(op) for (ind, op) in O.terms)...)
```
I don't think you need to first construct a tuple for this to work, it should be possible to directly splat the generator

Yue-Zhengyuan: Just fixed (and for `imag` as well). Thanks!

PR #181, opened 2025-04-17
Add modified `tsvd!` reverse-rule with Lorentzian broadening

pbrehmer:
In this PR we will add a rrule for `tsvd!` which supports Lorentzian broadening. To that end, I copy over TensorKit's [`svd_pullback!`](https://github.com/Jutho/TensorKit.jl/blob/fa1551472ac74d7f2a61bdb2135cf418c8c53378/ext/TensorKitChainRulesCoreExt/factorizations.jl#L150) and modify the `save_inv` function to support a broadened inverse.

Additionally, I will add a keyword argument to control the gauge-dependency warnings in that reverse-rule (which is currently not possible) since in some optimizations, the gauge-dependency warnings are really out of control even though they are non-problematic.

### DIFF
--- a/src/utility/svd.jl
+++ b/src/utility/svd.jl
@@ -522,7 +522,7 @@ function svd_pullback!(
         Δgauge = max(Δgauge, norm(view(aUΔU, rprange, rprange), Inf))
         Δgauge = max(Δgauge, norm(view(aVΔV, rprange, rprange), Inf))
     end
-    if verbosity == 1 && Δgauge < tol # warn if verbosity is 1
+    if verbosity == 1 && Δgauge > tol # warn if verbosity is 1
         @warn "`svd` cotangents sensitive to gauge choice: (|Δgauge| = $Δgauge)"
     elseif verbosity ≥ 2 # always info for debugging purposes
         @info "`svd` cotangents sensitive to gauge choice: (|Δgauge| = $Δgauge)"
--- a/src/utility/svd.jl
+++ b/src/utility/svd.jl
@@ -12,8 +12,8 @@ using TensorKit:
 const KrylovKitCRCExt = Base.get_extension(KrylovKit, :KrylovKitChainRulesCoreExt)
 
 """
-    struct FullReverse
-    FullReverse(; kwargs...)
+    struct FullSVDReverseRule
+    FullSVDReverseRule(; kwargs...)
 
 SVD reverse-rule algorithm which uses a modified version of TensorKit's `tsvd!` reverse-rule
 allowing for Lorentzian broadening and output verbosity control.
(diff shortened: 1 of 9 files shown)


pbrehmer:
I'll add a test for the Lorentzian broadening tomorrow and then it should be relatively ready to be merged.

But there's one thing where I'm not sure yet: I want a user verbosity kwarg for the gauge warnings in `svd_pullback!`. However, since we dispatch that rrule on `Nothing`, there's is no way for the user to determine anything in `svd_pullback!` except the broadening, since that is a separate field in `SVDAdjoint`.

My idea to circumvent this would be the following: Add a custom SVD rrule alg for the modified `svd_pullback!` (in the same way that GMRES, BiCGStab, etc. have their own algorithm struct) which contains a `verbosity` and a `broadening` field. I would then get rid of the broadening in `SVDAdjoint` since most algorithms don't support this anyway. I feel that it would make more sense for the `rrule_alg` to contain the broadening field since that is a setting that only has effect during the reverse-pass. Also, I would find it very reasonable if the `SVDAdjoint` only contained a `fwd_alg` and a `rrule_alg` since it is a priori not clear if all SVD adjoints should even support some kind of broadening.

@lkdvos, what do you think about this?

src/utility/svd.jl
```diff
@@ -389,3 +411,177 @@ function ChainRulesCore.rrule(
...
+    if abs(x) < tol
+        return zero(x)
+    else
+        return iszero(ε) ? inv(x) : _lorentz_broaden(x, ε)
+    end
+end
+
+# Lorentzian broadening for divergent term in SVD rrule, see
+# https://journals.aps.org/prresearch/abstract/10.1103/PhysRevResearch.7.013237
+function _lorentz_broaden(x, ε=1e-13)
```
lkdvos, line 426:
```suggestion
function _lorentz_broaden(x, ε=eps(real(scalartype(x)))^(3/4))
```
The power is a bit arbitrary, but at least this should be better behaved for non 64bits types

lkdvos (commented):
Overall looks good to me!
For the default value of the broadening, it would probably be better to have that as a relative thing (and this depends on the datatype), so maybe this could be somewhat improved?

Otherwise I can only comment on the name, `FullReverse` is very non-descriptive, I can't think of anything much better but I would maybe add `SVD` somewhere in there?

pbrehmer:
Yes I agree, the `FullReverse` name was not very good, so I renamed it to `FullSVDReverseRule`.

I adjusted the default broadening value but inside the `Defaults` I left it at `svd_rrule_broadening=1e-13` since there is no way (I think) to determine the relevant floating-point accuracy.

For me this would be good to go now!

lkdvos (approved):
Okay for me once tests pass

PR #182, opened 2025-04-20
Implement rotation for `InfiniteWeightPEPS`

Yue-Zhengyuan:
This PR implements rotations `rotl90`, `rotr90` and `rot180` for `InfiniteWeightPEPS` and `SUWeight`. In addition, we now restrict the virtual arrow directions in an `InfiniteWeightPEPS` to be either downward or leftward for easier operations in simple update algorithms. As a result, the arrow directions in an `InfiniteWeightPEPS` will not rotate together with the state.


Yue-Zhengyuan:
The major problem is the following: suppose we start from an InfiniteWeightPEPS with the "standard" arrow directions. After `rotl90`, the axis order of x-weights needs to be reversed. For fermions, this will make the odd sector of x-weights negative (if it is done with `permute`, not `transpose`). Then I need to twist one of the axes of x-weights to make the weight elements positive. This should be cancelled by another twist on the vertex tensors. But if I define `rotl90` in this way, I found that applying it 4 times will not go back to the original peps because of the extra twists on the virtual indices of vertex tensors. This is shown in the new test file `iwpeps_rotation.jl`. Do you have suggestions on this?

sanderdemeyer:
The problem lies indeed in the added twists, but this does not necessarily have to be a problem. For example, an `InfinitePEPS` is invariant under twists on 'either side', so e.g. calling `twist!(peps[row,col], [2 4]`, `twist!(peps[row,col], [3 5]`, or `twist!(peps[row,col], [2 3 4 5]` will not change the physical state, since the twists cancel on either side of the PEPS tensor. This is also what happens here in the tests. Even though `compose_n(rotl90, 4)(obj) ≈ obj` returns `false`, if you convert `compose_n(rotl90, 4)(obj)` to an `InfinitePEPS`, it will return the same expectation values as `InfinitePEPS(obj)`. This can be easily checked with a random two-site operator. This gauge freedom is inherent. So I can try to help you in defining the rotations in such a way that the tests run, but in terms of getting physical results, the current implementation is probably fine. This is a choice we have to make. I could get behind wanting a different implementation such that the tests run. If we do decide to keep the current implementation, we could add checks on the expectation values, rather than the PEPS tensors themselves.

Yue-Zhengyuan:
I'm aware that the physical state is indeed the same. I just wonder if there is a way that can keep the `InfiniteWeightPEPS` itself invariant without a further gauge transformation. After all, `==` for `InfinitePEPS` is defined without gauge freedom, so I want to mimic this for `InfiniteWeightPEPS` as well.

Yue-Zhengyuan:
If this is indeed too cumbersome for `InfiniteWeightPEPS`, then maybe we just modify `absorb_weight` in such a way that `InfinitePEPS(pepswt1) == InfinitePEPS(pepswt2)` if `pepswt1`, `pepswt2` only differ by some twists that cancel each other.

lkdvos:
I'm not entirely sure that asking for elementwise equality is necessary (or even always possible). In some sense, I'm slightly confused that twists appear here at all, and I would guess that you could simply permute all tensors, and the state would still be unchanged (I didn't check this). After this, you could use the `flip` function to restore the arrows, but in an ideal world this wouldn't even be necessary.

In any case, as @sanderdemeyer said, I think checking that expectation values of random operators are left unchanged is more meaningful than trying to make the tensors be exactly equal.

Yue-Zhengyuan:
@lkdvos The twists is because I want to keep the direction (axis order) of the weight matrices. For example, after `rotl90`, the weight axes order become
```
          2
          |
          y
          |
          1
2 -- x -- 1
```
I need to permute the new weights back to `1 -- x -- 2`, resulting in negative elements for fermions. But these weights are rough analogs of the Schmidt weights in 1D MPS, so I want to keep the weight elements positive. It can also make it easier to absorb the square roots of the weights into the vertex tensors. In principle I can lift the positiveness constraint, but I just feel uncomfortable about doing it.

Yue-Zhengyuan:
@sanderdemeyer I'll try implementing the rotation that does not rotate the arrows together with the state this week, and add the constraint that virtual arrows in `InfiniteWeightPEPS` can only be ← or ↓. What do you think of this approach?

sanderdemeyer:
@Yue-Zhengyuan This is perfect for me. It feels like the safest choice to always impose that an `InfinitePEPS` or `InfiniteWeightPEPS` has arrows ← or ↓ on the virtual legs.

test/utility/iwpeps_rotation.jl
```diff
@@ -0,0 +1,66 @@
...
+    @test compose_n(rotl90, 4)(wts) ≈ wts
+    @test compose_n(rot180, 2)(wts) ≈ wts
+    return nothing
+end
+
+function test_rotation(psi::InfiniteWeightPEPS)
+    peps = InfinitePEPS(psi)
+    @test compose_n(rotl90, 2)(psi) ≈ compose_n(rotr90, 2)(psi)
+    @test compose_n(rotr90, 2)(psi) ≈ rot180(psi)
+    # flipping twice results in a twist
```
Yue-Zhengyuan, line 25: I notice a related PR on TensorKit (https://github.com/Jutho/TensorKit.jl/pull/241), and in v0.14.6 it is stated that `flip ∘ flip` is not identity. Here for fermions I find that it is equal to a twist. Is this also true for more general tensors, i.e. for anyons?

test/utility/iwpeps_rotation.jl
```diff
@@ -0,0 +1,66 @@
...
+    psi_l4 = compose_n(rotl90, 4)(psi)
+    psi_r4 = compose_n(rotr90, 4)(psi)
+    psi_2 = compose_n(rot180, 2)(psi)
+    @test psi_l4 ≈ psi_r4 ≈ psi_2
+    @test all(twist(v1, Tuple(2:5)) ≈ v2 for (v1, v2) in zip(psi_2.vertices, psi.vertices))
+    # conversion to InfinitePEPS
+    psi_l = rotl90(psi)
+    peps_l = InfinitePEPS(psi_l)
+    for (i, t) in enumerate(peps_l.A)
+        peps_l.A[i] = flip(t, (3, 5); inv=true)
```
Yue-Zhengyuan, line 40: The test requires TensorKit 0.14.6 or higher. Should we change Project.toml?

Yue-Zhengyuan:
@sanderdemeyer Could you please take a look and see if there are further improvements you want?

BTW, after this is merged, I'll use rotation instead of mirroring to swap x/y directions in SU. I find this leads to a more stable result in some of my use cases. Then `mirror_antidiag` will be used nowhere in PEPSKit, and I plan to remove it then.

test/utility/iwpeps_rotation.jl
sanderdemeyer: Even though I'm still of the opinion that we do not need to do this, we might be able to retrieve the same tensors (i.e. without twists) by choosing appropriately between forward and inverse flips. This might be safer for the anyonic case as well. I do not know enough about the anyonic case to be sure about this, so I'll ask the people who have more experience with this.

sanderdemeyer:
I had hoped that we could recover the same state (without the twists) if we would properly distinguish between `inv=true` and `inv=false` flips. Even though this is theoretically possible, this would again require us to relax the constraint to have positive weight elements. Since I agree with you that this is not the best idea, I am okay with keeping the current implementation.

In conclusion, I think we can merge this.

Yue-Zhengyuan:
> properly distinguish between `inv=true` and `inv=false` flips

I did consider adding a kwarg to the rotation functions to specify whether `inv=true` or `inv=false` flips are used. But in the end I think this is too much, and makes repeated application of rotations more confusing.

src/states/infiniteweightpeps.jl
```diff
@@ -24,7 +24,16 @@ struct SUWeight{E<:PEPSWeight}
...
+        eltype(data[1]) <: Real || error("Weight elements must be real numbers.")
+        for wt in data
+            isa(wt, DiagonalTensorMap) ||
+                error("Each weight matrix should be a DiagonalTensorMap")
+            domain(wt, 1) == codomain(wt, 1) ||
+                error("Domain and codomain of each weight matrix must be the same.")
+            !isdual(codomain(wt, 1)) ||
+                error("Domain and codomain of each weight matrix cannot be a dual space.")
+            all(wt.data .>= 0) || error("Weight elements must be non-negative.")
+        end
```
lkdvos, line 36: Are we sure we want to put these restrictions on the inner constructor? This also means that there is no way to bypass them, so this might be slightly unpractical.
I would prefer to have an outer constructor that has these checks instead.

src/states/infiniteweightpeps.jl
```diff
@@ -24,7 +24,16 @@ struct SUWeight{E<:PEPSWeight}
     data::Array{E,3}
 
     function SUWeight(data::Array{E,3}) where {E<:PEPSWeight}
-        @assert eltype(data[1]) <: Real
+        eltype(data[1]) <: Real || error("Weight elements must be real numbers.")
```
lkdvos, line 27:
```suggestion
        scalartype(data) <: Real || error("Weight elements must be real numbers.")
```

src/states/infiniteweightpeps.jl
```diff
@@ -50,6 +59,21 @@ VectorInterface.scalartype(::Type{T}) where {T<:SUWeight} = scalartype(eltype(T)
...
 Base.setindex!(W::SUWeight, args...) = (Base.setindex!(W.data, args...); W)
 Base.axes(W::SUWeight, args...) = axes(W.data, args...)
+Base.iterate(W::SUWeight, args...) = iterate(W.data, args...)
+
+## (Approximate) equality
+function Base.:(==)(wts1::SUWeight, wts2::SUWeight)
+    for (wt1, wt2) in zip(wts1, wts2)
+        !(wt1 == wt2) && return false
+    end
+    return true
```
lkdvos, line 69:
```suggestion
    return wts1.data == wts2.data
```

src/states/infiniteweightpeps.jl
```diff
@@ -296,17 +324,212 @@ function InfiniteWeightPEPS(peps::InfinitePEPS)
     return InfiniteWeightPEPS(peps.A, SUWeight(weights))
 end
 
+## (Approximate) equality (gauge freedom is not allowed)
+function Base.:(==)(peps1::InfiniteWeightPEPS, peps2::InfiniteWeightPEPS)
+    for (v1, v2) in zip(peps1.vertices, peps2.vertices)
+        !(v1 == v2) && return false
+    end
+    !(peps1.weights == peps2.weights) && return false
+    return true
```
lkdvos, line 333:
```suggestion
    return peps1.vertices == peps2.vertices && peps1.weights == peps2.weights
```

src/states/infiniteweightpeps.jl
```diff
@@ -296,17 +324,212 @@ function InfiniteWeightPEPS(peps::InfinitePEPS)
...
+"""
+function mirror_antidiag(wts::SUWeight)
+    weights2_x = mirror_antidiag(wts[2, :, :])
+    weights2_y = mirror_antidiag(wts[1, :, :])
+    return SUWeight(weights2_x, weights2_y)
+end
+function Base.rotl90(wts::SUWeight)
+    wts_x = rotl90(wts[2, :, :])
+    Nc = size(wts_x, 2)
+    wts_x = wts_x[:, vcat(2:Nc, 1)]
```
lkdvos, line 467:
```suggestion
    wts_x = circshift(rotl90(wts[2, :, :]), (0, 1))
```
I always forget the direction so I'm not sure if it should be +1 or -1 for the shift.

src/states/infiniteweightpeps.jl
```diff
@@ -296,17 +324,212 @@ function InfiniteWeightPEPS(peps::InfinitePEPS)
...
+        wts_y[i] = DiagonalTensorMap(flip(permute(wt, ((2,), (1,))), (1, 2)))
+    end
+    return SUWeight(wts_x, wts_y)
+end
+function Base.rot180(wts::SUWeight)
+    wts_x = rot180(wts[1, :, :])
+    wts_y = rot180(wts[2, :, :])
+    Nr, Nc = size(wts_x)
+    wts_x = wts_x[:, vcat(2:Nc, 1)]
+    wts_y = wts_y[vcat(Nr, 1:(Nr - 1)), :]
```
lkdvos, line 489: Same comment about circshift here

test/utility/iwpeps_rotation.jl
lkdvos: I think the answer here is definitely no for anyons. Really, the `flip` function is probably only well-defined for symmetric braiding rules anyways, and even there it is not a unique choice, it's just a canonical choice of isomorphism between the space and the dual, implemented in such a way that it does not alter subsequent contractions through `@tensor` (which would not work for anyons anyways).

test/utility/iwpeps_rotation.jl
lkdvos: Alternatively you can put this test behind a `@static if pkgversion(TensorKit) >= v"0.14.6"`, especially if it is only for the test.

src/states/infiniteweightpeps.jl
Yue-Zhengyuan: I was intentionally doing so to help avoid unintended changes to InfiniteWeightPEPS instances in other functions. Until we meet an application that can benefit significantly from more flexibility I don't quite want to relax these restrictions.

lkdvos: I don't mean to not have these checks, I mean to have them by default but still have the option to not have them. For example, the current construction would not allow you to have an uninitialized struct that you fill up afterwards, which can be convenient for `similar` or `copy` functions.

The code pattern I'm thinking of is:
```julia
struct SUWeight{E<:PEPSWeight}
    data::Array{E,3}
end
function SUWeight(data::Array{<:PEPSWeight,3})
    ... # checks here
    return SUWeight{eltype(data)}(data)
end
```
Effectively, calls to `SUWeight(data)` will do the normal constructor checks, while `SUWeight{E}` is the "expert level" constructor that doesn't.

src/states/infiniteweightpeps.jl
Yue-Zhengyuan: Just checked: it is -1 in this case.
```julia
julia> a = collect(1:5);

julia> println(circshift(a, -1))
[2, 3, 4, 5, 1]
```

test/utility/iwpeps_rotation.jl
Yue-Zhengyuan: Then should I change the test to use `flip` only? i.e. replace the twists below with two inverse flips.

src/states/infiniteweightpeps.jl
Yue-Zhengyuan: That's a fair reason (being able to fill up the state afterwards). But when I try your code pattern I get pre-compilation warning complaining that the constructor is being overwritten. How can I avoid this conflict?

lkdvos: You might need an inner constructor as well, I'm not entirely sure:
```julia
struct SUWeight{E<:PEPSWeight}
    data::Array{E,3}
    SUWeight{E}(data::Array{E,3}) where {E} = new{E}(data)
end

function SUWeight(data::Array{<:PEPSWeight,3})
    scalartype(data) <: Real || error("Weight elements must be real numbers.")
    for wt in data
        isa(wt, DiagonalTensorMap) ||
            error("Each weight matrix should be a DiagonalTensorMap")
        domain(wt, 1) == codomain(wt, 1) ||
            error("Domain and codomain of each weight matrix must be the same.")
        !isdual(codomain(wt, 1)) ||
            error("Domain and codomain of each weight matrix cannot be a dual space.")
        all(wt.data .>= 0) || error("Weight elements must be non-negative.")
    end
    return SUWeight{eltype(data)}(data)
end
```

Yue-Zhengyuan: This works. Thanks!

test/utility/iwpeps_rotation.jl
Yue-Zhengyuan: This version check is now added for the entire `test_rotation(psi)`.

Yue-Zhengyuan:
@lkdvos Should we update the format check with JuliaFormatter v2 (currently v1 is being used)? They give different result for a few files.

lkdvos:
I am holding off on that for now since it indeed gives different results for a few files.
My biggest annoyance with the new vesion is that it transforms this:
```julia
function myfunction(args..., ::Type{T}=Float64, ...)
end
```
into this:
```julia
function myfunction(args..., (::Type{T}=Float64), ...)
```

which I think is just silly. I'm happy to rediscuss formatters, but this really has to be in a separate single PR so the formatting changes don't get mixed up with actual coding changes

Yue-Zhengyuan:
I've undone the v2 formatting changes.

src/states/infiniteweightpeps.jl
```diff
@@ -136,6 +176,9 @@ end
 function InfiniteWeightPEPS(
     f, T, Pspaces::M, Nspaces::M, Espaces::M=Nspaces
 ) where {M<:AbstractMatrix{<:Union{Int,ElementarySpace}}}
+    @assert all(!isdual(Pspace) for Pspace in Pspaces)
+    @assert all(!isdual(Nspace) for Nspace in Nspaces)
+    @assert all(!isdual(Espace) for Espace in Espaces)
```
lkdvos, line 181:
```suggestion
    @assert all(!isdual, Pspaces)
    @assert all(!isdual, Nspaces)
    @assert all(!isdual, Espaces)
```

lkdvos (approved):
Looks okay to go for me.

src/states/infiniteweightpeps.jl
Yue-Zhengyuan: Didn't know this shorthand. Thanks!

lkdvos: This is actually true for many "mapping" functions (`sum`, `any`, ...) and it even avoids creating an intermediate generator which tends to be better for type stability (and thus sometimes also performance) :)

lkdvos (approved):

PR #184, opened 2025-04-24
Add CTMRG algorithm for layers of `PEPO` tensors

sanderdemeyer:
This builds upon #134, where the CTMRG algorithm now includes `InfiniteSquareNetwork`s without a top and bottom layer, where the top physical leg of the top `PEPOTensor` is contracted with the bottom physical leg of the bottom `PEPOTensor`. This is based on the type `PEPOTraceSandwich` (The equivalent of `PEPOSandwich` in the previous PR).

This PR can be summarised as

- Define the local sandwich type `PEPOTraceSandwich`
- Add a constructor for an `InfiniteSquareNetwork` of `PEPOTraceSandwich`es based on an `InfinitePEPO`
- Define the function `dagger` for an `InfinitePEPO`
- Add a CTMRG algorithm for a `PEPOTraceSandwich`
- Generalise the current `mpotensor` function for `PEPSSandwich` to both `PEPOSandwich`es and `PEPOTraceSandwich`es.
- Add a test where the expectation value for a double layer `PEPO` is compared to the case where this double layer is now squashed into a single layer
- Add a test where the expectation value for a contraction of a `PEPO` on a `PEPS` is compared to the case where this application is squashed into a single layer `PEPS` 

Any suggestions on the implementation and naming of things in this PR is very much appreciated. What anyway needs to be done, is

- [x] Add documentation for e.g. `dagger`
- [x] Settle on a naming convention
- [x] Maybe add an additional test on the different `mpotensor` functions that now exist


src/networks/local_sandwich.jl
```diff
@@ -104,3 +71,128 @@ function virtualspace(O::PEPOSandwich, dir)
...
+        fuse(virtualspace(network, WEST)),
+        virtualspace(network, WEST),
+    )
+    F_south = isomorphism(
+        storagetype(network[1]),
+        fuse(virtualspace(network, SOUTH)),
+        virtualspace(network, SOUTH),
+    )
+    return _mpotensor_contraction(
+        F_south, F_west, twist(F_south, H + 3), twist(F_west, H + 3), network
```
lkdvos, line 182: I don't think these twists are correct in the generic case, and I think you deleted the arrow and space checks that verify that this is the specific case for which the twists are correct 😉

sanderdemeyer: That's a valid point... I added them again now.

src/networks/local_sandwich.jl
```diff
@@ -104,3 +71,148 @@ function virtualspace(O::PEPOSandwich, dir)
         virtualspace(bra(O), dir)',
     ])
 end
+
+## Only PEPO layers
+# In a CTMRG contraction, the top physical leg of the top PEPOTensor is contracted with the bottom physical leg of the bottom PEPOTensor
+
+const PEPOLayersSandwich{N,P<:PEPOTensor} = Tuple{Vararg{P,N}}
```
leburgel, line 78: I think I prefer `PEPOTrace` as a name here (and then everywhere else correspondingly). It's not really a sandwich, since there's no well-defined slices on either side.

leburgel: Or at the very least `PEPOTraceSandwich`, I think the `Layers` does not add much clarity.

src/operators/infinitepepo.jl
```diff
@@ -171,6 +171,21 @@ function InfiniteSquareNetwork(top::InfinitePEPS, mid::InfinitePEPO, bot::Infini
...
+    return InfiniteSquareNetwork(map(tuple, eachslice(unitcell(mid); dims=3)...))
+end
+
+"""
+    dagger(O::PEPOTensor)
+
+Create the dagger of a PEPOTensor such that `InfinitePEPO(dagger(O))` is the adjoint of `InfinitePEPO(O)`.
+"""
+
+function dagger(O::PEPOTensor)
```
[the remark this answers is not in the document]
sanderdemeyer: I have now defined `dagger` for an `InfinitePEPO`, with the helper function `_dagger` for `PEPOTensor`. Is this ideal? No. But since we don't have a better name for this function, and since we will not export this function anyway, I will keep this name for now.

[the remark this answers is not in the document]
leburgel: I think this is fine for now. We just want to make the operation on the tensor look unappealing so people don't use it by accident, and `_dag` feels just dangerous enough to dissuade people. There's already a lot of similar equally bad names, so I can certainly live with this one.

sanderdemeyer:
I think everything should be fine now. And if @lkdvos is okay with the naming convention, I think we can merge this.

src/operators/infinitepepo.jl
```diff
@@ -171,6 +171,32 @@ function InfiniteSquareNetwork(top::InfinitePEPS, mid::InfinitePEPO, bot::Infini
...
+    return InfiniteSquareNetwork(map(tuple, eachslice(unitcell(mid); dims=3)...))
+end
+
+"""
+    _dag(O::PEPOTensor)
+
+Calculate the conjugate of an operator O, while permuting the physical indices.
+Flips and twists are included to ensure the correct arrow convention of a PEPOTensor.
+"""
+
```
leburgel, line 184:
```suggestion
```
Docstrings really need to be attached to function signatures without blank lines in between.

src/operators/infinitepepo.jl
```diff
@@ -171,6 +171,32 @@ function InfiniteSquareNetwork(top::InfinitePEPS, mid::InfinitePEPO, bot::Infini
...
+    @tensor O_conj[-1 -2; -3 -4 -5 -6] := conj(O[-2 -1; -3 -4 -5 -6])
+    return twist(flip(O_conj, [3 4 5 6]), [3 4])
+end
+
+"""
+    dagger(O::InfinitePEPO)
+
+Create the dagger of a PEPOTensor such that `InfinitePEPO(dagger(O))` is the adjoint of `InfinitePEPO(O)` with respect to the physical action of the PEPO on a PEPS.
+"""
+
```
leburgel, line 195:
```suggestion
```
Docstrings really need to be attached to function signatures without blank lines in between.

test/ctmrg/pepotrace.jl
```diff
@@ -0,0 +1,130 @@
...
+    ψ[1; 3 5 7 9] *
+    fuser[2 3; -3] *
+    fuser[4 5; -4] *
+    fuser_conj[6 7; -5] *
+    fuser_conj[8 9; -6]
+
+(Nr, Nc) = (1, 1)
+Oinf = InfinitePEPO(O; unitcell=(Nr, Nc, 1))
+O_stack = fill(O, Nr, Nc, 2)
+O_stack[:, :, 2] .= unitcell(dagger(Oinf))
```
leburgel, line 88:
```suggestion
O_stack[:, :, 2] .= unitcell(PEPSKit.dagger(Oinf))
```
Or add `dagger` as an import above.

leburgel (dismissed):
This looks good to me, @lkdvos what do you think? The functionality is a bit experimental, but I think it's mostly well defined what one would do with a stack of PEPOs (similar to the density operator support in MPSKit I think). There's no changes to the public interface and the new features are tested, so good enough for me at this point. Once the things you'll use is it for are better established, it would be great to also add an example at a later point @sanderdemeyer.

As a side not, but since it's again relevant here: I think at some point we should add more 'public' support for daggering and transposing transfer operators, both 3D ones like Sander uses here and 2D ones like the MPSKit MPOs. Now every time someone needs a left (i.e. bottom in our view) fixed point people usually do some shady manual intervention. Even though it's inherently a bit ill-defined, I think a properly documented `dagger` that exactly states what it does (and does not do) would be a valuable addition to the public interface in the future.

lkdvos (commented):
Looks more or less okay to me, left some small comments.

I am also not entirely sold on `PEPOTrace`, but I also cannot think of that many better options.
If this is only used for finite-temperature things I'd say to call it something based of Projected Entangled Pair Density Operator (PEPDO), in analogy to the Matrix Product Density Operator (MPDO), but I've only ever really seen the MPDO in the literature so I don't think it matters too much.

sanderdemeyer:
> Looks more or less okay to me, left some small comments.
> 
> I am also not entirely sold on `PEPOTrace`, but I also cannot think of that many better options. If this is only used for finite-temperature things I'd say to call it something based of Projected Entangled Pair Density Operator (PEPDO), in analogy to the Matrix Product Density Operator (MPDO), but I've only ever really seen the MPDO in the literature so I don't think it matters too much.

The finite-temperature stuff is indeed why I am adding this. In principle, this code is not restricted to this case and could be applied to PEPOs, so I prefer `PEPOTrace` to `PEPDO` for that reason.

src/operators/infinitepepo.jl
```diff
@@ -218,6 +242,19 @@ function ChainRulesCore.rrule(
     return network, InfiniteSquareNetwork_pullback
 end
 
+function ChainRulesCore.rrule(
+    ::Type{InfiniteSquareNetwork}, mid::InfinitePEPO{P}
+) where {P<:PEPOTensor}
+    network = InfiniteSquareNetwork(top, mid, bot)
```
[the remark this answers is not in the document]
sanderdemeyer: I've now changed this. The problem with testing this (and other functions in this PR as well) is that I use this for the (cluster expansion) finite-temperature states, for which the code is not yet complete. When this is finished, more rigorous tests can be added, as opposed to now just adding trivial tests to ensure the code runs.

src/operators/infinitepepo.jl
```diff
@@ -171,6 +171,30 @@ function InfiniteSquareNetwork(top::InfinitePEPS, mid::InfinitePEPO, bot::Infini
...
+"""
+function _dag(O::PEPOTensor)
+    @tensor O_conj[-1 -2; -3 -4 -5 -6] := conj(O[-2 -1; -3 -4 -5 -6])
+    return twist(flip(O_conj, [3 4 5 6]), [3 4])
+end
+
+"""
+    dagger(O::InfinitePEPO)
+
+Create the dagger of a PEPOTensor such that `InfinitePEPO(dagger(O))` is the adjoint of `InfinitePEPO(O)` with respect to the physical action of the PEPO on a PEPS.
```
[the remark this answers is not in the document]
sanderdemeyer: Yes, you are correct, I forgot to change this. 

For me, changing this to `adjoint` is fine (and even slightly preferred), if @leburgel is also fine with this.

src/operators/infinitepepo.jl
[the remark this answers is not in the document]
lkdvos: I see what you mean, I'm just not a huge fan of adding functions that have never been run and have no purpose yet, since as demonstrated by this exact example it is very easy to have this result in errors. I would prefer to either not include it in this PR and add it along with the features where it is being used, or already include some tests to ensure that this functionality is correct.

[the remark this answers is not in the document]
leburgel: I agree with Lukas about not adding this if there's no use case we can currently test. We've had quite a few broken things added without a use case already that we didn't notice until later, and that's very annoying. Removing this here and then just adding it in again later may feel a bit silly, but I think it's the right thing to do. I think this is mainly about adding contraction support anyway, so unless you have a concrete test that also uses AD through PEPO traces I would remove this for the time being.

src/operators/infinitepepo.jl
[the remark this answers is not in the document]
leburgel: The docstring is indeed kind of wrong (I missed this before), it's meant to be the dagger of the entire `InfinitePEPO` as a map between physical indices. Therefore it does not perform an actual adjoint on the individual tensors. I would also be okay with overloading `adjoint` in principle, I just thought a different name might be appropriate since there's some ambiguity about what to do with the virtual spaces of the individual tensors (flip them back to the original directions or not, probably some twists missing for fermions...). So I think there's very clear use cases, but it might not be entirely mathematically solid and practically robust in all cases in its current form, which is why I thought it wasn't a bad idea to put it under a less standard name.

src/operators/infinitepepo.jl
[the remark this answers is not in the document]
sanderdemeyer: These are valid points. Since I can't think of a case where this will eventually be used, I removed the function. We'll add it again if it ever becomes useful.

src/operators/infinitepepo.jl
```diff
@@ -187,11 +187,11 @@ function _dag(O::PEPOTensor)
...
 
 """
-    dagger(O::InfinitePEPO)
+    adjoint(O::InfinitePEPO)
 
-Create the dagger of a PEPOTensor such that `InfinitePEPO(dagger(O))` is the adjoint of `InfinitePEPO(O)` with respect to the physical action of the PEPO on a PEPS.
+Create the adjoint of an InfinitePEPO.
 """
-function dagger(O::InfinitePEPO)
+function TensorKit.adjoint(O::InfinitePEPO)
```
lkdvos, line 194:
```suggestion
function Base.adjoint(O::InfinitePEPO)
```
This function is actually defined in base, otherwise this is some kind of strange method piracy.
Given that this is the case, and that this docstring doesn't really provide more information than the Base docstrings I would also just drop the docstring

leburgel: I agree the docstring doesn't add much, but I would prefer to exapand it rather than remove it. The adjoint with respect to the physical action of a PEPO on a PEPS is not that trivial I think, especially since both physical PEPO tensor indices are in the codomain from a tensormap point of view.

src/operators/infinitepepo.jl
```diff
@@ -171,6 +171,30 @@ function InfiniteSquareNetwork(top::InfinitePEPS, mid::InfinitePEPO, bot::Infini
...
+
+"""
+    _dag(O::PEPOTensor)
+
+Calculate the conjugate of an operator O, while permuting the physical indices.
+Flips and twists are included to ensure the correct arrow convention of a PEPOTensor.
+"""
+function _dag(O::PEPOTensor)
+    @tensor O_conj[-1 -2; -3 -4 -5 -6] := conj(O[-2 -1; -3 -4 -5 -6])
+    return twist(flip(O_conj, [3 4 5 6]), [3 4])
```
leburgel, line 186: Again, sorry for missing this before, but why are you flipping the virtual spaces here? I don't think this is what you should do, and then you also shouldn't add twists on the virtual spaces after that (these twists were only correct for specific space configurations anyway, similar to the `mpotensor` converter case).

For reference, if you look at Eq. (284) and (231) [here](https://arxiv.org/pdf/2404.14611), the virtual spaces are not flipped, and the twists appear on the physical spaces. For the PEPO case, I think you would have a twist on the 'domain' physical space if it is not dual in the tensormap codomain, and on the 'codomain' physical space if it is dual in the tensormap codomain.

leburgel: You only need to flip the virtual spaces back when you're trying to see if the PEPO as a whole is Hermitian by comparing local tensors, at which point you should be careful with the flips and twists. But this is not needed when just computing the full adjoint operator.

sanderdemeyer: I changed this and added a test

sanderdemeyer:
Several comments from above are now included. To try to resolve the discussion about the naming convention, i.e. to make sure that what we're doing is truly the (mathematically rigorous) adjoint, I added a test to verify that 
$$ \left< O \psi_1, \psi_2 \right> = \left< \psi_1, adjoint(O( \psi_2 \right> $$.
I also added this to the docstring.

To make sure that all the twists are correctly implemented, I also changed the test to the fermionic case.

sanderdemeyer:
After some minor fixes, I think it should now be okay if everyone is fine with the naming conventions and the current tests @lkdvos and @leburgel.

test/ctmrg/pepotrace.jl
lkdvos, line 1: There is quite a lot of code in the global scope here, which effectively means that most of it is type-unstable. This isn't necessarily a big problem, but given that our tests are quite computationally demanding this might not be that great.

It could be improved by turning some of it into functions, but I'm also slightly confused about what the tests are actually doing. How come there are no twists in any of the fusing contractions?

test/ctmrg/pepotrace.jl
```diff
@@ -0,0 +1,179 @@
...
+projector_algs = [:halfinfinite, :fullinfinite]
+
+@testset "mpotensor for PEPOTraceSandwich" begin
+    # Test whether the mpotensor of a double layer of PEPOs is the same as in the case where the layers are fused
+    mpo = InfiniteSquareNetwork(map(PEPSKit.mpotensor, PEPSKit.unitcell(network)))
+    mpo_fused = InfiniteSquareNetwork(
+        map(PEPSKit.mpotensor, PEPSKit.unitcell(network_fused))
+    )
+    @test mpo ≈ mpo_fused
+end
```
lkdvos, line 103: Would it not make more sense to check that these networks have the same values, rather than checking the tensors?

test/ctmrg/pepotrace.jl
```diff
@@ -0,0 +1,179 @@
...
+    )
+    env_fused, = leading_boundary(
+        CTMRGEnv(network_fused, envspace), network_fused; alg, maxiter=250, projector_alg
+    )
+
+    m = PEPSKit.contract_local_tensor((1, 1, 1), M, network, env)
+    m_fused = PEPSKit.contract_local_tensor((1, 1, 1), MO, network_fused, env_fused)
+
+    nrm = PEPSKit._contract_site((1, 1), network, env)
+    nrm_fused = PEPSKit._contract_site((1, 1), network_fused, env_fused)
```
lkdvos, line 122: If this is the actual norm, why can't you call `norm` explicitly? Why do we have to manually divide by the norm, is this not all equivalent to `expectation_value`?

test/ctmrg/pepotrace.jl
```diff
@@ -0,0 +1,179 @@
...
+    @test (m / nrm) ≈ (m_fused / nrm_fused) atol = 1e-9
+end
+
+projector_alg = projector_algs[1] # only use :halfinfinite for this test due to convergence issues
+@testset "Test adjoint of an InfinitePEPO using $alg with $projector_alg" for alg in
+                                                                              ctm_styles
+    # Test the definition of the adjoint of an operator, i.e. <Oψ₁, ψ₂> = <ψ₁, adjoint(O)ψ₂>, with Oψ the physical action of the PEPO O on the PEPS ψ.
```
lkdvos, line 129: Same comment about this being the mathematicians notation for adjoints. Also, could you add the case where this is equal to the peps-pepo-peps contraction?

lkdvos: only now noticed that this is probably done below, it might make more sense to combine all of these?

test/ctmrg/pepotrace.jl
sanderdemeyer: I don't think that there is currently a norm defined for a generic `InfiniteSquareNetwork` apart from the trivial norm of the tensor itself and the specific case of a PEPS. This would also not be well defined in the generic case, which is why I would be okay with keeping the current version. The `expectation_value` of an operator (either a PEPO or `LocalOperator`) of a `PEPOTraceSandwich` could be defined, but this is another PR (see #117).

Writing this down as an `expectation_value` of an `InfinitePartitionFunction` would require us to fuse the layers, which would defeat the purpose of the test.

test/ctmrg/pepotrace.jl
sanderdemeyer: Here, it satisfies the stricter test, but indeed this does not have to be the case. I've changed this now.

test/ctmrg/pepotrace.jl
sanderdemeyer: In the last test we check whether the expectation values of a PEPS-PEPO-PEPO-PEPS construct are returned correctly. This is something different than checking whether the definition of the adjoint holds. But the extra check that it is also equal to dot(psi, O, phi) is indeed a good addition.

test/ctmrg/pepotrace.jl
sanderdemeyer: The fusers do have a twist, but only when the arrows point the opposite direction than the conventional peps. That is why there are 
twists in 
`@tensor Odagϕ[-1; -3 -4 -5 -6] :=
    twist(PEPSKit._dag(O), 1:4)[-1 1; 2 4 6 8] *
    ϕ[1; 3 5 7 9] *
    fuser_adj[2 3; -3] *
    fuser_adj[4 5; -4] *
    conj(fuser_adj[6 7; -5]) *
    conj(fuser_adj[8 9; -6])
`
on the virtual legs of $$ O^\dagger $$, but not in
`@tensor Oψ[-1; -3 -4 -5 -6] :=
    O[-1 1; 2 4 6 8] *
    ψ[1; 3 5 7 9] *
    fuser[2 3; -3] *
    fuser[4 5; -4] *
    conj(fuser[6 7; -5]) *
    conj(fuser[8 9; -6])
`

sanderdemeyer: I've moved some code around to reduce the amount of code in the global scope. There is still some left, but I also don't want to recalculate too much in every test, or merge tests that have different goals.

test/ctmrg/pepotrace.jl
sanderdemeyer: Another small comment, I've changed the test to check `dot(psi, O', phi) == dot(psi, O' * phi) == dot(O * psi, phi)` instead of `dot(psi, O, phi) == dot(psi, O * phi) == dot(O' * psi, phi)`. This is of course equivalent, but it fits better in the tests and turned out to be a bit easier to converge.

sanderdemeyer:
I've taken care of the new comments. As a summary, I've changed the docstring to the `dot` notation, reduced the amount of code in the global scope, and changed the tests to compare the `network_value`s of the mpos instead of the tensors themselves.

test/ctmrg/pepotrace.jl
```diff
@@ -0,0 +1,190 @@
...
+I = fℤ₂
+pspace = Vect[I](0 => 1, 1 => 1)
+vspace = Vect[I](0 => 1, 1 => 1)
+envspace = Vect[I](0 => χenv, 1 => χenv)
+
+Random.seed!(97646475)
+
+# Construct random PEPO tensors
+O = randn(T, pspace ⊗ pspace', vspace ⊗ vspace ⊗ vspace' ⊗ vspace')
+O = O + twist(flip(PEPSKit._dag(O), 3:6), [4 6])
```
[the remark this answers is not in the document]
sanderdemeyer: I would indeed have liked to omit this line. The problem is that a random PEPO is quite hard to converge. For the (physical) operators that I am computing in the finite-temperature case, they all converge nicely. Since these are not in PEPSKit, I can unfortunately not use them for these tests. For the bosonic case, the convergence is also easier.

I think the tests are still checking the definition of the adjoint, even though it's a hermitian operator (especially in the fermionic case, there are some nontrivial twists). If the `_dag` function were incorrect, this would just correspond to a different initialisation of the tensor `O`, and the tests would fail)

test/ctmrg/pepotrace.jl
```diff
@@ -0,0 +1,190 @@
...
+
+    nrm = PEPSKit._contract_site((1, 1), network, env)
+    nrm_fused = PEPSKit._contract_site((1, 1), network_fused, env_fused)
+    @test (m / nrm) ≈ (m_fused / nrm_fused) atol = 1e-9
+end
+
+projector_alg = projector_algs[1] # only use :halfinfinite for this test due to convergence issues
+@testset "Test adjoint of an InfinitePEPO using $alg with $projector_alg" for alg in
+                                                                              ctm_styles
+    # Test the definition of the adjoint of an operator, i.e. dot(psi, O', phi) == dot(psi, O' * phi) == dot(O * psi, phi) for any two states psi and phi.
```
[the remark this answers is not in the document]
sanderdemeyer: I don't really get what the problem is with the mathematical convention instead of the physical one. Isn't this just a convention change from $$O$$ to $$O^\dagger$$? Again the same comment about the convergence for $$O^\dagger$$ holds...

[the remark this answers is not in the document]
lkdvos: As the name says, it is definitely a convention, but I would say that it makes a lot more sense to use the physics convention given that the application is physics, or is that crazy? I'm happy to hear if there is a good reason to prefer it the other way around, but it does seem nice to remain consistent.

test/ctmrg/pepotrace.jl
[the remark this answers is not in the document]
sanderdemeyer: There reason is that the code is present in [my code for cluster expansions](https://github.com/sanderdemeyer/ClusterExpansions), which would be a relatively big addition to PEPSKit (and to be fully correct, requires some functionality for BigFloats that either requires my personal fork of TensorKit (since this is not yet merged, and maybe never will) or, in the long run, maybe MatrixAlgebraKit).

I am still in doubt whether the long-term plan is to add this to PEPSKit, or to keep it as a different package to put on the QuantumKitHub. It will anyway still require a large effort to make everything consistent with PEPSKi.

test/ctmrg/pepotrace.jl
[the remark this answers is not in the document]
sanderdemeyer: Well, I definitely prefer the mathematical convention, since that is how I learned about adjoints, and so the 'physics convention' seems a bit weird to me. I was also not aware that there were even different conventions, because it seems to me the convention only applies to how you would write the defining equation for the adjoint, and has no influence on the definition itself. 

At first glance, I also didn't find anything in PEPSKit or TensorKit where we already chose a convention for this. Is this something of Julia itself?

test/ctmrg/pepotrace.jl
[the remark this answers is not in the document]
lkdvos: Why not generate the PEPO there and hardcode the entries in the tests here then?

test/ctmrg/pepotrace.jl
```diff
@@ -0,0 +1,190 @@
...
+    O_stack = fill(O, Nr, Nc, 2)
+    O_stack[:, :, 2] .= PEPSKit.unitcell(adjoint(Oinf))
+    OOdag = InfinitePEPO(O_stack)
+
+    Oψ = fuse_PEPO_PEPS(O, ψ, fuser)
+    Mψ = fuse_PEPO_PEPS(M, ψ, fuser)
+
+    network_O = InfiniteSquareNetwork(InfinitePEPS(ψ), OOdag, InfinitePEPS(ψ))
+    network_fused_MO = InfiniteSquareNetwork(InfinitePEPS(Mψ), InfinitePEPS(Oψ))
+    network_fused_OO = InfinitePEPS(Oψ)
```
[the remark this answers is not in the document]
sanderdemeyer: It is not, `InfinitePEPS(Mpsi)` would indeed correspond to replacing M at every site. That is why this is not used in the calculation of an environment. We only need this because the function `network_value` requires a `InfinitePEPS` or `InfiniteSquareNetwork` and not just a single tensor.

test/ctmrg/pepotrace.jl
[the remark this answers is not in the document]
sanderdemeyer: That is also a possibility, but this is at least a 2x2x5x5x5x5 fermionic tensor (and I checked, with 184 nonzero elements) and would be quite a hassle to implement. Not to mention that it would be extremely ugly, since the elements are derived from exponentials and are not 'pretty' numbers.

lkdvos:
@ogauthe do you maybe have some reference for some high temperature expansion results we could use to verify this implementation on? Alternatively, do you have any inspiration for any testcases that we could use?

lkdvos:
> Therefore, I would opt for using Simple Update to produce some finite-temperature state, and then compare with figure 8a or 10a of [this paper](https://arxiv.org/abs/1311.7272).

Sounds like a solid plan. I read through that paper, it seems like they are just using the fact that it is free fermions to obtain exact expressions and compare to that. This sounds like an ideal case to put in our tests, since it isn't trivial but also we are certain of the desired output.

For the expectation values, while I get your point, it is a bit strange to tackle that first since `expectation_value(::InfinitePEPO, op, ::CTMRGEnv)` really needs a `CTMRGEnv`, which we cannot obtain without first implementing the CTMRG.
Am I understanding it correctly that for your implementation, since you are converting to a partition function, you would have to provide the CTMRGEnv for the partition function, which only coincides when there is a single layer?

I think that actually the reduced density matrix approach might be a better fit after all. It should be somewhat straightforward to generate the contractions for building these reduced density matrices, after which applying operators really is just `tr(O * rho) / tr(rho)`. This turns out to be a lot cheaper as well, since we don't have to perform two contractions (numerator and denominator), and I verified that the leading cost of computing the density matrix is actually equal to the leading cost of immediately applying an operator, so there really is no lost performance.

I do like the idea as well of having additional consistency checks between the different algorithms: For example turning the density matrix into a PEPS by fusing the legs, we can evaluate expectation values that correspond to the partition function at `2beta`. Similarly, the network value (partition function) should be the same for the partition function, the PEPS, and the PEPO approach, which provides good checks.

I'm not entirely convinced that the `expectation_value` of a partition function is something we want to throw in the mix already, mostly because it seems like the interface still is missing something. It would be great if we could formulate it in a similar framework (density-matrix-like, so we would obtain an object we can then use to evaluate arbitrary operators), but I would have to think through how (and if?) that translates well. It sounds a lot like source terms etc, but I don't have that knowledge directly available.

Yue-Zhengyuan:
@sanderdemeyer @leburgel @lkdvos What's the status of this PR? I realize this will be needed to measure Green's function like $\langle e^{-\beta H/2} a_i e^{-\beta H/2} b_j \rangle$, which involves two layers of iPEPO. (If purification approach is to be used, it will even involve 4 layers of iPEPO)

PR #186, opened 2025-04-29
Fix environment constructor scalar type

lkdvos:
This is a minor change that preserves the scalartype of the network for the `CTMRGEnv` constructor


pbrehmer (approved):
Good catch! Thanks for taking care of that :-)

PR #191, opened 2025-05-07
Rewrite svd to avoid empty blocks

### LINKED ISSUE #185: Stack on an empty collection for iterative svd
https://github.com/QuantumKitHub/PEPSKit.jl/blob/994a94f8d0069490aac9b106d56dadfd587e50a8/src/utility/svd.jl#L265C1-L284

When the truncation space has less sector than the domain of the TensorMap f, there is an error
'''
ERROR: LoadError: ArgumentError: `stack` on an empty collection is not allowed
'''

lkdvos:
This avoids looping over blocks that will end up being truncated completely, ie that would get `howmany = 0`.

This probably fixes #185, although I don't have an explicit case to check this on.

### DIFF
--- a/src/utility/svd.jl
+++ b/src/utility/svd.jl
@@ -286,7 +286,9 @@ function TensorKit._compute_svddata!(
     I = sectortype(f)
     dims = SectorDict{I,Int}()
 
-    generator = Base.Iterators.map(blocks(f)) do (c, b)
+    sectors = trunc isa NoTruncation ? blocksectors(f) : blocksectors(trunc.space)
+    generator = Base.Iterators.map(sectors) do c
+        b = block(f, c)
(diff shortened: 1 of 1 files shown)


Confusio:
> This avoids looping over blocks that will end up being truncated completely, ie that would get `howmany = 0`.
> 
> This probably fixes #185, although I don't have an explicit case to check this on.

Thanks for the fixing. This will come out if we set `svd_alg=SVDAdjoint(; fwd_alg=IterSVD())` in the `boundary_alg` in the following example:
https://github.com/QuantumKitHub/PEPSKit.jl/blob/72547471233955defb07dc5d365315ab753ad11d/examples/xxz/main.jl#L74

pbrehmer (approved):
Thanks for catching and fixing that!

Issue #192, opened 2025-05-08
About broadening


Confusio:
https://github.com/QuantumKitHub/PEPSKit.jl/blob/7cfae1aad1578dd0f51346cbf20d18f7c76dbf1b/src/utility/svd.jl#L442-L448

In the above lines, when the singular values are degenerate, the `inv` is set to zero. When they are not degenerate, the broadening is applied. Is this the purpose for broadening? I think it is in the opposite.

pbrehmer:
Thanks for bringing that up! I think you're right that when the singular values are degenerate, such that `x` is approximately zero, we want to apply broadening instead of setting it to zero, which we don't do here - that should definitely be fixed.

On the other hand, we also want to catch the case where singular values might be really small such that taking the inverse directly would be a bad idea. So perhaps we would need a separate function for taking inverses with broadening that only applies to singular value differences.

I'm wondering, if it would make a difference to only apply broadening to the singular value differences that fall below a certain tolerance. I thought that applying the broadening to singular values that are not degenerate will only have a small or negligible effect. @lkdvos What do you think?

Yue-Zhengyuan:
> In the above lines, when the singular values are degenerate, the `inv` is set to zero

This is only when all singular values are very small, as Paul said. I think this is effectively doing truncation of the SVD when changing the dimension of the SVD result tensors is not convenient.

lkdvos:
There might also be some implementation detail here: we are abusing this function to simply take `_safe_inv.(S' .- S)`, counting on the fact that the diagonal entries are zero and should remain zero, and only the off-diagonal entries are relevant. I would have to check in more detail how this function is used throughout the rest of the code though

Confusio:
As a follow up suggestion, during the `ctmrg`, can we first use some fast svd to get a `env`, then switch to `sdd` for a better 'env' and finally 'svd'? The reason of 'svd' is for the improvement of the gradient.

pbrehmer:
That is an interesting suggestion. Although I do think that for many purposes the `sdd` algorithm should be sufficient since we often don't need CTMRG gradients up to such a high precision that the `svd` would be a significant improvement. Of course there are pathological cases where `sdd` might fail but I don't think these are super common? I'll keep the suggestion in mind but for now I'll have to work on something else :-)

In any case, #194 is now merged so that should hopefully take care of the Lorentzian broadening issue.

PR #194, opened 2025-05-09
Update SVD reverse-rule broadening

pbrehmer:
Initiated from #192, this PR updates/fixes the Lorentzian broadening in the SVD reverse-rule. I also updated the corresponding test where we can see that degenerate singular values lead to wrong gradients if there is no broadening and *no cutoff*.

In the tests on random matrices it seems that using a cutoff for the degenerate inverse singular value differences - thereby just setting them to zero - kind of has the same effect as the broadening.

### DIFF
--- a/src/utility/svd.jl
+++ b/src/utility/svd.jl
@@ -444,7 +444,7 @@ _safe_inv(x, tol) = abs(x) < tol ? zero(x) : inv(x)
 # compute inverse singular value difference contribution to SVD gradient with broadening ε
 function _broadened_inv_S(S::AbstractVector{T}, tol, ε=0) where {T}
     F = Matrix{T}(undef, length(S), length(S))
-    @inbounds for i in axes(F, 1), j in axes(F, 2)
+    @inbounds for j in axes(F, 2), i in axes(F, 1)
         F[i, j] = if i == j
             zero(T)
         else
--- a/src/utility/svd.jl
+++ b/src/utility/svd.jl
@@ -593,8 +593,8 @@ function svd_pullback!(
         end
 
         X = @. (1//2) * (
-            (UrΔU + VrΔV) * _safe_inv(Sp' .- Sr, tol) +
-            (UrΔU - VrΔV) * _safe_inv(Sp' .+ Sr, tol)
+            (UrΔU + VrΔV) * _safe_inv(Sp' - Sr, tol) +
+            (UrΔU - VrΔV) * _safe_inv(Sp' + Sr, tol)
         )
         Y = @. (1//2) * (
             (UrΔU + VrΔV) * _safe_inv(Sp' .- Sr, tol) -
--- a/src/utility/svd.jl
+++ b/src/utility/svd.jl
@@ -438,13 +438,21 @@ function ChainRulesCore.rrule(
     return (U, S, V, info), tsvd!_itersvd_pullback
 end
 
-# scalar inverses with a cutoff tolerance and Lorentzian broadening
-function _safe_inv(x, tol, ε=0)
-    if abs(x) < tol
-        return zero(x)
-    else
-        return iszero(ε) ? inv(x) : _lorentz_broaden(x, ε)
+# scalar inverses with a cutoff tolerance
+_safe_inv(x, tol) = abs(x) < tol ? zero(x) : inv(x)
+
+# compute inverse singular value difference contribution to SVD gradient with broadening ε
+function _broadened_inv_S(S::AbstractVector{T}, tol, ε=0) where {T}
+    F = Matrix{T}(undef, length(S), length(S))
+    @inbounds for i in axes(F, 1), j in axes(F, 2)
+        F[i, j] = if i == j
+            zero(T)
+        else
+            Δsᵢⱼ = S[j] - S[i]
+            ε > 0 ? _lorentz_broaden(Δsᵢⱼ, ε) : _safe_inv(Δsᵢⱼ, tol)
+        end
     end
+    return F
 end
 
 # Lorentzian broadening for divergent term in SVD rrule, see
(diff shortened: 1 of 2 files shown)


src/utility/svd.jl
```diff
@@ -438,13 +438,21 @@ function ChainRulesCore.rrule(
...
-        return zero(x)
-    else
-        return iszero(ε) ? inv(x) : _lorentz_broaden(x, ε)
+# scalar inverses with a cutoff tolerance
+_safe_inv(x, tol) = abs(x) < tol ? zero(x) : inv(x)
+
+# compute inverse singular value difference contribution to SVD gradient with broadening ε
+function _broadened_inv_S(S::AbstractVector{T}, tol, ε=0) where {T}
+    F = Matrix{T}(undef, length(S), length(S))
+    @inbounds for i in axes(F, 1), j in axes(F, 2)
```
lkdvos, line 447:
```suggestion
    @inbounds for j in axes(F, 2), i in axes(F, 1)
```

Inner loops should be columns 😉

src/utility/svd.jl
```diff
@@ -438,13 +438,21 @@ function ChainRulesCore.rrule(
...
-    if abs(x) < tol
-        return zero(x)
-    else
-        return iszero(ε) ? inv(x) : _lorentz_broaden(x, ε)
+# scalar inverses with a cutoff tolerance
+_safe_inv(x, tol) = abs(x) < tol ? zero(x) : inv(x)
+
+# compute inverse singular value difference contribution to SVD gradient with broadening ε
+function _broadened_inv_S(S::AbstractVector{T}, tol, ε=0) where {T}
+    F = Matrix{T}(undef, length(S), length(S))
```
lkdvos, line 446:
```suggestion
    F = similar(S, (axes(S, 1), axes(S, 1)))
```
I'm okay with this being hardcoded to `length` and `Matrix` as well, but then the signature should probably be `Vector{T}` instead of `AbstractVector` (or at least `DenseVector`).

src/utility/svd.jl
```diff
@@ -554,9 +562,8 @@ function svd_pullback!(
         @info "`svd` cotangents sensitive to gauge choice: (|Δgauge| = $Δgauge)"
     end
 
-    UdΔAV =
-        (aUΔU .+ aVΔV) .* _safe_inv.(Sp' .- Sp, tol, broadening) .+
-        (aUΔU .- aVΔV) .* _safe_inv.(Sp' .+ Sp, tol)
+    inv_S_minus = _broadened_inv_S(Sp, tol, broadening) # possibly divergent/broadened contribution
+    UdΔAV = @. (aUΔU + aVΔV) * inv_S_minus + (aUΔU - aVΔV) * _safe_inv(Sp' .+ Sp, tol)
```
lkdvos, line 566: Note that this has a slightly different implementation than before: this is actually instantiating an intermediate array, whereas previously everything was fused into a single computation.
I'm not entirely sure if this really matters though, just wanted to mention that.

src/utility/svd.jl
```diff
@@ -585,16 +592,14 @@ function svd_pullback!(
...
-                (UrΔU .- VrΔV) .* _safe_inv.(Sp' .+ Sr, tol)
-            )
-        Y =
-            (1//2) .* (
-                (UrΔU .+ VrΔV) .* _safe_inv.(Sp' .- Sr, tol, broadening) .-
-                (UrΔU .- VrΔV) .* _safe_inv.(Sp' .+ Sr, tol)
-            )
+        X = @. (1//2) * (
+            (UrΔU + VrΔV) * _safe_inv(Sp' .- Sr, tol) +
+            (UrΔU - VrΔV) * _safe_inv(Sp' .+ Sr, tol)
```
lkdvos, line 597:
```suggestion
            (UrΔU + VrΔV) * _safe_inv(Sp' - Sr, tol) +
            (UrΔU - VrΔV) * _safe_inv(Sp' + Sr, tol)
```

src/utility/svd.jl
```diff
@@ -585,16 +592,14 @@ function svd_pullback!(
...
+        )
+        Y = @. (1//2) * (
+            (UrΔU + VrΔV) * _safe_inv(Sp' .- Sr, tol) -
+            (UrΔU - VrΔV) * _safe_inv(Sp' .+ Sr, tol)
```
lkdvos, line 601:
```suggestion
            (UrΔU + VrΔV) * _safe_inv(Sp' - Sr, tol) -
            (UrΔU - VrΔV) * _safe_inv(Sp' + Sr, tol)
```

src/utility/svd.jl
```diff
@@ -585,16 +592,14 @@ function svd_pullback!(
...
-                (UrΔU .+ VrΔV) .* _safe_inv.(Sp' .- Sr, tol, broadening) .+
-                (UrΔU .- VrΔV) .* _safe_inv.(Sp' .+ Sr, tol)
-            )
-        Y =
-            (1//2) .* (
-                (UrΔU .+ VrΔV) .* _safe_inv.(Sp' .- Sr, tol, broadening) .-
-                (UrΔU .- VrΔV) .* _safe_inv.(Sp' .+ Sr, tol)
-            )
+        X = @. (1//2) * (
+            (UrΔU + VrΔV) * _safe_inv(Sp' .- Sr, tol) +
```
lkdvos, line 596: You seem to have removed a broadening here, was this on purpose?

src/utility/svd.jl
```diff
@@ -585,16 +592,14 @@ function svd_pullback!(
...
+            (UrΔU - VrΔV) * _safe_inv(Sp' .+ Sr, tol)
+        )
+        Y = @. (1//2) * (
+            (UrΔU + VrΔV) * _safe_inv(Sp' .- Sr, tol) -
```
lkdvos, line 600: Same comment about removed broadening here.

src/utility/svd.jl
pbrehmer: Whoopsie, rookie mistake! :)

src/utility/svd.jl
pbrehmer: I'll leave the `AbstractVector` since that is also the type restriction from `svd_pullback!` and then rather use the `similar` constructor

src/utility/svd.jl
pbrehmer: I found it easier to instantiate the `inv_S_minus` intermediate array since the `_broadened_inv_S` function explicitly iterates over the axes of `Sp`. I thought it would probably not matter that much.

lkdvos (approved):
I'm okay with merging this, but keep in mind that I have very little experience or tests to support a good intuition for whether or not these changes are better, worse, or don't really change anything at all. Perhaps a second opinion could be nice, but I also guess it can't hurt too much to merge this either.

pbrehmer:
I would think that these changes overall have only minor consequences. But at least we now test a case with degenerate singular values where we do see that the current changes do what they are supposed to do. The thing that is kind of unclear is whether the broadening is an improvement/deterioration over just having a cutoff (which sets the divergent contributions to zero). I talked with Anna & Bram and they do something very similar but for the truncated contribution in the SVD rrule they solve a linear problem (kind of KrylovKit-style) instead of using the truncated bits from the full SVD - so there it is kind of unclear what the broadening should do.

In any case, overall I'm also okay with merging this but it's good to keep in mind the places where we're still unsure.

PR #203, opened 2025-05-26
Add full update bond environment and gauge fixing

Yue-Zhengyuan:
This PR adds the bond environment for nearest-neighbor full update, which is expressed using the CTM tensors. For the horizontal bond between sites at `[r, c]` and `[r, c+1]`, the bond environment `env` is 
```
    C1 --- T1 --------- T1 --- C2     r-1
    |      ‖             ‖     |
    T4 === XX ==     == YY === T2     r
    |      ‖             ‖     |
    C4 --- T3 --------- T3 --- C3     r+1
    c-1    c           c+1    c+2
```
where `XX = X' X` and `YY = Y' Y`, and `X`, `Y` are Q-tensors (`PEPSOrth`) obtained from `_qr_bond`. 

However, this simple construction has two problems.

- Because of the approximate nature of CTMRG, `env` is not Hermitian.
- For physical (not random) iPEPS, its condition number (when solving equations to obtain the updated bond tensors) can be quite large (~1e+5 to 1e+7). 

Thus this PR also implements a "gauge fixing" algorithm introduced in [PhysRevB.92.035142](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.92.035142) (it can also be applied to NTU later), which does the following:

- Replace the bond environment `env` with `(env + env') / 2`; in addition, eigen-decomposition is used to remove possible small negative eigenvalues of `env` to make it semi-positive-definite. 
- Use QR decomposition to "fix gauge" of `env` and greatly reduce its condition number (to ~1e+1).


Yue-Zhengyuan:
@lkdvos I hope to get this merged before the NTU bondenv (#144) since I think FU may be a more widely acknowledged algorithm.

src/algorithms/contractions/bondenv/benv_ctm.jl
```diff
@@ -0,0 +1,77 @@
...
+        c3[χ9 χ7] *
+        t2[χ10 DEY0 DEY1 χ9] *
+        c2[χ8 χ10] *
+        t3Y[χ7 DSY0 DSY1 χ5] *
+        Y[DNY0 DEY0 DSY0 DY0] *
+        conj(Y[DNY1 DEY1 DSY1 DY1]) *
+        t1Y[χ6 DNY0 DNY1 χ8]
+    )
+    # combine
+    @autoopt @tensor benv[DX1 DY1; DX0 DY0] := (lhalf[DX1 DX0 χ5 χ6] * rhalf[DY1 DY0 χ5 χ6])
```
lkdvos, line 74: Is it possible to do the contraction in a single `@tensor` call, ie without intermediate tensors? This should allow `@autoopt` to optimize the contraction order over the entire contraction.

src/algorithms/contractions/bondenv/gaugefix.jl
```diff
@@ -0,0 +1,114 @@
...
+    ┌-|    |-- 1  2 --┐     ┌-- Z†-- 1  2 --┐
+    | └----┘          |     |               |
+    └-----------------┘     └---------------┘
+```
+"""
+function positive_approx(benv::BondEnv)
+    # eigen-decomposition: benv = U D U'
+    D, U = eigh((benv + benv') / 2)
+    # determine if `env` is (mostly) positive or negative
+    sgn = sign(mean(D.data))
```
lkdvos, line 18:
```suggestion
    sgn = sign(sum(D.data))
```

src/algorithms/contractions/bondenv/gaugefix.jl
```diff
@@ -0,0 +1,114 @@
...
+    # When optimizing the truncation of a bond, 
+    # its environment can always be multiplied by a number.
+    # If `benv` is negative (e.g. obtained approximately from CTMRG), 
+    # we can multiply it by (-1).
+    (sgn == -1) && (D *= -1)
+    # set (small) negative eigenvalues to 0
+    D.data .= max.(D.data, 0)
+    Z = sdiag_pow(D, 0.5) * U'
```
lkdvos, line 26:
```suggestion
    data = D.data
    @inbounds for i in eachindex(data)
        d = sgn == -1 ? -data[i] : data[i]
        data[i] = ifelse(d > 0, sqrt(d), zero(d))
    end
    Z = D * U'
```
Probably does not matter too much, but this does all of the logic in a single sweep without intermediate allocations.

src/algorithms/contractions/bondenv/gaugefix.jl
```diff
@@ -0,0 +1,114 @@
...
+            ↓                           ↓
+            1                           2
+
+                    =   1 - QR - 3  1 - R - 2
+                            ↓
+                            2
+    =#
+    QL, L = leftorth(Z, ((2, 1), (3,)))
+    QR, R = leftorth(Z, ((3, 1), (2,)))
+    Linv, Rinv = inv(L), inv(R)
```
lkdvos, line 62: I'm not sure how much this really matters, but in principle it would be more efficient to avoid computing the inverses and simply using `Z / R` and `Z / L`. Is this feasible?

If not, I also think it might be necessary at some point to do a regularized inverse instead, I'm not sure how small the singular values of `L` and `R` become.

src/utility/svd.jl
```diff
@@ -147,12 +147,20 @@ end
 ## Forward algorithms
 
 # Compute condition number smax / smin for diagonal singular value TensorMap
-function _condition_number(S::AbstractTensorMap)
+function _condition_number_sdiag(S::AbstractTensorMap)
```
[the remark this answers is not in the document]
Yue-Zhengyuan: The (now) `_condition_number_sdiag` computes the condition number using the already calculated SVD `s` matrix. But in the tests I need a function that does SVD and `cond` together. So I defined a new `_condition_number` function.

src/algorithms/contractions/bondenv/gaugefix.jl
Yue-Zhengyuan: Could you please elaborate on how `Z/R` (or `Z/L`) is different? In addition, `Z` has 3 indices, so I need to `permute` it accordingly when calculating `Z/R` and `Z/L`.

I'll do a test with Heisenberg model on the singular values of `R` and `L`.

src/utility/svd.jl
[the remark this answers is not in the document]
pbrehmer: > But in the tests I need a function that does SVD and cond together. So I defined a new _condition_number function.

Note that this is actually already done in [`LinearAlgebra.cond(::AbstractTensorMap)`](https://github.com/Jutho/TensorKit.jl/blob/0c73bd2ff03dee9b6eb597863d0dbdcefce4ab3b/src/tensors/linalg.jl#L299)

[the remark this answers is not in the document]
Yue-Zhengyuan: To use `LinearAlgebra.cond` the tests again need to restrict the version of TensorKit to 0.14.6 or higher. Now I already have a few tests that requires this. When do we plan to drop support for older TensorKit in `Project.toml`?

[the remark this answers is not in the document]
pbrehmer: Since we anyway already require `v0.14` we might as well restrict to `v0.14.6` (or higher). Especially if we want to start using `cond` that would make sense in my opinion.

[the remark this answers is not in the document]
lkdvos: I'm happy to restrict to v0.14.6+, especially if we can then remove some code here.

src/algorithms/contractions/bondenv/gaugefix.jl
lkdvos: It's mostly about efficiency, although it depends on the sizes. For example:
```julia
A = rand(1000, 1000)
X = rand(1000)

@btime $A \ $X;
# 3.236 ms (9 allocations: 7.67 MiB)

@btime inv($A) * $X;
# 11.100 ms (13 allocations: 8.17 MiB)
```

But honestly it might not be worth it to deal with that.
The problem being that `L` and `R` are already triangular, so in principle only a triangular solve is needed to solve the linear system, but I don't think we really have the infrastructure set up to make that actually work.

Yue-Zhengyuan:
Do you know the reason on why the bond environment condition number is quite large for a physical state, but small for a random state? And why can QR decomposition help?

src/algorithms/contractions/bondenv/gaugefix.jl
lkdvos: oh you are right, it should indeed be `d > 0 ? sqrt(d) : zero(d)`

src/utility/svd.jl
```diff
@@ -149,7 +149,7 @@ end
 # Compute condition number smax / smin for diagonal singular value TensorMap
 function _condition_number(S::AbstractTensorMap)
     smax = maximum(first ∘ last, blocks(S))
-    smin = maximum(last ∘ last, blocks(S))
+    smin = minimum(last ∘ last, blocks(S))
```
lkdvos, line 152: Can we completely remove this and replace with `LinearAlgebra.cond`?

pbrehmer: That would mean an extra cost of a small SVD, right? But that would be probably negligible anyways, so I guess we can completely replace `_condition_number` with `LinearAlgebra.cond`

Yue-Zhengyuan: > That would mean an extra cost of a small SVD

That's why I didn't remove `_condition_number`. What do you suggest @lkdvos ?

lkdvos:
I'm not sure I really have an explanation for the condition number of a physical state vs a random state. Speculation: It could be that for a random initial guess the gauge is just very bad, which affects the condition number more than the actual state it represents. Therefore it's not actually the state, but rather the gauging that makes this better, which is why the QR might help since that is a partial gauging procedure?

src/algorithms/contractions/bondenv/gaugefix.jl
Yue-Zhengyuan: I tested the largest condition number of R, L that will occur in the following cases. 

- Heisenberg model D = 4, χ = 16:  about 50. 
- Hubbard model (U/t = 8, half filling) with D = 4, χ = 16: about 300 (for one bond in the 2x2 unit cell; other bonds are only about 30). 
- t-J model (t/J = 3, 10% doping, d-wave) with D = 12, χ = 64: about 30. (this is surprisingly small for me considering the much larger bond dimension in this case)

I think these are still moderate values, and we can keep the current implementation. Anyway maybe pseudo-inverse can help?

BTW, I also want to make a change in the bond truncation algorithms. Previously I was using `linsolve`, but now I'm considering using pseudo-inverse (obtained from truncated SVD); this is used in [YASTN](https://github.com/yastn/yastn/blob/3d8a435d81efbd2cc09b67f2248c4220cc87dfdd/yastn/tn/fpeps/_evolution.py#L487-L505). @lkdvos What do you think about these two ways to solve linear equations?

src/utility/svd.jl
pbrehmer: I opened #208 to do this

src/algorithms/contractions/bondenv/gaugefix.jl
lkdvos: I definitely don’t have enough experience with this to have a feel about whether or not it would be better. It might be good to just keep it in mind? I don’t want to block this PR for this reason

Yue-Zhengyuan: I've added a TODO reminder here. In the real time evolution example, I find it useful to change `linsolve` to `pinv` when doing ALSTruncation, or use FET instead, but that's out of the scope of the current PR...

PR #206, opened 2025-05-28
Properly normalize truncation error in `compute_projector`

### LINKED ISSUE #204: Possible useless `@set` in `compute_projector`
https://github.com/QuantumKitHub/PEPSKit.jl/blob/cd7ee2c7e9708f2ba1540f5bc5320daae91c3d72/src/algorithms/ctmrg/projectors.jl#L173

https://github.com/QuantumKitHub/PEPSKit.jl/blob/cd7ee2c7e9708f2ba1540f5bc5320daae91c3d72/src/algorithms/ctmrg/projectors.jl#L194

In these places, the truncation error of computing the projectors is to be normalized. However, using `@set` will not change `info` but return a modified copy of `info` instead. @pbrehmer Did you intended to use `@reset` instead? I see that `@reset` is imported but not used anywhere.

pbrehmer:
This PR replaces `@set` by `@reset` to normalize the truncation error in `compute_projector`, as noted by @Yue-Zhengyuan. Closes #204.

### DIFF
--- a/src/algorithms/ctmrg/projectors.jl
+++ b/src/algorithms/ctmrg/projectors.jl
@@ -170,7 +170,7 @@ function compute_projector(enlarged_corners, coordinate, alg::HalfInfiniteProjec
         end
     end
 
-    @set info.truncation_error = info.truncation_error / norm(S) # normalize truncation error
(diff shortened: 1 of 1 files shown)


lkdvos (approved):
ok if tests pass!

PR #208, opened 2025-05-30
Remove `_condition_number` and replace by `LinearAlgebra.cond`

pbrehmer:
As [discussed](https://github.com/QuantumKitHub/PEPSKit.jl/pull/203#discussion_r2109629532) in PR #203.


Yue-Zhengyuan:
`cond` is added in TensorKit [0.14.5](https://github.com/Jutho/TensorKit.jl/releases/tag/v0.14.5). If PR is to be merged before #203, I think we just change `Project.toml` here to support only TensorKit 0.14.6 or higher.

lkdvos (approved):

Issue #163, opened 2025-03-20
Memory usage when calculating expectation value on two far away sites on the same row/column


Yue-Zhengyuan:
When calculating the expectation value `<A(i) B(j)>` on two sites `i = (0, 0)`, `j = (r, 0)` or `(0, r)` using `expectation_value`, I notice that the memory usage increases as `r` increases, which can reach over 100GB when D = 12, χ = 64 and r = 7. But when r = 1 or 2 it only uses ~30GB of memory. 

The memory usage should be more or less the same for any `r` when the two sites are on the same row (or same column) of the iPEPS. I think either the internal logic of `expectation_value` can be improved, or some cache clearing should be added.

Yue-Zhengyuan:
I realized where the problem is...

What I was trying to measure is `<c+_up(i) c_up(j)>` in the t-J model. I cannot construct each `c_up` separately since it has odd parity, so I must create `c+_up(i) c_up(j)` first as a whole. This make the contraction highly inefficient. The memory will not increase with `r` only when `A`, `B` can be construct separately as a one-site operator. 

In general, the issue will arise whenever `A`, `B` by themselves are not symmetric / have even fermion parity, but the tensor product `A ⊗ B` is symmetric / has even fermion parity. We may still need to think of a way to calculate such correlation functions efficiently.

lkdvos:
It's definitely possible that this makes the problem slightly worse, but that does not really explain why it scales with `r`: in principle this is simply a constant factor `d^2`.

I'm assuming this is not in the context of AD right? Because there we would expect that you need to store all temporary objects, so that does increase with `r`.

If you have some MWE I can definitely try and look into it a bit more, but if I had to guess I would say that this might just be one of those cases where the Julia GC is not aggressive enough.
Large tensor contractions allocate quite a few temporary objects that are then immediately discarded, and sometimes Julia just does not trigger the GC in time. There is a `--heap-size-hint` CLI flag you can pass it to be more aggressive and see if that helps anything, and if so we can also experiment with some allocation strategies that are supported by TensorOperations.

Yue-Zhengyuan:
Thanks for the suggestion! No AD is currently involved; just plain measurement. I tried adding `--heap-size-hint=60GB` when running my code, and found that it doesn't work. 

Below is an MWE.

```julia
using TensorKit
using MPSKitModels
using PEPSKit
using PEPSKit: LocalOperator
using Random
using BenchmarkTools

[...]

PR #210, opened 2025-06-04
add correlation functions

sanderdemeyer:
This PR aims to add correlation functions for an `InfinitePEPS`. The aim is to extend this to `InfinitePartitionFunction`s and `InfinitePEPO`s. This extension would use 'excited' partition functions, the use of which is discussed in #117.

The preferred user interface and implementation are open for discussion. Once we have reached a consensus, I'll add the vertical correlators.


lkdvos:
This was precisely the original discussion in one of the earlier PRs. I would prefer if we default to giving the two-body operator and split it up ourselves, ie always work with an auxiliary leg, because I don't like code that only works for non-symmetric cases

src/algorithms/correlators.jl
```diff
@@ -0,0 +1,154 @@
+function correlator_horizontal(
+    ket::InfinitePEPS,
+    bra::InfinitePEPS,
+    env::CTMRGEnv,
+    O₁::AbstractTensorMap{T,S,1,2},
+    O₂::AbstractTensorMap{T,S,2,1},
+    inds::Tuple{CartesianIndex{2},CartesianIndex{2}};
```
Yue-Zhengyuan: Since the two sites are in the same row, how about using `(r, c), d` to indicate that the two sites are  `(r, c)` and `(r, c + d)`? Then we have a slightly less redundant input.

lkdvos:
Looking at this a bit more in depth: can we just mimick the MPSKit functions from [this file](https://github.com/QuantumKitHub/MPSKit.jl/blob/master/src/algorithms/correlators.jl)?

For example, I like that in MPSKit there is a distinction between specifying two `Int`s, which gives you a single value, and one `Int` and a range, which gives you all values you requested.
A straightforward generalization would be `i::CartesianIndex`, `j::CartesianIndex` giving a single value (basically an expectation value), while `i::CartesianIndex`, `j::Range{CartesianIndex}` would give you all values inbetween.
Note that IIRC this works: `CartesianIndex(1, 1):CartesianIndex(0, 1):CartesianIndex(21, 1)`.

For having the option to have different `bra` and `ket`, I'm happy to generalize the MPSKit function too: what do you think about `MPSKit.correlator((bra, ket), ...)` for that? Alternatively there are many parts of MPSKit that have `bra, operator, ket, envs` as argument orders, so I would also be fine with `bra, O12, i, j, ket, envs`.

For the implementation I have more or less the same remarks, is there an option to define this in a similar way as MPSKit, using transfer matrices? I am guessing there might be some missing `transfer_left` and `transfer_right` functions to pass the string, but otherwise this should be more or less the same. The main advantage being that this avoids having to define contractions in too many places, and more easily generalizes to things with other numbers of legs.
In an ideal world, we would define the MPSKit `correlator` in terms of generic functions that can be overloaded here, but I am fine not trying to merge these two for now.

lkdvos:
Just making sure we understand each other: the MPSKit implementation does not actually build the transfer matrix, it makes an object that holds the tensors that represent it, and the transfer functions only implement the actions of that transfer operator. In other words, it should more or less be the same implementation you have, just organized differently

src/algorithms/correlators.jl
sanderdemeyer: I've added both now. The first index `i` has to be a `CartesianIndex`, whereas the index `j` can both be a `CartesianIndex` or a range thereof (in line with MPSKit). If `j` is an `Int`, it automatically converts this to a range of `CartesianIndex`.

sanderdemeyer:
I now changed the code to be more closely related to how MPSKit does it, by defining transfer matrices and the function `transfer_left`. For now we don't need a `transfer_right`, so the implementation is somewhat simpler. I kept the ability to use operators without auxiliary leg (in a tuple), since it does not require an extra function now, just a slightly different implementation of `transfer_left`. 

After thinking about it a little bit more, the order `bra, O12, i, j, ket, envs` makes a lot of sense, so I changed the order to that one. It now accepts only a `CartesianIndex` for `i`, and a `CartesianIndices`, a `CartesianIndex` (in line with MPSKit, returning a single element), and an `Int` specifying the distance, like @Yue-Zhengyuan suggested.

Any additional comments or suggestions are very much appreciated.

src/algorithms/correlators.jl
```diff
@@ -0,0 +1,140 @@
...
+                )
+            final_norm = end_right(
+                left_side_norm_final,
+                env.corners[2, _prev(r, Nr), _next(c, Nc)],
+                env.edges[2, mod1(r, Nr), _next(c, Nc)],
+                env.corners[3, _next(r, Nr), _next(c, Nc)],
+            )
+            push!(corr, final / final_norm)
+        end
+        if c ≠ cs[end]
```
lkdvos: I think here you are doing every "transfer contraction" except for the final one twice, `left_side_final` is already the same as the updated `left_side` you are computing here. It's probably better to revert the logic: first you transfer `left_side` and `left_side_norm`, then you add the right caps if necessary.

src/algorithms/transfermatrix.jl
```diff
@@ -0,0 +1,101 @@
+abstract type AbstractTransferMatrix end;
+
+struct HorizontalTransferMatrix{T,S} <: AbstractTransferMatrix
+    above::TensorMap{T,S}
```
lkdvos: Note that this would be an abstract type, and therefore lead to type instabilities, but as mentioned before, using `MPSKit.TransferMatrix` should resolve this issue.

src/algorithms/correlators.jl
```diff
@@ -0,0 +1,140 @@
+function correlator_horizontal(
+    bra::InfinitePEPS,
+    O::Tuple{AbstractTensorMap{T,S},AbstractTensorMap{T,S}},
+    i::CartesianIndex{2},
+    j::CartesianIndices{2},
```
lkdvos: It seems a bit strange to have to provide both "horizontal" and "vertical" in the function names, and then also require general coordinates.
Can we not simply re-export `MPSKit.correlator(bra, O, i, j, ket, env)`, do the checks there and then dispatch to private/unexported implementations that might be called `correlator_horizontal`?

lkdvos (commented):
You don't actually have to define your own transfer operators, the ones from MPSKit are generic enough so you can simply overload them. This is precisely how the boundary MPS code works, so there already are some transfer functions defined: see [here](https://github.com/QuantumKitHub/PEPSKit.jl/blob/master/src/algorithms/contractions/vumps_contractions.jl).
I would therefore recommend to simply reuse `MPSKit.TransferMatrix` and go from there, this should reduce the number of transfer operations you have to define (and also reduce number of things to compile).

src/algorithms/correlators.jl
sanderdemeyer: I slightly modified the order now. I can't do it exactly like in MPSKit, since the last transfer function I call on `left_side` requires the operator as well, in contrast to `left_side_norm`. This change should eliminate the double evaluation. On the downside, the `if` statement now appears twice and there will be one useless `transfer_left` call on `left_side_norm` at the end.

src/algorithms/transfermatrix.jl
sanderdemeyer: Like you suggested above, I now use MPSKit.TransferMatrix instead of the one I defined earlier. I couldn't use any functions from vumps_contractions.jl, since there the lower environment is conjugated and twisted, which shouldn't happen here. Conjugating and twisting the environment tensors manually to compensate for that seems unnecessarily confusing in my opinion.

src/algorithms/correlators.jl
sanderdemeyer: Yes, that's a good idea. The only downside is that we can't specify a distance then, since we wouldn't know in which direction we want to evaluate the correlator, but that is fine with me. I also added a version where you only specify one `state`, which will then be both the ket and bra, to mimic the correlation function in MPSKit `correlator(state, O12, i, j)` a bit more.

lkdvos:
You can check the correlation length function for how this was calculated before, see https://github.com/QuantumKitHub/PEPSKit.jl/blob/c2a32f2f980cbc483dc63cdd66e2ba4a28cb5ec3/src/algorithms/toolbox.jl#L259

src/algorithms/correlators.jl
```diff
@@ -0,0 +1,140 @@
...
+    ket::InfinitePEPS,
+    env::CTMRGEnv,
+) where {T,S}
+    @assert size(ket) == size(bra) "The ket and bra must have the same unit cell."
+    (r, c₁) = Tuple(i)
+    cs = sort([ind[2] for ind in j]; dims=2)
```
lkdvos: I think it's better to `@assert issorted` here, it would be really strange to ask for the correlator for `js = 5:-1:1` and then produce an output that is silently sorted.

src/algorithms/correlators.jl
```diff
@@ -0,0 +1,140 @@
...
+function correlator_vertical(
+    bra::InfinitePEPS,
+    O,
+    i::CartesianIndex{2},
+    j::CartesianIndices{2},
+    ket::InfinitePEPS,
+    env::CTMRGEnv,
+)
+    i_rot = CartesianIndex(i[2], i[1])
+    j_rot = CartesianIndex(j[1][2], j[1][1]):CartesianIndex(j[end][2], j[end][1])
```
lkdvos: This looks like it will not produce correct results if the initial range did not have step size one I fear

lkdvos (commented):
Do you mind if I add some changes directly to this PR?

sanderdemeyer:
Sure, go ahead

lkdvos:
I updated the implementation to more closely resemble MPSKit's version, and tried to be a bit more generic with the input types. I also added some tests to ensure that the rotation is correctly handled, and that the range does not need to be contiguous (this is important when dealing with mixed physical spaces, where an operator might not even fit on each site).

I repurposed the `MPSKit.FiniteMPO` to convert the operator to things that can be handled by the contractions, I think it is fair to accept anything that can be converted to that.

Let me know what you think about this.

Yue-Zhengyuan:
I'll try to use these functions for an iPEPS and CTMRGEnv with large bond dimensions and see if the memory usage is improved over the generic `expectation_value` (#163).

pbrehmer (approved):
Looks great, thanks for the effort :-) 

It would be nice to show these new correlator capabilities in one of the examples or perhaps write up a new example specifically on correlators. But let's do that in a separate PR. Maybe I'll just convert the new correlator tests into a commented example.

Issue #207, opened 2025-05-29
[ENH] CTMRG arbitrary unit cell


ogauthe:
Currently, PEPSKit CTMRG assumes each site in the unit cell is different. For a 2x2 unit cell, it imposes a `AB//CD` pattern and computes 4 different environments. However, unit cell with repeated sites such as `AB//BA` are very common.

It is possible to hard code a bipartite IPEPS, however one can be slightly more ambitious and allow for any self-consistent unit cell pattern, such as `ABC//BCA//CAB`. The implementation can be made pretty simple. One needs to store unique site tensors in a vector and define `getitem` function to go back and forth between the unit cell coordinates and the internal 1D storage. The same accessors are used for the unique environment tensors.

As the CTMRG complexity is linear in the number of environments, making use of repetitions in the unit cell is a low hanging fruit. In my experience, it also improves stability and convergence.

pbrehmer:
Thanks for for the suggestion, I'd really like to get that in in the near future! I was thinking a bit about how to implement this, given that the state and operator types like `InfinitePEPS`, `InfinitePartitionFunction`, etc. wrap around a `Matrix`.

- We could indeed instead wrap a `Vector` and define custom `getitem` functions depending on the underlying lattice. This might, however, result in a relatively big rewrite of the state and operator types and their user interfaces.
- Another option might be to keep the `Matrix` as the fundamental unit cell type but instead make certain elements of the matrix point to the same tensor. For example, an `A B // B A` sublattice would satisfy `peps[1, 2] === peps[2, 1]` and `peps[1, 1] === peps[2, 2]`. To take advantage of that in the actual CTMRG algorithms we could then iterate only over `unique(peps.A)` and `unique(env.corners)` and so on.

These are only loose thoughts but I do think we should be careful which route we choose because these are changes that might propagate quite a lot throughout the code. @lkdvos Any thoughts on this?

PR #211, opened 2025-06-06
Allow bond-dependent truncation scheme in Simple Update

sanderdemeyer:
This is a small PR that adds an anisotropic version of Simple Update, i.e. where we allow for different bond dimensions on the vertical and horizontal bonds. This amounts to changing the parameter in `_su_bondx!`. I also added a constructor for when both truncation schemes should be the same. Maybe this is also a good moment to add a default version of Simple Update, which was still a TODO.


lkdvos:
Definitely cool! I had a vague idea somewhere in my mind to attempt this with a new truncationscheme that is fully site- or direction dependent, since this is also something that could be useful for ctmrg. There, we already have some preprocessing of the trscheme too, to allow for fixed space truncation, we might be able to leverage a similar system here?

sanderdemeyer:
Currently, we define the Simple Update struct without any reference to (the unit cell of) the PEPS. This would change if we make this truncation scheme to have the same dimensions as the PEPS, which would induce breaking changes. Maybe we can avoid this, but I haven't found a solution yet. I would be okay with introducing breaking changes, but what is your opinion on this?

To give some more context: I am thinking about stripe order in the Hubbard model. In this case, it could make sense to have a larger bond dimension along the direction of the charge modulation, and a lower one along the uniform direction.

lkdvos:
This can definitely be done in a non-breaking way, looking at CTMRG for inspiration. The clue is to define access functions that give the truncation strategy at a given site, which by default return the strategy just like before, but can be overloaded to return a specific one.

For example, you can see here that before each projection step we determine the truncation strategy based on some additional information that might be relevant: https://github.com/QuantumKitHub/PEPSKit.jl/blob/16956d5e52b42f6eab0caa73b9174656a398ea03/src/algorithms/ctmrg/sequential.jl#L85

As a side note, reiterating what @Yue-Zhengyuan said: while it is indeed reasonable to specify different **dimensions** for different sites/directions, a more fair truncation strategy is based on the truncation **cutoff/error**, which would automatically give you different dimensions.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -94,7 +94,13 @@ function su_iter(
                     direction == 1 ? gate : gate_mirrored,
                     (CartesianIndex(r, 1), CartesianIndex(r, 2)),
                 )
-                ϵ = _su_bondx!(r, 1, term, peps2, alg)
+                ϵ = _su_bondx!(
+                    r,
+                    1,
+                    term,
+                    peps2,
+                    truncation_scheme(alg.trscheme; direction, r, c, mirror_antidiag=true),
```
lkdvos: I'm slightly confused that both these truncation schemes have `mirror_antidiag=true`, I would think it might be a little cleaner to not have this as a keyword argument, and simply call `truncation_scheme(mirror_antidiag(alg.trscheme), ...)` in this case.

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -25,3 +41,22 @@ function _TruncationScheme(; alg=Defaults.trscheme, η=nothing)
...
+
+function truncation_scheme(trscheme::T; kwargs...) where {T<:TensorKit.TruncationScheme}
+    return trscheme
+end
+
+function truncation_scheme(
+    trscheme::SiteDependentTruncation;
+    direction::Int,
+    r::Int,
+    c::Int,
```
lkdvos: I think we have used coordinates as positional arguments throughout the rest of the code, so I am leaning slightly more towards that, especially since these aren't optional arguments, and this avoids having to remember that it is `r` instead of `row` and `c` instead of `col` or `column` etc.

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -7,6 +7,22 @@ have different spaces, this truncation style is different from `TruncationSpace`
 """
 struct FixedSpaceTruncation <: TensorKit.TruncationScheme end
 
+struct SiteDependentTruncation <: TensorKit.TruncationScheme
+    trschemes::Array{T,3} where {T<:TensorKit.TruncationScheme}
+end
+
+function SiteDependentTruncation(trscheme::TensorKit.TruncationScheme, Nr::Int, Nc::Int)
+    return SiteDependentTruncation(reshape(fill(trscheme, Nr, Nc), 2, Nr, Nc))
```
lkdvos: This reshape looks a bit off, you reshape a `Nr x Nc` array into a `2 x Nr x Nc` array?
Additionally, it might make more sense to have a syntax similar to the `InfinitePEPS` constructor, which has a keyword argument `unitcell` IIRC.

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -7,6 +7,22 @@ have different spaces, this truncation style is different from `TruncationSpace`
...
+end
+
+function SiteDependentTruncation(
+    trschemes::Tuple{T,S}, Nr::Int, Nc::Int
+) where {T<:TensorKit.TruncationScheme,S<:TensorKit.TruncationScheme}
```
lkdvos: I'm not sure how intuitive this constructor is, I understand that your specific case is `(xtrscheme, ytrscheme)::Tuple`, but this is not really something I would immediately guess. In that sense, there either would have to be a docstring to explain this, or we simply leave that to the user, `trschemes = stack((fill(xschemes, Nr, Nc), fill(yschemes, Nr, Nc)); dims=1)` may not be too inconvenient?

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -34,7 +34,7 @@ function _su_bondx!(
     col::Int,
     gate::AbstractTensorMap{T,S,2,2},
     peps::InfiniteWeightPEPS,
-    alg::SimpleUpdate,
+    trscheme::TensorKit.TruncationScheme,
```
lkdvos: I think you still want `alg::SimpleUpdate` here and extract the truncation scheme from that

lkdvos (commented):
I like the direction of this a lot, looks very versatile without having to alter a lot of code. I left a bunch of minor comments and questions that might be improvements or bring things more in line with the rest of the library in the meantime.

(Also, it might at this point make sense to add a `using TensorKit: TruncationScheme` somewhere instead of having to qualify that everywhere).

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -418,7 +422,13 @@ function su3site_iter(
         for site in CartesianIndices(peps2.vertices)
             r, c = site[1], site[2]
             gs = gatempos[i][r, c]
-            _su3site_se!(r, c, gs, peps2, alg)
+            _su3site_se!(
+                r,
+                c,
+                gs,
+                peps2,
+                truncation_scheme(alg.trscheme, direction, r, c; mirror_antidiag=true),
```
Yue-Zhengyuan: This is not correct. Here you need to extract the truncation schemes of two bonds `[r, c] -- [r, c+1]` and `[r, c+1] -- [r-1, c+1]` instead of using `mirror_antidiag`.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -384,7 +388,7 @@ function _su3site_se!(
     coords = ((row, col), (row, cp1), (rm1, cp1))
     # weights in the cluster
     wt_idxs = ((1, row, col), (2, row, cp1))
-    wts, ϵ = apply_gatempo!(Ms, gs; trunc=alg.trscheme)
+    wts, ϵ = apply_gatempo!(Ms, gs; trunc=trscheme)
```
Yue-Zhengyuan: Just want to remind you that the truncation used to convert the 3-site gate to an MPO is different from the PEPS truncation below.

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -7,6 +7,22 @@ have different spaces, this truncation style is different from `TruncationSpace`
 """
 struct FixedSpaceTruncation <: TensorKit.TruncationScheme end
 
+struct SiteDependentTruncation <: TensorKit.TruncationScheme
+    trschemes::Array{T,3} where {T<:TensorKit.TruncationScheme}
+end
+
+function SiteDependentTruncation(trscheme::TensorKit.TruncationScheme, Nr::Int, Nc::Int)
```
Yue-Zhengyuan: If we ever want to use `SiteDependentTruncation` for CTMRG, it is best to keep the first dimension not fixed to 2, but specified by users.

src/algorithms/time_evolution/simpleupdate3site.jl
sanderdemeyer: Thanks for noticing! I've taken a deeper look into the 3-site SU code, and muy solution would be to change the `trunc` argument in the function `apply_gatempo!` to be a vector of size `length(Ms)-1`, where we extract what it should be somewhere in the function `_su3site_se!`. The alternative would be to add `row` and `col` as arguments through the chain of functions, but this seems suboptimal to me. Please let me know if you would prefer a different solution.

src/algorithms/time_evolution/simpleupdate.jl
sanderdemeyer: This would indeed be a nicer solution. However, for now I've kept it as a keyword argument, since there is no function to call `mirror_antidiag` on a `TensorKit.TruncationScheme`. I could add this function, though, applying `mirror_antidiag` when it is a `SiteDependentTruncation` and doing nothing when it is a different one.

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: That's indeed a good idea. I've changed all 3 keyword argument to positional ones.

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: That's a good idea. I changed it in both constructors.

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: As a side remark, I changed it to be also compatible with multiple directions (So a change to an `NTuple`). 

I do feel this constructor is something that is going to be used a lot, since having a different truncation scheme for the different directions, but keeping translational invariance, it is the most natural extension to the completely uniform case. I will add a docstring to explain this thoroughly, though.

src/algorithms/time_evolution/simpleupdate.jl
sanderdemeyer: I was indeed a bit doubtful about this. The reason I changed it was because the function `_su_bondx!` is agnostic about `direction` and thus doesn't know whether it should apply `mirror_antidiag`. 
It also only requires the truncation scheme (and nothing else) from the `SUAlgorithm`. This might change later if we would change the SU algorithm, but I can't think of some extensions that would require this.

So TL;DR, I would be inclined to keep it like this. If you prefer another solution, like inputting a new `SimpleUpdate` struct where `mirror_antidiag` was applied to `trscheme`, please let me know.

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: That's a good idea. Although the struct itself can stay the same, I changed the constructors that assumed this. This would in principle then also work for non-square lattices, if we would ever add those.

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: > change the `trunc` argument in the function `apply_gatempo!` to be a vector of size `length(Ms)-1`

I agree. This amounts to using different truncation schemes on each bond of an MPS. You also need to change several upstream functions that `apply_gatempo!` depends on.

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: @lkdvos Sander's approach is fine. In the old version only `trscheme` of the `SimpleUpdate` struct is used. Now `trscheme` will be the truncation scheme of the single bond that `_su_xbond!` acts on (@sanderdemeyer I have renamed `bondx` to `xbond` in #219). We don't need to pass to it the entire array of truncation schemes for all bonds.

src/algorithms/time_evolution/simpleupdate.jl
sanderdemeyer: Update: I removed the keyword argument in favor of implementing a new `mirror_antidiag`

src/algorithms/truncation/truncationschemes.jl
lkdvos: I think I actually strongly disagree with you here.
Having a different truncation scheme for the different directions to me feels incredibly arbitrary.
Checkerboard patterns seem to me equally likely to come up, or stripes, or ...?
I really prefer clarity here, and looking at the function `SiteDependent...`, it doesn't make too much sense to default to a direction dependent truncation.

Additionally, this looks like it is trying to conflate the constructor too much as well.
Note how for arrays there are multiple functions to do multiple things, there is `fill`, `repeat`, ..., and the constructor is actually only used to construct the object.
It's also just extra maintenance from our side, since we do have to test these extra functions as well.

Therefore I think I still vote against having this constructor.
If you want to have a site dependent truncation, I think it is better to be obvious about it, rather than make too many arbitrary choices.
If nothing else, it should probably be a separate function, rather than a type constructor.
Especially given that it is a one-liner to actually have a direction dependent array, it feels a bit unnecessary?

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -25,3 +47,38 @@ function _TruncationScheme(; alg=Defaults.trscheme, η=nothing)
...
+    trscheme::T, direction::Int, row::Int, col::Int; kwargs...
+) where {T<:TensorKit.TruncationScheme}
+    return trscheme
+end
+
+function truncation_scheme(
+    trscheme::SiteDependentTruncation, direction::Int, row::Int, col::Int;
+)
+    return trscheme.trschemes[direction, row, col]
+end
```
Yue-Zhengyuan: I don't like adding the function `truncation_scheme`. 

- When `trscheme` is a single `TruncationScheme`, it feels too redundant that I still have to wrap it in `truncation_scheme`. And in this case the function arguments become useless.  
- For `SiteDependentTruncation`, you can define `Base.getindex` to access its elements. It allows you to directly write `trscheme[direction, row, col]` elsewhere. 

To get rid of this function, we can define a constructor of `SimpleUpdate`, such that when the input `trscheme` is a single TruncationScheme as before, it automatically converts it to a SiteDependentTruncation. Then in SU, we only need to deal with the SiteDependentTruncation array. 

I know this will introduce some overhead when `SiteDependentTruncation` is actually site-independent, but I think this does not hurt the performance too much.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -418,7 +422,11 @@ function su3site_iter(
         for site in CartesianIndices(peps2.vertices)
             r, c = site[1], site[2]
             gs = gatempos[i][r, c]
-            _su3site_se!(r, c, gs, peps2, alg)
+            trschemes = [
+                truncation_scheme(alg.trscheme, 1, r, c)
+                truncation_scheme(alg.trscheme, 2, r, _next(c, Nc))
+            ]
```
Yue-Zhengyuan: With `Base.getindex` defined for `SiteDependentTruncation`, I can simply write 
```julia
trschemes = [
    alg.trscheme[1, r, c], alg.trscheme[2, r, _next(c, Nc)]
]
```

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -25,3 +47,38 @@ function _TruncationScheme(; alg=Defaults.trscheme, η=nothing)
...
+
+# Mirror a TruncationScheme by its anti-diagonal line.
+# When the number of directions is 2, it swaps the first and second direction, consistent with xbonds and ybonds, respectively.
+# When the number of directions is 4, it swaps the first and second, and third and fourth directions, consistent with the order NORTH, EAST, SOUTH, WEST.
+function mirror_antidiag(trscheme::T) where {T<:TensorKit.TruncationScheme}
+    return trscheme
+end
```
Yue-Zhengyuan: With the constructor of `SimpleUpdate` I suggested, you can avoid dealing with this trivial case.

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: Would you be okay with renaming it as `DirectionDependentTruncation`, instead of keeping it a type constructor? Then it would seem obvious enough what we are doing.

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: I tried making an inner constructor for this, but I ran into the problem that this requires the constructor of SimpleUpdate to know the size of the unit cell, which would introduce breaking changes. 

I think your proposal would be a very similar approach to when we changed the SimpleUpdate struct to contain a 3D array of TruncationSchemes. The current implementation does feel slightly redundant, but maybe as redundant as the overhead introduced by converting a site-independent truncation scheme to a site-dependent one? If you have a way around this, please let me know.

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: If we do that, we maybe should also change the name of `SiteDependentTruncation` to something else for more clarity, since it can also depend on the direction and not only on the site. Maybe something like `AnisotropicTruncation` or `SpaceDependentTruncation`?

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: If we keep the current strategy, I could still define Base.getindex to remove the function `truncation_scheme`, both for SiteDependentTruncation as the 'normal' truncation schemes by defining
`Base.getindex(trscheme::TruncationScheme, args...) = trscheme`
admittedly, this is again as redundant as before, but would look indeed a bit nicer.

Yue-Zhengyuan: > I ran into the problem that this requires the constructor of SimpleUpdate to know the size of the unit cell

Oh I overlooked this... then for now I don't have a better solution.

src/algorithms/truncation/truncationschemes.jl
Yue-Zhengyuan: > If we do that, we maybe should also change the name of `SiteDependentTruncation` to something else for more clarity, since it can also depend on the direction and not only on the site. Maybe something like `AnisotropicTruncation` or `SpaceDependentTruncation`?

This is going a bit too involved, and makes code maintenance hard. A single struct that allows different schemes on each bond should be enough. I agree with Lukas that how to fill the array should be left to users. 

Honestly, I don't feel comfortable at all with being too flexible with truncation schemes; a single truncerr/truncbelow should work for most cases...

sanderdemeyer: Okay, then I'll remove the constructor. 

I still think that we should change the name to emphasise that it can also depend on the direction. If that's okay for you and if you don't have a better alternative, I'll change `SiteDependentTruncation` to `SpaceDependentTruncation`.

Yue-Zhengyuan: For time evolution, I would use `BondDependentTruncation`. But I don't think it describes the CTMRG case very well.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -418,7 +422,11 @@ function su3site_iter(
...
+            _su3site_se!(r, c, gs, peps2, trschemes)
         end
         peps2 = rotl90(peps2)
```
Yue-Zhengyuan, line 431: In 3-site SU I use `rotl90` instead of `mirror_antidiag`. Here I think you need to rotate `trschemes` as well, so you still need to define rotations for it.

src/algorithms/truncation/truncationschemes.jl
lkdvos: `VariableTruncation`, `AdaptiveTruncation`, `TruncationGrid`, ...

I think `SpaceDependentTruncation` is a bit misleading again since it does not take into account the spaces of the tensors, but rather the geographical space. I like `AnisotropicTruncation`, but that technically only includes directional variability, so maybe `VariableTruncation` or `TruncationGrid` is the better option? Honestly I think `SiteDependentTruncation` was pretty good, it's just that I don't think the constructor you proposed should be there.

src/algorithms/truncation/truncationschemes.jl
lkdvos: You cannot define `Base.getindex(::TruncationScheme, ...)` since that is type piracy, we don't own `Base.getindex` nor `TruncationScheme` so that isn't great. I think the added function to use dispatch on is really the correct Julian way to do this, so I'd stick with that.

src/algorithms/truncation/truncationschemes.jl
Yue-Zhengyuan: `VariableTruncation` is not as descriptive as `SiteDependentTruncation` (I can't tell what is variable here). I suggest we keep `SiteDependentTruncation`.

sanderdemeyer: Also fine with me.

sanderdemeyer:
Currently, the patch coverage test fails because the case where the number of directions is 4 (which will be used for CTMRG) is not covered. I can remove this is it is preferred that we only add this when we would actually use it.

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -7,6 +7,17 @@ have different spaces, this truncation style is different from `TruncationSpace`
...
+struct SiteDependentTruncation <: TensorKit.TruncationScheme
+    trschemes::Array{T,3} where {T<:TensorKit.TruncationScheme}
+end
+
+function SiteDependentTruncation(
+    trscheme::TensorKit.TruncationScheme, directions::Int; unitcell::Tuple{Int,Int}=(1, 1)
+)
+    Nr, Nc = unitcell
+    return SiteDependentTruncation(fill(trscheme, directions, Nr, Nc))
+end
```
lkdvos: After looking at this constructor a bit more, I think this might not make too much sense: this would construct a sitedependent location which is constant over all sites and directions.

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -7,6 +7,17 @@ have different spaces, this truncation style is different from `TruncationSpace`
 """
 struct FixedSpaceTruncation <: TensorKit.TruncationScheme end
 
+struct SiteDependentTruncation <: TensorKit.TruncationScheme
```
lkdvos: Can you add `using TensorKit: TruncationScheme` and shorten this throughout the library?

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -25,3 +36,59 @@ function _TruncationScheme(; alg=Defaults.trscheme, η=nothing)
...
+        trschemes_mirrored[4, :, :] = mirror_antidiag(trscheme.trschemes[3, :, :])
+    else
+        error("Unsupported number of directions for mirror_antidiag: $directions")
+    end
+    return SiteDependentTruncation(trschemes_mirrored)
+end
+
+function Base.rotl90(trscheme::T) where {T<:TensorKit.TruncationScheme}
+    return trscheme
+end
```
lkdvos: This is a bit unfortunate in the sense that it is type piracy again. I'm somewhat okay to just leave it and deal with this later, but can you leave a `# TODO: type piracy` in here?

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -25,3 +36,59 @@ function _TruncationScheme(; alg=Defaults.trscheme, η=nothing)
...
+
+function Base.rotl90(trscheme::T) where {T<:SiteDependentTruncation}
+    directions = size(trscheme.trschemes)[1]
+    trschemes_rotated = permutedims(trscheme.trschemes, (1, 3, 2))
+    if directions == 2
+        trschemes_rotated[1, :, :] = circshift(rotl90(trscheme.trschemes[2, :, :]), (0, -1))
```
lkdvos: Can you rewrite these directions with the constants `NORTH, EAST, SOUTH, WEST`? I think that's easier to read and has the added advantage that it is a little more future-proof

test/examples/heisenberg.jl
```diff
@@ -71,7 +71,30 @@ end
...
+                    ) & truncdim(
+                        Dbond2
+                    ) truncerr(tol) & truncdim(Dbond2 + 1) truncerr(tol) & truncdim(Dbond2) truncerr(
+                        tol
+                    ) & truncdim(
+                        Dbond2
+                    ) truncerr(tol) & truncdim(Dbond2) truncerr(tol) & truncdim(Dbond2)],
+                    2,
+                    2,
+                    2,
```
lkdvos: Could you split this a bit so the formatter doesn't make this into this? It's a bit hard to follow what is going on here

src/algorithms/truncation/truncationschemes.jl
```diff
@@ -25,3 +36,59 @@ function _TruncationScheme(; alg=Defaults.trscheme, η=nothing)
...
+
+# Mirror a TruncationScheme by its anti-diagonal line.
+# When the number of directions is 2, it swaps the first and second direction, consistent with xbonds and ybonds, respectively.
+# When the number of directions is 4, it swaps the first and second, and third and fourth directions, consistent with the order NORTH, EAST, SOUTH, WEST.
+function mirror_antidiag(trscheme::T) where {T<:TensorKit.TruncationScheme}
+    return trscheme
+end
+function mirror_antidiag(trscheme::T) where {T<:SiteDependentTruncation}
+    directions = size(trscheme.trschemes)[1]
+    trschemes_mirrored = permutedims(trscheme.trschemes, (1, 3, 2))
```
lkdvos: Would this not be a mirror around the diag instead of the antidiag?

src/algorithms/truncation/truncationschemes.jl
Yue-Zhengyuan: This reminds me that I defined `rotl90` etc for 2D `CartesianIndex`; I guess this is also type piracy?

Can we fix it by defining `_rotl90` etc (change the function name) for external structs?

lkdvos: I think the best fix would be to define our own `PEPSKit.rotl90(x) = Base.rotl90(x)`, and then overload that one. I did notice it for the `CartesianIndex` as well, it's not great in the long run but it's such a convenient thing to have...
This can be done in a follow-up PR though

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: I just checked it and it should be fine, I think? I tried it for a 2x2 unit cell and it was indeed mirrored around the diagonal through the lower left corner, as expected.

src/algorithms/truncation/truncationschemes.jl
sanderdemeyer: I removed it.

src/algorithms/truncation/truncationschemes.jl
lkdvos:
```julia
julia> A = [1; 1;; 2; 2;;; 3; 3;; 4; 4]
2×2×2 Array{Int64, 3}:
[:, :, 1] =
 1  2
 1  2

[:, :, 2] =
 3  4
 3  4

julia> A[1,:,:]
2×2 Matrix{Int64}:
 1  3
 2  4

julia> A[2,:,:]
2×2 Matrix{Int64}:
 1  3
 2  4

julia> B = permutedims(A, (1, 3, 2))
2×2×2 Array{Int64, 3}:
[:, :, 1] =
 1  3
 1  3

[:, :, 2] =
 2  4
 2  4

julia> B[1, :, :]
2×2 Matrix{Int64}:
 1  2
 3  4

julia> B[2, :, :]
2×2 Matrix{Int64}:
 1  2
 3  4
```
I dont think so, permuting is exchanging rows and columns, so that would be reflecting around the diagonal. The printing of arrays is a bit confusing here though since our directions are the first index.

There's a function for this in the utility file though, so you can just reuse that one:
https://github.com/QuantumKitHub/PEPSKit.jl/blob/47e66755411673a969f95bec4ab788de3a256b48/src/utility/mirror.jl#L9-L12

sanderdemeyer: That is why we have the lines below, which indeed use this `mirror_antidiag` function. As an example (with directions == 2)

```julia
julia> A = [1; 2;; 3; 4;;; 5; 6;; 7; 8];
julia> B = permutedims(A, (1, 3, 2));
julia> B[1, :, :] = PEPSKit.mirror_antidiag(A[2, :, :]);
julia> B[2, :, :] = PEPSKit.mirror_antidiag(A[1, :, :]);

julia> A[1,:,:]
2×2 Matrix{Int64}:
 1  5
 3  7

julia> A[2,:,:]
2×2 Matrix{Int64}:
 2  6
 4  8

julia> B[1,:,:]
2×2 Matrix{Int64}:
 8  6
 4  2

julia> B[2,:,:]
2×2 Matrix{Int64}:
 7  5
 3  1
```
which, if you draw it as bonds (i.e. the first index shows the unit cell of the x-bonds, the second one of the y-bonds) it is indeed mirrored around the antidiagonal.

lkdvos: Oh, I see what is going on. Did you just mean to preallocate an output object that has the correct sizes? I didn't realize you are not using anything of the `permutedims(A, (1, 3, 2))` entries, and you are fully overwriting it.

I think it might be slightly less confusing to use something like:
`stack((mirror_antidiag(A[2, :, :]), mirror_antidiag(A[1, :, :])); dims=1)`
or alternatively `B = similar(A, permute(size(A), (1, 3, 2))` and then writing into that.

Yue-Zhengyuan: Personally I prefer `stack`. I was once looking for such a function, but GPT kept telling me to use `cat` 😅 maybe because `stack` was only introduced in Julia 1.9.

lkdvos: `stack(mirror_antidiag.(reverse(eachslice(A; dims=1)))` is a fun way to write completely incomprehensible code :)

Some alternative which is more reasonable:
```julia
B = reverse!(permutedims(A, (1, 3, 2))) # only for 2 directions!
```
Makes use of the fact that `mirror_antidiag` is transpose + reverse, and the directions are swapped.

sanderdemeyer: I think the current code is a good balance between readability and compactness (and it would also work for more than 2 directions)

sanderdemeyer:
The amount of iterations in the current test of `SiteDependentTruncation` could indeed be much lower, so for me your proposal is fine. You can push directly to this branch to make your changes.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -185,6 +188,8 @@ function simpleupdate(
     nnonly = is_nearest_neighbour(ham)
     use_3site = force_3site || !nnonly
     @assert !(bipartite && use_3site) "3-site simple update is incompatible with bipartite lattice."
+    bipartite &&
+        @assert size(peps) == (2, 2) "Bipartite structure is only compatible with square unit cells."
```
Yue-Zhengyuan: I already have a check inside `su_iter`, so no need to check unit cell size here. I will remove it later.

test/examples/heisenberg.jl
Yue-Zhengyuan: I have avoided this in the latest test.

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: > For 2-site SU with `bipartite=true`, we may add a check on the input SiteDependentTruncation to ensure it is compatible with the bipartite structure.

I think we can postpone it until we decide to solve #207. Then we need to implement a more general check.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -235,16 +235,20 @@ end
 Given a cluster `Ms` and the pre-calculated `R`, `L` bond matrices,
 find all projectors `Pa`, `Pb` and Schmidt weights `wts` on internal bonds.
 """
-function _get_allprojs(Ms, Rs, Ls, trunc::TensorKit.TruncationScheme, revs::Vector{Bool})
+function _get_allprojs(
+    Ms, Rs, Ls, trschemes::Vector{E}, revs::Vector{Bool}
+) where {E<:TensorKit.TruncationScheme}
```
Yue-Zhengyuan: With `using TensorKit: TruncationScheme`, I think we can replace all `TensorKit.TruncationScheme` with `TruncationScheme` throughout the repo.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -235,16 +235,20 @@ end
...
     N = length(Ms)
+    @assert length(trschemes) == N - 1
     projs_errs = map(1:(N - 1)) do i
-        trunc2 = if isa(trunc, FixedSpaceTruncation)
+        trunc = trschemes[i]
```
Yue-Zhengyuan:
```suggestion
```
Line 244 is redundant and can be removed.

Yue-Zhengyuan (changes requested):
Apart from a few minor changes below, this PR is good to go for me.

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: @sanderdemeyer I think you missed one in `src/algorithms/time_evolution/simpleupdate.jl`. If you're using VS Code, you can use the side bar search (command + shift + F) to make replacements throughout the repo; then you'll not miss anything.

sanderdemeyer: You're right, thanks! I changed it now.

lkdvos (approved):
Made some final clean-up changes I noticed, approved once tests pass.

lkdvos (dismissed):

PR #212, opened 2025-06-09
Update correlation length normalization

### LINKED ISSUE #199: `correlation_length` may return wrong values
I unpacked the reason why the AKLT correlation length mentioned in #197 is wrong. It is not a missing factor 2, it is a bug.

The correlation length is obtained from the spectrum of the MPS transfer matrix defined by the edges. The leading eigenvalue is always a singlet and is normalized at 1. However, the second leading eigenvalue, corresponding to the leading excitation, may be in a different sector. This is indeed the case for AKLT. In this case, the leading singlet eigenvalue provides the normalization and the leading triplet eigenvalue defines the gap. The correlation length is defined as `xi = -Lx / log(abs(λ_triplet[1] / λ_singlet[1]))`. It can be obtained using `MPSKit` interface as
```julia
mps = InfiniteMPS([env_c4v.edges[1, 1, 1]]);
ξ = correlation_length(mps, sector=SU2Irrep(1))  # around 2.0 for AKLT
```

However `PEPSKit` cannot return this value: a given sector is fixed (default to singlet) and `xi = -Lx / log(abs(λ_sector[2] / λ_sector[1]))`. When `sector=SU2Irrep(0)` , `λ_sector[2]` is not the global leading excitation and the returned correlation length is not the largest one. When `sector=SU2Irrep(1)`, `λ_sector[1]` is the leading excitation (and not the normalizing factor), and `λ_sector[2]` is the second excitation.

As a check, one recovers the correct result when dropping SU(2) symmetry in the `TensorMap`.

lkdvos:
This should resolve #199 by correctly normalizing the transfer spectra.

@ogauthe, do you happen to have some explicit results that I can use to add tests?

### DIFF
--- a/src/algorithms/toolbox.jl
+++ b/src/algorithms/toolbox.jl
@@ -275,25 +275,45 @@ in a specific symmetry sector).
 MPSKit.correlation_length(state, env::CTMRGEnv; num_vals=2, kwargs...) =
     _correlation_length(env; num_vals, kwargs...)
 
-function _correlation_length(env::CTMRGEnv; num_vals=2, kwargs...)
+function _correlation_length(
+    env::CTMRGEnv; num_vals=2, sector=one(sectortype(env)), kwargs...
+)
     _, n_rows, n_cols = size(env)
 
     # Horizontal
     λ_h = map(1:n_rows) do r
         above = InfiniteMPS(env.edges[NORTH, r, :])
         below = InfiniteMPS(_dag.(env.edges[SOUTH, _next(r, n_rows), :]))
-        vals = MPSKit.transfer_spectrum(above; below, num_vals, kwargs...)
-        return vals ./ abs(vals[1]) # normalize largest eigenvalue
+        vals = MPSKit.transfer_spectrum(above; below, num_vals, sector, kwargs...)
+
+        # normalize using largest eigenvalue in trivial sector
+        if isone(sector)
+            N = abs(first(vals))
+        else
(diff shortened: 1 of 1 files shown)


ogauthe:
The correlation length for the AKLT spin 2 wavefunction is slightly above 2. It may be added to https://github.com/QuantumKitHub/PEPSKit.jl/blob/master/test/ctmrg/correlation_length.jl
Getting a precise value is not very easy as the corner singular values decay super fast and the measured value is affected by the cutoff.
https://journals.aps.org/prb/abstract/10.1103/PhysRevB.96.121115 finds `xi_AKLT = 2.065` using the boundary state of a semi infinite cylinder with.
Using an asymmetric CTMRG with a cutoff of 1e-13, I converge at `chi=48` with `xi_AKLT = 1.976` 
Using an asymmetric CTMRG with a cutoff of 1e-20, I converge at `chi~430` with `xi_AKLT = 2.004` (quite sensitive to details) 
With a C4v symmetric CTMRG with `chi=104`, I find `2.010`
With a C4v symmetric CTMRG with `chi=202`, I find `2.028`

I guess "between `1.9` and `2.1` should do the trick

ogauthe:
As discussed, I find it slightly misleading that by default the returned correlation may not be the largest one. Especially that it depends on which symmetry is implemented: in the AKLT example, `correlation_length(ψ, env)` returns 2 different results depending whether `ψ` makes use of U(1) or SU(2). As a user, I would expect that it returns the largest one by default, then `sector` is an advanced mode with finer control. Of course I am not using MPSKit so I may be unaware of established conventions.

Issue #215, opened 2025-06-09
2nd neighbor simple update results


ogauthe:
I try to reproduce my results on the J1-J2 model at finite temperature using PEPSKit. I use the 3 site simple update code from #171. An example of script can be found at https://github.com/ogauthe/BenchmarkPEPS/blob/main/run/run_j1j2.jl.

The simple update Hamiltonian is defined as
```julia
    terms = [CartesianIndex.(t) => h_pa for t in nearest_neighbours(lattice)]  # 1st order SU, 1st neighbor
    append!(
            terms,
            [CartesianIndex.(t) => J2 * h_pa for t in next_nearest_neighbours(lattice)],
        )
    su_hamilt = LocalOperator(physical_spaces, terms...)
```

The rest of the code is nearly the same for the first neighbor.

For J2=0.0, the results look correct and match those obtained with both `frostspin` and `run_heisenberg.jl`. However, for `J2 > 0`, the results do not match frostspin. I am pretty confident in frostspin results as they match high temperature series expansion. It is unclear to me where the difference comes from. It is pretty easy to get a factor 2 wrong, so I computed several data points. There is no match outside of `J2=0`:

 `D=7`, `chi=49`, `beta=1.0`
| J2 / energy |  PEPSKit 1sntei  |  frostspsin  1stnei |   PEPSKit 2nd|       frostspin 2nd |
|-------------|------------------|---------------------|--------------|----------------|
| J2 = 0.0    |   -0.1927        |   -0.1928           | 0.07598      |        0.07599 |
| J2 = 0.15   |   -0.0941        |   -0.1742           | 0.00836      |        0.04712 |
| J2 = 0.30   |   -0.0877        |   -0.1561           |-0.00695      |        0.01630 |
| J2 = 0.60   |   -0.075         |   -0.1196           |-0.0386       |       -0.04736 |

 `D=7`, `chi=49`, `beta=2.0`
| J2 / energy |  PEPSKit 1sntei  |  frostspsin  1stnei |   PEPSKit 2nd|       frostspin 2nd |
|-------------|------------------|---------------------|--------------|----------------|
| J2 = 0.0    |   -0.29060       |   -0.29060        | 0.1576      |        0.1576 |
| J2 = 0.15   |    -0.175        |   -0.267           | 0.04728      |        0.125 |
| J2 = 0.30   |   -0.156        |   -0.2371           |0.0158      |        0.0769 |
[...]

PR #219, opened 2025-06-13
Fix 3-site SU `dt` and reduce artificial C4v breaking

Yue-Zhengyuan:
This PR aims to solve issues of the 3-site simple update discovered in #215 (credit to @ogauthe). 

- Now each `su3site_iter` actually performs evolution over a period of `dt`. In particular, the exponentiation of the Hamiltonian is corrected. For each `⅃`-shaped cluster
```  
     r-1         M3 
                 | 
                 ↓ 
     r   M1 -←- M2 
         c      c+1 
 ``` 
the involved Hamiltonian terms are $H = (H_{12} + H_{23}) / 2 + H_{13}$ (the 1/2 factor is due to repeated counting from `L`-shaped clusters). Previously I mistakably exponentiated each H-term first and add them up:
```math
\exp({-\epsilon H_{12}})/2 + \exp({-\epsilon H_{23}})/2 + \exp({-\epsilon H_{13}})
\quad \text{(wrong)}
```
But this is approximately
```math
\begin{aligned}
& \frac{1}{2}(1 - \epsilon H_{12})
+ \frac{1}{2}(1 - \epsilon H_{23})
+ (1 - \epsilon H_{13}) + O(\epsilon^2)
\\
&= 2 - \epsilon \left[
\frac{1}{2}(H_{12} + H_{23}) + H_{13}
\right] + O(\epsilon^2)
= 2 \, \exp({-\epsilon H / 2})
\end{aligned}
```
with a useless extra factor of 2 in the front, and it reduced the time step to $\epsilon / 2$. Now I directly calculate $\exp({-\epsilon H})$ for each cluster. 

- The C4v symmetry is better preserved by more thorough rotation of the iPEPS during evolution. Previously, I only did `rotl90` -> `rotr90` in each iteration. Now I do `rotl90` four times. As a result now each NN bond is being updated 4 times in each iteration, so the exponentiation of the Hamiltonian is done with `alg.dt / 4`.

### DIFF
--- a/src/algorithms/time_evolution/evoltools.jl
+++ b/src/algorithms/time_evolution/evoltools.jl
@@ -181,19 +181,20 @@ Obtain the 3-site gate MPO on the southeast cluster at position `[row, col]`
         c      c+1
 ```
 """
-function _get_gatempo_se(gate::LocalOperator, row::Int, col::Int)
-    Nr, Nc = size(gate.lattice)
+function _get_gatempo_se(ham::LocalOperator, dt::Number, row::Int, col::Int)
+    Nr, Nc = size(ham.lattice)
     @assert 1 <= row <= Nr && 1 <= col <= Nc
-    unit = id(space(gate.terms[1].second, 1))
+    unit = id(space(ham.terms[1].second, 1))
     sites = (
         CartesianIndex(row, col),
         CartesianIndex(row, col + 1),
         CartesianIndex(row - 1, col + 1),
     )
-    nb1x = get_gateterm(gate, (sites[1], sites[2]))
-    nb1y = get_gateterm(gate, (sites[2], sites[3]))
-    nb2 = get_gateterm(gate, (sites[1], sites[3]))
+    nb1x = get_gateterm(ham, (sites[1], sites[2]))
+    nb1y = get_gateterm(ham, (sites[2], sites[3]))
+    nb2 = get_gateterm(ham, (sites[1], sites[3]))
     op = (1 / 2) * (nb1x ⊗ unit + unit ⊗ nb1y) + permute(nb2 ⊗ unit, ((1, 3, 2), (4, 6, 5)))
+    op = exp(-dt * op)
     return gate_to_mpo3(op)
 end
 
@@ -201,7 +202,7 @@ end
 Construct the 3-site gate MPOs on the southeast cluster 
 for 3-site simple update on square lattice.
 """
-function _get_gatempos_se(gate::LocalOperator)
-    Nr, Nc = size(gate.lattice)
-    return collect(_get_gatempo_se(gate, r, c) for r in 1:Nr, c in 1:Nc)
+function _get_gatempos_se(ham::LocalOperator, dt::Number)
+    Nr, Nc = size(ham.lattice)
+    return collect(_get_gatempo_se(ham, dt, r, c) for r in 1:Nr, c in 1:Nc)
 end
--- a/src/algorithms/time_evolution/simpleupdate3site.jl
+++ b/src/algorithms/time_evolution/simpleupdate3site.jl
@@ -414,13 +414,13 @@ function su3site_iter(
         ),
     )
     peps2 = deepcopy(peps)
-    for i in 1:2
+    for i in 1:4
         for site in CartesianIndices(peps2.vertices)
             r, c = site[1], site[2]
             gs = gatempos[i][r, c]
             _su3site_se!(r, c, gs, peps2, alg)
         end
-        peps2 = (i == 1) ? rotl90(peps2) : rotr90(peps2)
+        peps2 = rotl90(peps2)
     end
     return peps2
 end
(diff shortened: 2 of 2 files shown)


src/algorithms/time_evolution/evoltools.jl
```diff
@@ -181,27 +181,28 @@ Obtain the 3-site gate MPO on the southeast cluster at position `[row, col]`
...
         CartesianIndex(row - 1, col + 1),
     )
-    nb1x = get_gateterm(gate, (sites[1], sites[2]))
-    nb1y = get_gateterm(gate, (sites[2], sites[3]))
-    nb2 = get_gateterm(gate, (sites[1], sites[3]))
+    nb1x = get_gateterm(ham, (sites[1], sites[2]))
+    nb1y = get_gateterm(ham, (sites[2], sites[3]))
+    nb2 = get_gateterm(ham, (sites[1], sites[3]))
     op = (1 / 2) * (nb1x ⊗ unit + unit ⊗ nb1y) + permute(nb2 ⊗ unit, ((1, 3, 2), (4, 6, 5)))
+    op = exp(-dt * op)
```
lkdvos:
```suggestion
    @tensor Odt[i' j' k'; i j k] := -dt * ((nb1x[i' j'; i j] * unit[k' k] + unit[i'; i] * nb1y[j' k'; j k]) / 2 + nb2[i' k'; i k] * unit[j'; j])
    op = exp(Odt)
```
This is just a suggestion because I think it reads slightly easier, but also because it avoids a bunch of intermediate objects

src/algorithms/time_evolution/evoltools.jl
```diff
@@ -181,27 +181,28 @@ Obtain the 3-site gate MPO on the southeast cluster at position `[row, col]`
         c      c+1
 ```
 """
-function _get_gatempo_se(gate::LocalOperator, row::Int, col::Int)
-    Nr, Nc = size(gate.lattice)
+function _get_gatempo_se(ham::LocalOperator, dt::Number, row::Int, col::Int)
+    Nr, Nc = size(ham.lattice)
     @assert 1 <= row <= Nr && 1 <= col <= Nc
-    unit = id(space(gate.terms[1].second, 1))
+    unit = id(space(ham.terms[1].second, 1))
```
lkdvos: As a side note: this will fail if the spaces on different sites are different, which in principle would be allowed.
Not necessarily something that has to be fixed now, but good to keep in mind.

Additionally, `physicalspace(ham)[row, col]` is probably cleaner.

src/algorithms/time_evolution/evoltools.jl
```diff
@@ -181,27 +181,28 @@ Obtain the 3-site gate MPO on the southeast cluster at position `[row, col]`
         c      c+1
 ```
 """
-function _get_gatempo_se(gate::LocalOperator, row::Int, col::Int)
-    Nr, Nc = size(gate.lattice)
+function _get_gatempo_se(ham::LocalOperator, dt::Number, row::Int, col::Int)
+    Nr, Nc = size(ham.lattice)
```
lkdvos: How about directly defining `Base.size(H::LocalOperator)` like this?

lkdvos (commented):
I realize I left a bunch of slightly unrelated comments here, but it might be good to fix this while we're at it. Otherwise looks good to go for me!

lkdvos (dismissed):
Update: I realized none of these comments are really relevant to the specific PR, so I'm okay with merging this as is and leaving the rest for follow-up PRs

lkdvos:
Additional comment: it seems @ogauthe still cannot get the results to agree so maybe we should hold off for a second

Yue-Zhengyuan:
I think I get `dt` wrong again... I should divide `dt` by 2 instead of 4. Hopefully except the size of `dt` there are no other serious issues.

src/algorithms/time_evolution/evoltools.jl
Yue-Zhengyuan: I'll apply this in next push.

src/algorithms/time_evolution/evoltools.jl
Yue-Zhengyuan: There may be many other places that can use `Base.size(H::LocalOperator)`. It's better to make the change in a separate PR.

Yue-Zhengyuan:
I took a closer look of the J1-J2 SU + AD example. It turns out that the SU result there also have C4v symmetry breaking (see [here](https://github.com/QuantumKitHub/PEPSKit.jl/issues/215#issuecomment-2972222220)). I thought I solved it before but didn't. We then need to regenerate the J1-J2 example before merging this PR.

src/operators/localoperator.jl
```diff
@@ -105,6 +105,10 @@ function physicalspace(O::LocalOperator)
     return O.lattice
 end
 
+function Base.size(O::LocalOperator)
+    return size(O.lattice)
+end
```
lkdvos:
```suggestion
Base.size(O::LocalOperator) = size(physicalspace(O))
```

This is slightly more generic, if we ever decide to change how this is stored or the fieldname etc.

Yue-Zhengyuan:
Indeed we have discussed before that storing the Trotter gates as a LocalOperator is a little bit weird. But the only difference is that `exp(-dt * H)` is the _product_ of each gate (up to 1st order Trotter decomposition), instead of their sum. So I think the data structure that stores the Trotterized Hamiltonian will be essentially the same as the current LocalOperator. I just mentally shift to a different interpretation of each term when doing time evolution. 

To check each evolution step actually corresponds `dt`, we can use either real time evolution (I use this in the full update PR) or finite temperature calculation (like Olivier's test).

PR #222, opened 2025-06-13
Add (fast) full update (attempt 2)

Yue-Zhengyuan:
This PR tries again to add the fast full update algorithm ([arXiv 1503.05345](https://arxiv.org/abs/1503.05345)) (see also discussion in the old PR #106). It uses the unified interface for time evolution algorithms introduced in #280.

The performance of FU is benchmarked with the real-time evolution of the transverse field Ising model (TFIM) (see `test/timeevol/tf_ising_realtime.jl`), which calculates the magnetization in the x-direction after the infinite transverse field is reduced to a finite value. 

In addition, I added the option to use pseudo-inverse to solve linear equations in `ALSTruncation`. Otherwise the test on TFIM will fail (but `FullEnvTruncation` is still fine with `KrylovKit.linsolve`). 

Limitations:

- Current implementation does not support bipartite unit cell structure. We can try it later after proposals in #207 are implemented. But it won't work for finite-T calculations anyway (?), so maybe it's not quite worth it...
- Only nearest neighbor Hamiltonians are supported (in principle support for next-nearest neighbor terms can be added; see [arXiv 1711.07584](https://arxiv.org/abs/1711.07584)). 

To do later:

- [x] Add support for PEPO evolution (the purification approach only).


Yue-Zhengyuan:
@ogauthe You can now try the full update on models you are interested in. You can know the basic usage of the related functions from the test file `test/timeevol/tf_ising_fu.jl`.

lkdvos:
@Yue-Zhengyuan , can you let me know what the status is on this (or other) PRs, and which ones should be reviewed/merged? I would like to consider releasing a new version of PEPSKit somewhere in the week of July 7th.

Yue-Zhengyuan:
> ```julia
> time_evolve(wpeps, hamilt, su_alg)
> time_evolve(ψ, env, hamilt, fu_alg)
> ```

This can be a good improvement for future work. YASTN started on [time evolution](https://github.com/yastn/yastn/blob/master/yastn/tn/fpeps/_evolution.py) earlier than us and they have a unified interface for different algorithms. I didn't use it mainly because I'm more familiar with how TensorKit handles fermions. But currently I really don't have the energy for interface revamps to catch up with them...

In particular, `fu_iter2` does not have an SU counterpart: it packs several FU steps together, because in fast FU we don't want to reconverge the CTMRG environment after every `fu_iter`. 

Also, FU uses the usual `InfinitePEPS`, but SU uses `InfiniteWeightPEPS`. I actually regret writing SU in this way. Looking back on it, I only need to introduce `SUWeight`, which can later be generalized to `BPEnv` (#223). SU is actually FU with the bond environment given by the weights instead of the CTM tensors. After `BPEnv` is finished, we can unify SU, FU interface to

```
time_evolve(ψ::InfinitePEPS, env::BPEnv, H, alg::SimpleUpdate)
time_evolve(ψ::InfinitePEPS, env::CTMRGEnv, H, alg::FullUpdate)
```

Yue-Zhengyuan:
@ogauthe I have added full update for finite-T PEPO. I think it should be faster than the old workaround which always keeps the ancilla leg and the physical leg fused. The fusing is now avoided when applying the gate and truncating the bond (unfortunately it seems that I can avoid the fusing only for nearest neighbor gates). If you have time, could you please help doing a comparison?

In SU, I have two ways to evolve ρ: 
- exp(-H dβ) ρ, if the expectation values are to be calculated from ⟨ρ|H|ρ⟩ (purification 2-layer approach);
- exp(-H dβ/2) ρ exp(-H dβ/2), if the expectation values are to be calculated from tr(ρH) (1-layer approach). 

But in FU, I have to use the 2-layer CTMRG environment when truncating the bonds. So I decide to only keep the 2-layer approach, without adding `gate_bothsides` option to `fullupdate` and `fu_iter`.

I also did some preliminary test by comparing FU with SU results in `test/examples/tf_ising_finiteT.jl`. The magnetization produced by FU can deviate a lot from HOTRG or SU. I'll try if using a larger D and CTMRG χ can help when I got some more time.

Yue-Zhengyuan:
Figured out what happened. Before adding the normalization, the fullinf env has a small norm. Since I use `atol` in `truncerror`, the resulting CTMRG bond dimension after truncation is smaller than it should be. This smaller bond dimension makes convergence of the environment easier and faster.

PR #223, opened 2025-06-14
Belief propagation gauge fixing

lkdvos:
This is an initial implementation for gauge fixing a PEPS based on the algorithm described [here](https://www.scipost.org/SciPostPhys.15.6.222?acad_field_slug=chemistry)

- Tindall, Joseph, and Matthew Fishman. “Gauging Tensor Networks with Belief Propagation.” SciPost Physics 15, no. 6 (December 1, 2023): 222. https://doi.org/10.21468/SciPostPhys.15.6.222.

From what I can tell it is mostly functional, although it requires some tests etc to actually see how well it performs in practice. 
The idea would be to attempt to stabilize some gradient methods by re-gauging the state every N steps.

### TODO:

- [x] Support for fermions.
- [x] Enforce positive semi-definiteness of message matrices.
- [x] Allow any virtual leg arrow directions in the PEPS.
- [x] Rotation (`rotl90` etc.) of `BPEnv`.
- [x] Conversion from `BPEnv` to `CTMRGEnv` and `SUWeight`.

### DIFF
--- a/src/environments/bp_environments.jl
+++ b/src/environments/bp_environments.jl
@@ -55,9 +55,9 @@ end
     ) where {A <: AbstractMatrix{<:ProductSpace}}
 
 Construct a BP environment by specifying matrices of north and east virtual spaces of the
-corresponding `InfiniteSquareNetwork`. Each matrix entry corresponds to a site in the unit cell.
+corresponding [`InfiniteSquareNetwork`](@ref). Each matrix entry corresponds to a site in the unit cell.
 
-When `posdef` is true, all messages will be made semi-posdef definite
+When `posdef` is true, all messages will be made semi-positive definite
 (when interpreted as a `bra ← ket` TensorMap).
 
 Each entry of the `Ds_north` and `Ds_east` matrices corresponds to an effective local space


Yue-Zhengyuan:
I remember that (at least for nearest neighbor Hamiltonians) BP is equivalent to simple update ([PhysRevResearch.3.023073](https://journals.aps.org/prresearch/abstract/10.1103/PhysRevResearch.3.023073)), although I haven't looked at the details (such as whether the bond messages are guaranteed to be real diagonal, and thus already represented by `SUWeight`). We may use this equivalence to avoid introducing redundant structs, or generalize the existing simple update algorithm.

BTW, YASTN people are also working on BP, and with NNN Hamiltonian support (in one of the branches) as well.

lkdvos:
@pbrehmer these are some really cool plots! This was indeed one of the primary use-cases I had in mind, so it would be cool to see how this alters these gradient plots.

As a side note, at such low bond dimensions I would also be slightly cautious about picking bond dimensions that necessarily break the symmetry, for example in a U1 case with charge conjugation I would expect that you might need to check for each D whether or not it gives a valid combination. This is motivated by similar problems for MPS simulations, where certain bond dimensions don't work because they have to cut inbetween multiplets.

GlebFedorovich:
@pbrehmer Indeed, we have seen this a lot for periodic PEPS, but pushing $$\chi$$ and imposing some rotation and mirror symmetries help to stabilize the optimization and avoid these discontinuous points. However, there are many cases, beyond ising and heisenberg models, when imposing spatial symmetries leads to higher energies compared let's say with just SU (or with AD fixed-point one before the optimization arrives at "bad" point :) )

src/environments/bp_environments.jl
```diff
@@ -0,0 +1,98 @@
+struct BPEnv{T}
+    "4 x rows x cols array of message tensors, where the first dimension specifies the spatial direction"
```
Yue-Zhengyuan, line 27: Could you please add a unicode diagram illustrating the convention?

src/algorithms/bp/beliefpropagation.jl
```diff
@@ -0,0 +1,138 @@
...
+    # Compute belief propagation fixed point solutions
+    env, err = bp_fixedpoint(env, InfiniteSquareNetwork(psi), alg)
+
+    # Bring PEPS to the Vidal gauge
+    sqrtmsgs = map(env.messages) do M
+        U, S, Vᴴ = tsvd!(M)
+        sqrtM = U * sdiag_pow(S, 1 / 2) * Vᴴ
+        isqrtM = U * sdiag_pow(S, -1 / 2) * Vᴴ
+        return sqrtM, isqrtM
+    end
```
Yue-Zhengyuan: In the original BP gauging paper, the message tensors are chosen to be positive definite. Here I see that `BPEnv` are initialized simply with `randn` without enforcing positiveness. I'm curious whether this is valid.

Yue-Zhengyuan (commented):
Since adding BP may need some revisions for SU and InfiniteWeightPEPS, I want to ask a few things to get more familiar with it.

src/environments/bp_environments.jl
```diff
@@ -0,0 +1,98 @@
...
+    return BPEnv(f, T, N, fill(D_north, unitcell), fill(D_east, unitcell))
+end
+
+function BPEnv(network::InfiniteSquareNetwork)
+    Ds_north = _north_env_spaces(network)
+    Ds_east = _east_env_spaces(network)
+    return BPEnv(randn, scalartype(network), Ds_north, Ds_east)
+end
+function BPEnv(f, T, network::InfiniteSquareNetwork)
+    Ds_north = _north_env_spaces(network)
```
kshyatt: `_north_env_spaces` (and `_east_env_spaces`) doesn't seem to be defined?

src/algorithms/bp/beliefpropagation.jl
lkdvos: Good point, we should probably enforce that as well

src/environments/bp_environments.jl
Yue-Zhengyuan: I'll try doing this as I learn through the algorithm

src/algorithms/contractions/bp_contractions.jl
```diff
@@ -0,0 +1,252 @@
...
+            conj(bra(A)[d; DNb DEb DSb DWb]) *
+            M_south[DSt; DSb] *
+            M_west[DWt; DWb] *
+            M_north[DNt; DNb]
+    end
+end
+
+# Belief Propagation Expectation values
+# -------------------------------------
+function MPSKit.expectation_value(peps::InfinitePEPS, O::LocalOperator, env::BPEnv)
```
Yue-Zhengyuan, line 53: I kind of want to replace `expectation_value`, `contract_local_operator`, `contract_local_norm` etc with wrappers over the CTMRGEnv versions using the BPEnv-to-CTMRGEnv conversion.

lkdvos: This is definitely okay for testing, but probably won't be great for actual performance, as the contraction order for ctmrg assumes that the boundary dimension is D^2, not 1

kshyatt:
Should the doc build here be fixed by https://github.com/QuantumKitHub/PEPSKit.jl/pull/301?

src/algorithms/bp/gaugefix.jl
```diff
@@ -0,0 +1,81 @@
...
+        #= 
+        - dir = 1: x-bond (r,c) → (r,c+1)
+            m[(r,c) → (r,c+1)] = env[4, r, c]
+            m[(r,c+1) → (r,c)] = env[2, r, c+1]
+        - dir = 2: y-bond (r,c) → (r-1,c)
+            m[(r,c) → (r-1,c)] = env[3, r, c]
+            m[(r-1,c) → (r,c)] = env[1, r-1, c]
+        =#
+        MM = if dir == 1
+            transpose(sqrtmsgs[WEST, r, c][1]) * sqrtmsgs[EAST, r, _next(c, end)][1]
```
Yue-Zhengyuan: @lkdvos Could you please explain your choice to use `transpose` instead of `permute` here? As @leburgel trying to make BP differentiable, we might need to use `permute` with appropriate `twist`s in the end.

lkdvos: I think the transpose is exactly there because I don't want any twists here, and I think this should still be differentiable?

lkdvos (commented):
Thanks for continuing working on this, looks nice!

I have a couple things that I would like to bring up, that I still remember from when I was implementing this.

The first is that because of the layout of the algorithm, at some point I was wondering if it would actually make sense to have the south and west messages defined as the adjoint of what they are right now. The idea being the same as how typically for left/right eigenvectors we write `A * R` and `L' * A`, and I somehow had in my head that this would make things a bit more symmetric. Additionally, I think that means that for standard arrows no twists appear, while otherwise that would be the case. (although I have to double check this)

The second is that I need to double check if I implemented the efficient stopping criterion, rather than simply playing around with the default slow one.

I'm not a huge fan of the updated convention for the coordinates, since I find it a bit unintuitive if I just consider the algorithm itself, but I understand that the similarity to CTMRGEnv is also important.

Finally, I don't remember what we decided in the end, but I would not be opposed to gaugefixing resulting in a state that has different arrows than the original one. We can still add in convenience flipping functionality to fix that afterwards, but I am not a big fan of always paying that price. (This is also how MPSKit works, this will change the arrows to the canonical convention)

I should have some time next week to look into this myself as well though.

src/algorithms/bp/gaugefix.jl
leburgel: I didn't look into why much, but when trying this out it seemed `transpose` wasn't differentiable for non-trivial symmetries (specifically SU(2) in my test case). It errored somewhere when trying to pullback `fusionblocks` or something like that, so I assumed there was just no rrule for this and it got stuck in the automatic pullback.

lkdvos: Oh ok, I'll just add the rule then, shouldn't be too hard :)

lkdvos: https://github.com/QuantumKitHub/TensorKit.jl/pull/319

lkdvos:
> Here we are using this choice, which I think is more natural. Although it requires transposing during gauging, I think it makes the BP equations (in bp_contractions.jl) simpler. Otherwise implicit permutes (or transposes) are needed in BP update @tensor calls, which is more error-prone for fermions.

I'm pretty sure that the implicit permutes shouldn't affect anything, at least that's what the `@tensor` macro should take care of. My suggestion for the `adjoint definition would result in the following contractions by the way:

<details>

```julia
function contract_north_message(
        A::PEPSSandwich, M_west::PEPSMessage, M_north::PEPSMessage, M_east::PEPSMessage
    )
    return @autoopt @tensor begin
        M_north′[DSt; DSb] :=
            ket(A)[d; DNt DEt DSt DWt] * conj(bra(A)[d; DNb DEb DSb DWb]) *
            conj(M_west[DWt; DWb]) * M_north[DNt; DNb] * M_east[DEt; DEb]
    end
end
function contract_east_message(
        A::PEPSSandwich, M_north::PEPSMessage, M_east::PEPSMessage, M_south::PEPSMessage
    )
    return @autoopt @tensor begin
        M_east′[DWt; DWb] :=
            ket(A)[d; DNt DEt DSt DWt] * conj(bra(A)[d; DNb DEb DSb DWb]) *
            M_north[DNt; DNb] * M_east[DEt; DEb] * conj(M_south[DSt; DSb])
    end
end
function contract_south_message(
        A::PEPSSandwich, M_east::PEPSMessage, M_south::PEPSMessage, M_west::PEPSMessage
    )
    return @autoopt @tensor begin
        M_south′[DNt; DNb] :=
            conj(ket(A)[d; DNt DEt DSt DWt]) * bra(A)[d; DNb DEb DSb DWb] *
            conj(M_east[DEt; DEb]) * M_south[DSt; DSb] * M_west[DWt; DWb]
    end
end
function contract_west_message(
        A::PEPSSandwich, M_south::PEPSMessage, M_west::PEPSMessage, M_north::PEPSMessage
    )
    return @autoopt @tensor begin
        M_west′[DEt; DEb] :=
            conj(ket(A)[d; DNt DEt DSt DWt]) * bra(A)[d; DNb DEb DSb DWb] *
            M_south[DSt; DSb] * M_west[DWt; DWb] * conj(M_north[DNt; DNb])
    end
end

absorb_north_message(A::PEPSTensor, M::PEPSMessage) =
    @tensor A′[d; N' E S W] := A[d; N E S W] * M[N; N']
absorb_east_message(A::PEPSTensor, M::PEPSMessage) =
    @tensor A′[d; N E' S W] := A[d; N E S W] * M[E; E']
absorb_south_message(A::PEPSTensor, M::PEPSMessage) =
    @tensor A′[d; N E S' W] := A[d; N E S W] * conj(M[S; S'])
absorb_west_message(A::PEPSTensor, M::PEPSMessage) =
    @tensor A′[d; N E S W'] := A[d; N E S W] * conj(M[W; W'])
```

</details>

which is the equivalent of defining `M_south_new = adjoint(transpose(M_south))` and the same for `M_west_new`. Unless I'm mistaken, that would lead to the following permutation-free message combination:
```julia
MM = M_north * M_south'
```

What I like about this is that all index orders are still top to bottom and the spaces are still equal for north and south, there are no unneeded permutations, and this is the same kind of notation as thinking about eigenvectors like:

```math
\begin{align}
A \cdot r &= \lambda \cdot r \\
A^\dagger \cdot l &= \lambda \cdot l \\
\implies l^\dagger \cdot A &= \lambda^* \cdot l^\dagger
\end{align}
```

> For SUWeight, I do want to make its axis order independent of virtual arrow directions. But it is better to deal with it in a follow-up PR

Fully agree to not tackle any of the SU code in this PR, I just wanted to briefly think about it.

> A similar problem exists here for fermions when we initialize the messages with identity operators. 

This is a very good point. I think this is a great argument in favor of just canonicalizing the arrows before the procedures, and flipping back after. I'll implement this accordingly for BP in this PR already.

---

By the way, how do you feel about me just restricting the BP implementation for two-layer for now, and not exporting it? I think most of it isn't actually working for PEPO nor partition functions and I'm not even sure yet how to handle these (e.g. hermitian projection does not really make sense)

leburgel:
> We probably will have to tinker a bit more with the `BeliefPropagation` struct in the near future, since there are a bunch of settings missing. Thinking of pseudo-inverse cutoffs, SVD/Eig algorithms, but also of "schemes" for doing the BP iterations. Currently we are updating all messages in parallel, while in principle we could snake or alternate updates which might be more efficient for larger unit cells.

Would it make sense to split the `BeliefPropagation` struct off from the gauge fixing algorithm, which would then be some other struct (`BPGauge`, or any other suitable name). The reason would be that `BeliefPropagation` actually contains all the settings needed to actually run BP and find the fixed point, i.e. the `BPEnv` (up to the iteration "schemes", these definitely also belong here). But things like pseudo-inverse cutoffs and decomposition algorithms are only needed when using the `BPEnv` to do an actual gauge fixing, which is a whole extra step on top that we don't always do I think. For example, to do a crude approximate contraction of a network with a `BPEnv` we only need the current `BeliefPropagation` struct, without any roots or decompositions.

This would also be compatible with other gauging algorithms, which also require an algorithm subroutine such as `BeliefPropagation` (e.g. an `LBFGS` or `Arnoldi` subroutine).

Yue-Zhengyuan:
> how do you feel about me just restricting the BP implementation for two-layer for now, and not exporting it?

I cannot imagine how to do BP gauging with messages with $n \ne 2$ legs either (though BP itself is fine, so one more reason to separate BP and gauging). So definitely OK to restrict to PEPS-norm networks.

pbrehmer:
> Would it make sense to split the BeliefPropagation struct off from the gauge fixing algorithm, which would then be some other struct (BPGauge, or any other suitable name).

I think this would in general make sense since there are a few different gauging algorithms which we might want to try out at some point (BP, MCF, maximal fidelity, ...). Would you also think it'd be a good idea to use the same function (`gauge_fix`) and same abstract algorithm struct (for example `GaugeFixingAlgorithm`) to also gauge fix the PEPS environments? Because in principle there are also different styles, namely those in the appendix from Bram and Anna's [paper](https://arxiv.org/pdf/2311.11894) and we also need a specialized `gauge_fix` for C4v environments.

src/algorithms/bp/gaugefix.jl
```diff
@@ -0,0 +1,77 @@
...
+    sqrtMᴴ, isqrtMᴴ = sqrt_invsqrt(transpose(Mᴴ); ishermitian)
+
+    U, Λ, Vᴴ = svd_compact!(sqrtM * sqrtMᴴ)
+    if !isdual(space(Mᴴ, 1))
+        U, Λ′, Vᴴ = flip_svd(U, Λ, Vᴴ)
+        Λ = transpose(Λ′)
+    end
+    sqrtΛ = sdiag_pow(Λ, 1 / 2)
+    X = isqrtM * U * sqrtΛ
+    invX = sqrtΛ * Vᴴ * isqrtMᴴ
```
Yue-Zhengyuan: It might be safer to `flip_svd` *after* `X` and `invX` have been obtained using the "canonical" virtual arrow.

```suggestion
    U, Λ, Vᴴ = svd_compact!(sqrtM * sqrtMᴴ)
    sqrtΛ = sdiag_pow(Λ, 1 / 2)
    X = isqrtM * U * sqrtΛ
    invX = sqrtΛ * Vᴴ * isqrtMᴴ
    if !isdual(space(Mᴴ, 1))
        X, Λ, invX = flip_svd(X, Λ, invX)
    end
```

Yue-Zhengyuan:
After merging #315 we can try bringing back the BPEnv to SUWeight conversion as part of the gauging process. And possibly include SU-gauging as well (at least for test purposes)?

src/algorithms/bp/gaugefix.jl
```diff
@@ -0,0 +1,76 @@
...
+    elseif dir == EAST
+        psi[row, col] = absorb_east_message(psi[row, col], X)
+        psi[row, _next(col, end)] = absorb_west_message(psi[row, _next(col, end)], transpose(invX))
+    end
+
+    return psi, X, invX
+end
+
+function sqrt_invsqrt(A; ishermitian::Bool = true)
+    if ishermitian
```
Yue-Zhengyuan: Does `eigh_full(A)` also check if `A` is Hermitian? If so, then why not

```suggestion
    if ishermitian(A)
```

so we don't need to pass `ishermitian` manually.

src/algorithms/bp/gaugefix.jl
```diff
@@ -0,0 +1,111 @@
...
+        return isdual(space(sqrtM, 1)) ? _fliptwist_s(Λ) : Λ
+    end
+    return SUWeight(wts)
+end
+
+function sqrt_invsqrt(A; ishermitian::Bool = true)
+    if ishermitian
+        D, V = eigh_full(A)
+        sqrtA = V * sdiag_pow(D, 1 / 2) * V'
+        isqrtA = V * sdiag_pow(D, -1 / 2) * V'
```
Yue-Zhengyuan, line 114: Here we have an unfortunate situation with fermions: because of the axis order `ket (top) <- bra (bottom)`, the eigenvalues in the parity-odd sector are negative. So, we cannot directly apply `sdiag_pow`.

src/algorithms/bp/gaugefix.jl
kshyatt: https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/main/src/implementations/eigh.jl#L41 if it's using this `eigh_full!` then yes

src/algorithms/bp/gaugefix.jl
Yue-Zhengyuan: Even for bosons, if the BPEnv is initialized with `randn` instead of `ones`, there can also be negative eigenvalues of the messages. In more extreme cases there can be complex eigenvalues. In such (non-Hermitian) cases I guess the gauge fixing should still be done with SVD instead of eigen-decomposition.

Yue-Zhengyuan: BTW, because of the kwarg `tol::Real = eps(scalartype(s))^(3 / 4)`, `sdiag_pow` cannot handle complex DiagonalTensorMaps unless `tol` is explicitly provided.

src/algorithms/bp/gaugefix.jl
Yue-Zhengyuan: I see that the non-mutating `eigh_full` is also calling `eigh_full!` under the hood, so I plan to let the code automatically check if the input message is Hermitian.

Yue-Zhengyuan: @lkdvos @kshyatt Now I think it is better to check `isposdef(A)`. Is `isposdef` also relatively cheap and differentiable?

Yue-Zhengyuan:
The tests now pass for all cases I can think of (bosons vs fermions, positive vs arbitrary messages, and standard vs random virtual PEPS arrows). For fermions I introduced some twists to make things work at some seemingly unexpected positions. I'll try to justify them later.

> * It might be more convenient in the long run to have the messages follow the "canonical arrow" convention, such that north and east we get `top <- bot` and south and west we get `bot <- top`

Now I use this axis order convention suggested by @lkdvos earlier, without additional adjoints.

src/algorithms/bp/gaugefix.jl
kshyatt: `isposdef` computes the Cholesky factorization if the input matrix isn't already diagonal, I think. There should be a builtin `ChainRules.jl` rule for `cholesky`.

src/algorithms/bp/beliefpropagation.jl
```diff
@@ -0,0 +1,111 @@
...
+    "Stopping criterion for the BP iterations in relative trace norm difference"
+    tol::Float64 = 1.0e-6
+
+    "Minimal number of BP iterations"
+    miniter::Int = 2
+
+    "Maximal number of BP iterations"
+    maxiter::Int = 50
+
+    "Toggle for projecting onto the hermitian subspace"
```
kshyatt: Can this get a little more descriptive? Where/when is the projection done?

src/algorithms/bp/gaugefix.jl
```diff
@@ -0,0 +1,127 @@
+"""
+    struct BPGauge
+
+Algorithm for gauging PEPS with belief propagation fixed point messages.
+"""
+@kwdef struct BPGauge
+    # TODO: add options
```
kshyatt, line 7: Is this TODO still relevant?

src/algorithms/contractions/bp_contractions.jl
```diff
@@ -0,0 +1,249 @@
...
+        bra::InfinitePEPS,
+        env::BPEnv,
+    )
+    ind_relative = inds[2] - inds[1]
+    return if ind_relative == CartesianIndex(1, 0)
+        contract_vertical_operator(inds[1], O, ket, bra, env)
+    elseif ind_relative == CartesianIndex(0, 1)
+        contract_horizontal_operator(inds[1], O, ket, bra, env)
+    else
+        error("Not implemented")
```
kshyatt: More descriptive message here would be nice

src/environments/bp_environments.jl
```diff
@@ -0,0 +1,225 @@
...
+    return posdef ? (m' * m) : m
+end
+
+"""
+    BPEnv(
+        [f=randn, T=ComplexF64], Ds_north::A, Ds_east::A; posdef::Bool = true
+    ) where {A <: AbstractMatrix{<:ProductSpace}}
+
+Construct a BP environment by specifying matrices of north and east virtual spaces of the
+corresponding `InfiniteSquareNetwork`. Each matrix entry corresponds to a site in the unit cell.
```
kshyatt: Link to `InfiniteSquareNetwork` docstring here?

src/algorithms/bp/gaugefix.jl
Yue-Zhengyuan: See @leburgel's comment here: https://github.com/QuantumKitHub/PEPSKit.jl/pull/223#issuecomment-3617222668. We may need to add options in the future, but currently `BPGauge` really does nothing besides being a multiple-dispatch label. Besides, we may in the end decide that we don't want to call `isposdef` (maybe to save some resources when backprop-ing through BP gauging) when converting the BPEnv to gauge transformations, but instead directly tell the algorithm that we are using posdef BPEnv. So for now I just leave `BPGauge` here.

src/algorithms/bp/beliefpropagation.jl
Yue-Zhengyuan: The projection is done after updating *each* message using the BP equations. Even when the input message is already Hermitian, the projection can still help avoid accumulation of numerical errors. (Will remember to add it to the docstring)

src/algorithms/bp/gaugefix.jl
lkdvos: I think the idea is that the BP updates should preserve positive definite-ness, so we can hoist this check out of the loop.
I definitely agree that changing this for `isposdef` (probably with a keyword argument `isposdef = isposdef(A)`) that can either be provided or automatically checked is a good idea!

Considering the differentiability, `isposdef` outputs a boolean so I don't think there should be any contributions with respect to that, so worst case scenario we can manually mark that as non-differentiable.

src/algorithms/bp/gaugefix.jl
lkdvos: Sorry, that kwarg should have been `eps(real(scalartype(s)))^(3/4)`.

For the remainder, I indeed messed up: the property to check is positive definiteness (which implies hermitian), and hermitian by itself isn't enough.

However, for the fermions I would like to investigate this more closely. Looking at eq (124) in the fermionic tensor network methods paper, we can investigate the special case of BP where the messages are the identity (e.g. in a tree or MPS where the tensors have been properly gauged), and we see that indeed we expect negative values in the parity-odd sector, so that seems correct.
However, I think we should be able to (and probably want to!) redefine or contractions by adding this twist explicitly, which will then restore the positive definiteness of our messages.

Thinking about that a bit more, I think this is effectively what we were doing by having the other index order for the messages as well, since for these (1, 1) tensors a permutation and a twist + planar transpose should effectively be the same thing.

Concretely, what I'm suggesting here is to explicitly factorize the message into a twist + the message, thereto adding a twist in the bp update contractions, as well as in the expectation value contractions, which will then make everything positive definite again.

I think a good test case could actually be a product state PEPS which has as virtual spaces `Vect[FermionParity](1 => 1)`.
Because of the product state structure, the messages should (up to normalization) all be the identity and positive definite, no matter the unit cell.
I think that for this case, missing twists would make odd unit cell sizes not converge, and even unit cell sizes get staggering positive and negative messages (might be wrong about this though)

Does any of this make sense, or is this starting to feel like too much? Feel free to disagree with me here, I think most of my reasoning here comes from the intuition we obtained from that paper, which I'm trying to couple back to, but alternative viewpoints are definitely also appreciated!

lkdvos:
Sorry I commented on the subcomments before I saw your full comment here.
I am guessing that the choice to omit the adjoints is actually exactly what is causing these twists to appear, as an adjoint would transpose without actually having to introduce a minus sign, but I'm not 100% sure about that. In any case, I would also be happy with the current choices, as that is more close to how we described the algorithms in the MPS case in the fermionic methods paper, although I still think in that case the twists are better placed in the contractions, rather than absorbed in the messages.

src/algorithms/bp/gaugefix.jl
Yue-Zhengyuan: > that kwarg should have been `eps(real(scalartype(s)))^(3/4)`

I have made this change. Glad that's the way to go.

> Concretely, what I'm suggesting here is to explicitly factorize the message into a twist + the message, thereto adding a twist in the bp update contractions, as well as in the expectation value contractions, which will then make everything positive definite again.

This is a good idea. The twists I added may actually be implicitly doing this.

src/algorithms/bp/gaugefix.jl
lkdvos: In principle we could put the BP iteration algorithm in here, such that we can automatically construct and converge an environment without having to manually call `leading_boundary`, but I am also fine with merging this without that functionality.

lkdvos:
I cannot approve since this is my own PR, but I think most of the TODOs and comments have been addressed, so I would be happy to get this one merged.

Yue-Zhengyuan:
The only remaining question is whether to absorb the twists into the messages. @lkdvos do you think it's OK to address this later if needed? @kshyatt Maybe you can have a final review and approve.

lkdvos:
Yes, I think this PR has lived for too long and it will be easier to review that change in a separate PR

src/algorithms/bp/gaugefix.jl
```diff
@@ -0,0 +1,127 @@
...
+        return X, Xinv
+    end
+    return psi′, XXinv
+end
+
+function _sqrt_bp_messages(I::CartesianIndex{3}, env::BPEnv)
+    dir, row, col = Tuple(I)
+    @assert dir == NORTH || dir == EAST
+    M12 = env[dir, dir == NORTH ? _prev(row, end) : row, dir == EAST ? _next(col, end) : col]
+    sqrtM12, isqrtM12 = sqrt_invsqrt(twist(M12, 1))
```
Yue-Zhengyuan, line 28: This twist is used to cancel the twist introduced because of the axis order choice `ket <- bra` for the north and east messages, in order to obtain a positive definite map.

src/algorithms/bp/gaugefix.jl
```diff
@@ -0,0 +1,127 @@
...
+    dir, row, col = Tuple(I)
+    @assert dir == NORTH || dir == EAST
+
+    sqrtM12, isqrtM12, sqrtM21, isqrtM21 = _sqrt_bp_messages(I, env)
+    U, Λ, Vᴴ = svd_compact!(sqrtM12 * sqrtM21)
+    sqrtΛ = sdiag_pow(Λ, 1 / 2)
+    X = isqrtM12 * U * sqrtΛ
+    invX = sqrtΛ * Vᴴ * isqrtM21
+    if isdual(space(sqrtM12, 1))
+        X, invX = twist(flip(X, 2), 1), flip(invX, 1)
```
Yue-Zhengyuan, line 61: This twist is used to cancel the twist introduced by `flip` when reversing the arrow direction. It can be equally applied to `invX` instead.

Yue-Zhengyuan: (Will we one day deal with sectors that cannot be flipped?)

PR #224, opened 2025-06-15
Update JuliaFormatter to v2

lkdvos:
Update the formatter to JuliaFormatter v2, keeping the style the same.

As a sidenote, I did play around a bit with the other formatting styles but honestly can't find anything better, so we might as well stick with the same style unless someone has strong opinions on this


pbrehmer (approved):
I would also advocate for sticking with the current formatting style, so this looks good to me :-)

PR #229, opened 2025-06-21
[Perf] Another round of performance improvements

lkdvos:
This is another round of improvements for generic performances.

In particular, the `renormalize_west_edge` function now got the same treatment with optimizing intermediate permutations, and I altered some implementations with `dtmap` as these were completely type-unstable.
At least now, from what I can test, `ctmrg_iteration` is actually type-stable, and while I don't think this would lead to massive speedups, I do believe this might help with compilation times.

### DIFF
--- a/src/PEPSKit.jl
+++ b/src/PEPSKit.jl
@@ -15,7 +15,7 @@ import MPSKit: leading_boundary, loginit!, logiter!, logfinish!, logcancel!, phy
 
 using MPSKitModels
 using FiniteDifferences
-using OhMyThreads: tmap
+using OhMyThreads: tmap, tmap!
 using DocStringExtensions
 
 include("Defaults.jl")  # Include first to allow for docstring interpolation with Defaults values
--- a/src/algorithms/contractions/ctmrg_contractions.jl
+++ b/src/algorithms/contractions/ctmrg_contractions.jl
@@ -1310,12 +1310,19 @@ end
 function renormalize_west_edge(
     E_west::CTMRG_PEPS_EdgeTensor, P_bottom, P_top, A::PEPSSandwich
 )
-    return @autoopt @tensor edge[χ_S D_Eab D_Ebe; χ_N] :=
-        E_west[χ1 D1 D2; χ2] *
-        ket(A)[d; D3 D_Eab D5 D1] *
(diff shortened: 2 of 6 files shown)


pbrehmer:
This is really great, thanks! I have high hopes that the improved type-stable `dtmap` and `ctmrg_iteration` will actually improve the compilation times (especially for the backwards pass with Zygote) since these have gotten really bad.

I was wondering, the other edge renormalization functions (`renormalize_north_edge`, etc.) should have the exact same intermediate permutations, right? The only difference should be which of the bra/ket bond indices will be connected. 

Let me know if I can help out or when I should review!

lkdvos:
For me this would be ready to review/merge, I'm actually 🤒 right now so I won't be able to continue this anytime soon. 
Indeed the other contractions should get the same treatment, I focused on the sequential ctmrg to start with though. It's a bit manual in any case, I can't say I'm super confident I always have the best permutations so I'd say it's better to check explicitly the performances, and that takes quite a bit of effort.

test/ctmrg/flavors.jl
```diff
@@ -1,4 +1,4 @@
-using Test
+using Test, TestExtras
```
pbrehmer: Is this import used anywhere?

pbrehmer (commented):
Looking good! Other than the import thing, this would be good to go for me.

Get well soon!

test/ctmrg/flavors.jl
lkdvos: `@constinferred` is from `TestExtras`. It might be better to have this as `using TestExtras: @constinferred` if you prefer that?

pbrehmer: As far as I can see `TestExtras` is again imported in `test/ctmrg/fixed_iterscheme.jl` and there also `@constinferred` is used. But not in this file, right?

lkdvos: Oh right! I added a check on the `leading_boundary` here but our algorithm selection is not type stable, and therefore `leading_boundary` with keyword arguments also isn't.
I'd argue this is completely fine though, as soon as the algorithm is selected everything is type stable again so the performance impact of that should be rather minor.
As a result I removed the check, but forgot to remove the import. I'll update this first.

pbrehmer (approved):

PR #230, opened 2025-06-21
Reduced density matrices

lkdvos:
Replace the expectation value calculation for PEPS with computing `tr(H * rho)`. 

I checked the contraction costs and the leading cost is the same for both local operators, local norms and local density matrices. (In fact, the optimal contraction order for a single site local operator is exactly first computing the reduced densitymatrix and then contracting with the operator).
The main difference here is that instead of computing both the operator and the norm network, we can simply compute the density matrix once, divide it by its trace, and then evaluate the operator.

I'll try and include some benchmarks as well later, just to prove that everything is behaving as desired, but overall it seems to be between 25-40% faster to evaluate expectation values like this, and the memory usage scales accordingly as well.
Note that this also affects the gradient calculations, and even the compile times since the contraction functions are quite demanding for the compiler.

We should also try and hardcode the "easy" cases to optimize the intermediate permutations, but that might have to wait for a future PR.

### DIFF
--- a/src/PEPSKit.jl
+++ b/src/PEPSKit.jl
@@ -90,8 +90,9 @@ export CTMRGEnv, SequentialCTMRG, SimultaneousCTMRG
 export FixedSpaceTruncation, SiteDependentTruncation
 export HalfInfiniteProjector, FullInfiniteProjector
 export LocalOperator, physicalspace
-export expectation_value, cost_function, product_peps, correlation_length, network_value
-export correlator
+export product_peps
+export reduced_densitymatrix, expectation_value, network_value, cost_function
+export correlator, correlation_length
 export leading_boundary
 export PEPSOptimize, GeomSum, ManualIter, LinSolver, EigSolver
 export fixedpoint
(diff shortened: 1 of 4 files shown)


Yue-Zhengyuan:
`reduced_densitymatrix` can be useful when users want to measure different physical quantities on the same sites (e.g. hopping energy, spin correlation, total energy on the same bond) by avoiding repeated contractions. I suggest to export it for end users.

lkdvos:
Good for me, are we happy with the interface in that case?

lkdvos:
Lets merge #229 first, I'll update this one and see if I find the energy to actually test the performances a bit.

ogauthe:
As a comment, _if_ you plan to support PEPO CTMRG and tensors with ancilla in the future, you may consider implementing it here.
You can dispatch over leg number, add a dummy leg for PEPS tensor (=pure wavefunction) and then use a single contraction function, written with an ancilla. The dummy leg would not add any cost.

PR #234, opened 2025-07-19
Improve normalizations during simple update

Yue-Zhengyuan:
This is a minor improvement to properly normalize the tensors during simple update. 

Previously, after absorbing the weights into the vertex tensors, the norm of the vertex tensors may become too small, causing issues for the following SVD truncation (the singular values can be too small as well).

### DIFF
--- a/src/algorithms/time_evolution/simpleupdate.jl
+++ b/src/algorithms/time_evolution/simpleupdate.jl
@@ -45,6 +45,8 @@ function _su_xbond!(
     sqrtsB = ntuple(dir -> (dir == WEST), 4)
     A = _absorb_weights(A, peps.weights, row, col, Tuple(1:4), sqrtsA, false)
     B = _absorb_weights(B, peps.weights, row, cp1, Tuple(1:4), sqrtsB, false)
+    normalize!(A, Inf)
+    normalize!(B, Inf)
     # apply gate
     X, a, b, Y = _qr_bond(A, B)
     a, s, b, ϵ = _apply_gate(a, b, gate, trscheme)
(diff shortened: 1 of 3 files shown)


Yue-Zhengyuan:
The gradient tests have been running out of time quite frequently in recent GitHub actions. We may need to further investigate it?

lkdvos:
While I can see that this is useful, I'm a bit hesitant to enforce this, mostly because it would prevent us from using this for finite temperature results since the norm is important there. Do you think we can just provide an easy way to manually ensure the norm doesn't run away between simple update steps?

Yue-Zhengyuan:
How about I also return the number used to normalize the PEPS in each 2/3-site SU iteration? But to test it's done correctly, can you provide an example where free energy needs to be extracted?

Yue-Zhengyuan:
> adding a keyword argument `p::Union{Real,Nothing}=1`

Good idea. I'll flesh it out when I get some time. For imaginary time evolution, normalization is inevitable since the norm of $e^{-\beta H}$ can be large as $\beta$ increases.

lkdvos:
Absolutely, you always have to renormalize but it is important to keep track of the norm in that case. I think typically, one computes `log(N)` along the way and simply add them together, but there are definitely different valid strategies. I would say that it is just important to:
A. have an easy way to normalize, with whatever norm we choose
B. have this information not lost in some inner part of the code.

Anything that achieves that is good for me :)

Yue-Zhengyuan:
I increasingly feel the difficulty in writing a `simpleupdate` that suits all different purposes:

| Purpose | Imag/Real time | Need convergence test | Need to remember<br>normalization |
|--:|:-:|:-:|:-:|
| Ground state search | Imag | Yes | No |
| Real time dynamics | Real | No | No | 
| Finite-T density operator | Imag | No | No |
| Free energy (at many T) | Imag | No | Yes |

In the last two cases, in the future I think we'll use iPEPS with 2 physical legs (physical + ancilla; though I'm not sure how much improvement this can achieve), further differentiating them from the first two cases. In addition, if sometimes I return the norm, and sometimes I don't, I guess this cause "type instability"? So I want to:

- Making the current `simpleupdate` an internal function for test purposes, and not export it. 
- Ask end users to build their own SU pipeline from the most basic `su_iter`; I can make it always return the normalization used in each step. 
- If we don't provide `simpleupdate`, then in the algorithm struct `SimpleUpdate`, the convergence criterion `tol` can be removed.

Though there are some things that only need to be done once at the beginning, and are cumbersome to be moved into `su_iter`:
- create the evolution gate $exp(-\epsilon H)$;
- auto dispatch to 2-site or 3-site version of SU.

@lkdvos @sanderdemeyer Do you have any design suggestions? (In full update #222 we have a similar situation.)

lkdvos:
I think I agree with this indeed. I'd propose to do this in different steps though, and not all at the same time to make it a bit more comprehensive

Yue-Zhengyuan:
Sure. The BP things can be done later, and we can use `SUWeight` as `env` for now. But currently I'm busy learning about Gaussian fPEPS, and cannot work on SU for a while.

Yue-Zhengyuan:
To be continued in another PR.

Issue #205, opened 2025-05-28
CTMRG performances


ogauthe:
Hello,

I recently started using PEPSKit. Kudos to all of you for all this work. I run finite temperature simulation and benchmark with my own python code (https://github.com/ogauthe/frostspin). I was happy to observe the results are nearly the same between frostspin and PEPSKit. However the performances were not as good as I was hoping for. I really think they can be improved, as I expect TensorKit to be much more efficient for the `permutedims` part.

Here I propose to give some benchmark points and find if some bottleneck can be improved in PEPSKit.

Setup: I use the simple update code to run a finite temperature simulation of the Heisenberg model on the square lattice. I add an ancilla leg to my PEPS tensor. As #184 is still a work in progress, I merge ancilla and physical leg together and define a Hamiltonian acting on both spaces to reuse PEPSKit code. I start from an infinite temperature product state, use simple update for imaginary time evolution. Once a given beta is reached, I use CTMRG to contract the tensor network and compute observables. I impose SU(2) symmetry. I can share the script if you want to reproduce.

Now there are of course several differences between the two codes. I think the most important ones are
- in frostspin I use and AB//BA unit cell, with only 2 sites.
- in the CTMRG, I cache enlarged corners to reuse them for the next iteration (PEPSKit simultaneous CTMRG looks a bit similar)
- I use a Lanczos based SVD and compute relatively few singular vectors
- when computing observables, I trace over the ancilla leg while contracting bra an ket tensors
- I do not have Zygote constrains on allocating memory (but I do not aggressively try to reuse it)

In frostspin, the bottleneck is always `permutedims`. Contractions are non negligible yet always cheaper. SVD are usually negligible.

[logs_profile_julia_1stnei_genx_CTM_SU2_D11_chi122_11b43a279c_c1.log](https://github.com/user-attachments/files/20489457/logs_profile_julia_1stnei_genx_CTM_SU2_D11_chi122_11b43a279c_c1.log)
[logs_profile_1stnei_genx_CTM_SU2_D11_chi121_rdm2nd_c7cb4eea37_c1.log](https://github.com/user-attachments/files/20489458/logs_profile_1stnei_genx_CTM_SU2_D11_chi121_rdm2nd_c7cb4eea37_c1.log)

[not finished editing]

Yue-Zhengyuan:
[...]

PR #237, opened 2025-07-26
[Perf] reuse `renormalize_west_edge` for all directions

### LINKED ISSUE #220: Projectors performance improvements
Similar to the changes from #214, the projector contractions of the CTMRG algorithms could probably be improved to hardcode optimal contraction and permutations.

ogauthe:
This PR rewrites `renormalize_north_edge`, `renormalize_east_edge` and `renormalize_south_edge` to have them reuse `renormalize_west_edge`. `renormalize_west_edge` was rewritten in #229 to optimize the permutations. This PR fixes #220, part of #205.

- the formatter made changes in the tensor diagrams. I can undo them if you prefer minimizing the diff
- I added a test for a 4x4 unit cell with each bond different

### DIFF
--- a/src/algorithms/contractions/ctmrg_contractions.jl
+++ b/src/algorithms/contractions/ctmrg_contractions.jl
@@ -1168,7 +1168,7 @@ end
 function renormalize_north_edge(
     E_north::CTMRG_PEPS_EdgeTensor, P_left, P_right, A::PEPSSandwich
 )
-    A_west = map(t -> permute(t, ((1,), (3, 4, 5, 2))), A)
+    A_west = (permute(ket(A), ((1,), (3, 4, 5, 2))), permute(bra(A), ((1,), (3, 4, 5, 2))))
     return renormalize_west_edge(E_north, P_left, P_right, A_west)
 end
 function renormalize_north_edge(E_north::CTMRG_PF_EdgeTensor, P_left, P_right, A::PFTensor)
@@ -1206,7 +1206,7 @@ end
 function renormalize_east_edge(
     E_east::CTMRG_PEPS_EdgeTensor, P_bottom, P_top, A::PEPSSandwich
 )
-    A_west = map(t -> permute(t, ((1,), (4, 5, 2, 3))), A)
+    A_west = (permute(ket(A), ((1,), (4, 5, 2, 3))), permute(bra(A), ((1,), (4, 5, 2, 3))))
     return renormalize_west_edge(E_east, P_bottom, P_top, A_west)
 end
 function renormalize_east_edge(E_east::CTMRG_PF_EdgeTensor, P_bottom, P_top, A::PFTensor)
@@ -1243,7 +1243,7 @@ end
 function renormalize_south_edge(
     E_south::CTMRG_PEPS_EdgeTensor, P_left, P_right, A::PEPSSandwich
 )
-    A_west = map(t -> permute(t, ((1,), (5, 2, 3, 4))), A)
+    A_west = (permute(ket(A), ((1,), (5, 2, 3, 4))), permute(bra(A), ((1,), (5, 2, 3, 4))))
     return renormalize_west_edge(E_south, P_left, P_right, A_west)
 end
 function renormalize_south_edge(E_south::CTMRG_PF_EdgeTensor, P_left, P_right, A::PFTensor)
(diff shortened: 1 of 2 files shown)


src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -1168,13 +1168,8 @@ end
...
     E_north::CTMRG_PEPS_EdgeTensor, P_left, P_right, A::PEPSSandwich
 )
-    out = @autoopt @tensor edge[χ_W D_Sab D_Sbe; χ_E] :=
-        E_north[χ1 D1 D2; χ2] *
-        ket(A)[d; D1 D3 D_Sab D5] *
-        conj(bra(A)[d; D2 D4 D_Sbe D6]) *
-        P_left[χ2 D3 D4; χ_E] *
-        P_right[χ_W; χ1 D5 D6]
-    return out
+    A_west = (permute(ket(A), ((1,), (3, 4, 5, 2))), permute(bra(A), ((1,), (3, 4, 5, 2))))
```
lkdvos: We have functions for this (hidden) somewhere in the package:
https://github.com/QuantumKitHub/PEPSKit.jl/blob/fdc9fc5f7fa0e76bdc1fe21193a1eef17f62a2c7/src/networks/local_sandwich.jl#L38-L40

I'd have to double check, but I think this is:
```suggestion
    A_west = _rotl90_localsandwich(A)
```
As a result, this function would be agnostic to the type of sandwich as well in the sense that it also works for partition functions and pepos etc.

(The same comment is true for the other directions).

test/ctmrg/unitcell.jl
```diff
@@ -113,4 +113,47 @@ end
...
+    ]
+
+    Nspaces = [
+        vspaces[10] vspaces[4] vspaces[7] vspaces[32]
+        vspaces[2] vspaces[6] vspaces[9] vspaces[11]
+        vspaces[13] vspaces[16] vspaces[18] vspaces[19]
+        vspaces[21] vspaces[24] vspaces[26] vspaces[27]
+    ]
+    Pspaces = map(_ -> phys_space, Iterators.product(1:4, 1:4))
+    chis = map(_ -> corner_space, Iterators.product(1:4, 1:4))
```
lkdvos:
```suggestion
    Pspaces = fill(phys_space, (4, 4))
    chis = fill(corner_space, (4, 4))
```
Minor formatting change, not super important

ogauthe:
> Is there any chance you could run a small benchmark comparing the performance of the different directions now, so we can easily conclude that the different directions have more or less the same performance?

for SU(2) `D=11`, median time using https://github.com/ogauthe/BenchmarkPEPS/blob/main/run/benchmark_ctmrg.jl
`renormalize_north_edge`: 190.0 ms
`renormalize_east_edge`: 192.2ms
`renormalize_south_edge`: 191.6 ms
`renormalize_west_edge`: 189.4 ms
it was 295 ms before this PR

lkdvos:
with the expected compile-time benefits, that's definitely sufficient for me.

I'll try and investigate why the docs build is failing, I'm not sure if this can be related to this PR

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -1168,13 +1168,8 @@ end
 function renormalize_north_edge(
     E_north::CTMRG_PEPS_EdgeTensor, P_left, P_right, A::PEPSSandwich
```
lkdvos: I think with this new implementation we can actually remove the type restrictions as well, and have this as a catch-all fallback. The idea is that this does not access any of the internals of the arguments, nor have any special requirements, so I think this is a good definition. (also for the other ones).
It might additionally be worth it to check if the other functions (for PEPO, PF etc) can benefit from the same treatment, simply by deleting them and letting these functions catch that.
```suggestion
    E_north, P_left, P_right, A
```

ogauthe:
I remove the partition function methods. The 2D Ising partition function example works, so I guess legs ordering is fine.

lkdvos (approved):
To me this looks like a good change: it improves overall performance while reducing the amount of code.
@leburgel or @pbrehmer , if there are no further opinions on this I will go ahead and merge this end of tomorrow, if the tests pass.

PR #239, opened 2025-08-04
Remove InfiniteWeightPEPS and mirror_antidiag

Yue-Zhengyuan:
This PR is an overhaul of simple update, which removes `InfiniteWeightPEPS` and `mirror_antidiag`. The related tests and examples are updated accordingly. I apologize for not planning carefully before and introducing them in the first place. 

- Simple update now works with `InfinitePEPS` and `SUWeight`. When a `wpeps::InfiniteWeightPEPS` was needed before, it is now replaced by `peps::InfinitePEPS, env::SUWeight`. The tensors in the InfinitePEPS correspond to InfiniteWeightPEPS.vertices with square root of bond weights always absorbed. Then `InfiniteWeightPEPS` is no longer useful, and we can get rid of its weird behaviors under lattice rotations. 

- `env::SUWeight` is regarded as the environment for simple update, and hence the file defining it is moved to "environments/suweight.jl". Sometimes when CTMRG is also involved, the variable name is `wts::SUWeight`, with `env` reserved for `CTMRGEnv`. 

- Axis order of `SUWeight` is no longer always (west, east) and (south, north), depending on the "dual-ness" of the Espaces and Nspaces of the PEPS. E.g. for x-weights, it is either 1 ← x ← 2 or 2 → x → 1. This makes logic in `absorb_weight` simpler.

- I added a `_su_ybond!` to directly update vertical bonds for the 2-site SU, without the need to mirror or rotate the PEPS. Hence `mirror_antidiag` is no longer needed and removed. 

- I added some tricks in the half-filling Hubbard model SU example to improve convergence of SU and CTMRG (initialize with a small D = 4, then ramp up to a large D = 12, and finally go back to D = 8). 

## Things deliberately left for future PRs

- Rename `su_iter`, `simpleupdate` to `timestep!` and `time_evolve` to match MPSKit. 
- Remember normalization used in each SU iteration. (I hope someone with a concrete use case can do it instead of myself.)


src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -19,111 +19,141 @@ end
...
+    normalize!(A, Inf)
+    normalize!(B, Inf)
+    normalize!(s, Inf)
+    peps.A[row, col], peps.A[rm1, col] = A, B
+    env.data[2, row, col] = s
+    return ϵ
+end
+
+"""
+    su_iter(gate::LocalOperator, peps::InfinitePEPS, env::SUWeight, alg::SimpleUpdate; bipartite::Bool=false)
```
lkdvos: How would you feel about slightly changing the order of the arguments:

```julia
su_iter(peps, gate, alg, env; kwargs...)
```

This is slightly more in sync with our usual: `state, operator, alg, environments`

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -177,7 +207,8 @@ Perform a simple update on the infinite PEPS (`peps`) using the Hamiltonian `ham
 - The 3-site simple update algorithm is incompatible with a bipartite PEPS. Using `bipartite = true` with either `force_3site = true` or a `ham` with next-nearest neighbor terms is not allowed. 
 """
 function simpleupdate(
```
lkdvos, line 203: Same remark about argument order here

lkdvos (commented):
Looks like an overall improvement, I like this.
I will try and make some more time to go over this, are you okay with me also pushing some small changes to this branch?

As a sidenote, this is definitely a breaking change, so shall I try and register a new patch version before we try to merge this?

Yue-Zhengyuan:
@lkdvos Sure, please go ahead to make necessary changes. 

We can register v0.6.1 for the current main branch. After this SU overhaul we may go to v0.7.

src/environments/suweight.jl
```diff
@@ -0,0 +1,355 @@
...
+        pow,
+    )
+    axp1 = ax + 1
+    t_idx = collect((n - 1 == ax) ? 1 : -n for n in 1:5)
+    wt_idx = if isdual(space(t, ax + 1))
+        [1, -axp1] # t ← wt
+    else
+        [-axp1, 1] # t → wt
+    end
+    return permute(ncon((t, wt), (t_idx, wt_idx)), ((1,), Tuple(2:5)))
```
Yue-Zhengyuan, line 238: I remember being told that `ncon` is not quite recommended since it is not type stable. Here because the `@tensor` indices depends on the axis which absorbs the weight, I tried to use `MPSKit.tensorexpr`, but did not quite get how macroexpand and @generated works. Can you see if this can be improved?

Yue-Zhengyuan: Another notable change of `absorb_weight`: now we *always* absorb/remove the square root of the bond weights. Should we change `SUWeight` to store the *square root* of bond weights instead?

lkdvos: I will have a look! There is however definitely also a question about whether or not this actually matters, without having a look at a profiler this might just not be necessary.
Given that all the weights are diagonal, we could also make the output type stable by doing something like this:

```julia
tdst = similar(t)
permute!(tdst, ncon((...)))
return tdst
```
While this still has some type instability in the inner parts of the function, at least that shouldn't propagate.

Alternatively, we can simply annotate the function to return something of type `typeof(t)`, which should also work.

In general, I do have this on my to-do list to go through the entire `simple_update` machinery and ensure the optimizations and memory-usage and type-stability are all as we want it, I just haven't gotten around to it just yet.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -221,12 +221,10 @@ function _proj_from_RL(
     rl = r * l
     @assert isdual(domain(rl, 1)) == isdual(codomain(rl, 1))
     u, s, vh, ϵ = tsvd!(rl; trunc)
-    sinv = PEPSKit.sdiag_pow(s, -1)
+    sinv = PEPSKit.sdiag_pow(s, -1 / 2)
```
lkdvos, line 224: Was this just a small mistake before, or did this change and am I missing something?

src/environments/suweight.jl
lkdvos: If this is okay, I'll leave the performance review for a separate PR?

src/utility/util.jl
```diff
@@ -106,11 +106,17 @@ end
...
+```
+The axis orders for `s`, `s2` are
+```
+    1 ← s ← 2   2 → s2 → 1
 ```
 """
 function flip_svd(u::AbstractTensorMap, s::DiagonalTensorMap, vh::AbstractTensorMap)
-    return flip(u, 2), DiagonalTensorMap(flip(s, (1, 2))), flip(vh, 1)
+    return flip(u, numind(u)),
+        permute(DiagonalTensorMap(flip(s, (1, 2))), ((2,), (1,))),
```
lkdvos, line 118: What was the idea behind changing the index order of `s2` here? This seems slightly unintuitive to me, but I'm probably just missing something

lkdvos (commented):
Left some small remarks, otherwise I'm happy to merge this!

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: This is intentional. Previously the bond weight `s` is not absorbed into the "projectors" 
```julia
Pa, Pb = l * vh' * sinv, sinv * u' * r
```
(in order to obtain the vertex tensors). But now they are.

src/utility/util.jl
Yue-Zhengyuan: This is to make the codomain and the domain of `s` always non-dualed spaces, which makes taking its inverse and square root easier.

src/environments/suweight.jl
Yue-Zhengyuan: Here we should decide whether `SUWeight` stores the bond weights as before, or the square root of them. For performance, saving the square root is better. But if we regard `SUWeight` as the "environment" for SU, by analogy with `CTMRGEnv`, it is the environment of the 2-layer bra-ket combination, which should be the bond weight itself. Only because of the "simplicity" of SU that 1-layer is sufficient for SU bond truncation, and we only need to absorb the square root into the PEPS tensors before truncation. 

So theoretically I prefer `SUWeight` still being the bond weights, at the cost of a slight overhead in the algorithm.

Other performance improvements can be in another PR.

lkdvos: I think I would be fine with using the square root directly, given that this is never really used for CTMRG runs anyways? It seems reasonable to me that if you provide an environment for a single layer algorithm, it is a single layer environment. Ultimately it is just a matter of properly documenting I guess?
I'm also fine with defining `sqrt` and `inv` etc on the entire `SUWeight` struct itself to easily go between the different options if that would be convenient, and simply have the `SUWeight` struct be the struct that is most convenient for the exact algorithm at that point?

src/environments/suweight.jl
```diff
@@ -0,0 +1,354 @@
...
+end
+function absorb_weight(
+        t::PEPSTensor,
+        weights::SUWeight,
+        row::Int,
+        col::Int,
+        ax::NTuple{N, Int};
+        invwt::Bool = false,
```
lkdvos: For named tuple arguments, I would prefer not to abbreviate the name. What do you think about `invert=false`, or just `invert_weights=false` instead?

src/utility/util.jl
lkdvos: Is this making it easier because of the parity matrices when you connect them, or is it really just that the implementation of inverse and square root could have better support?
I'm fine with having this in the other direction, it's just really not that intuitive to me that you no longer have `U * S * Vh, but now get `U * transpose(S) * Vh` to reconstruct the original tensor

src/environments/suweight.jl
Yue-Zhengyuan: I'll use `inv` simply, similar to `flip`, `twist` etc.

src/utility/util.jl
Yue-Zhengyuan: Indeed it's mainly to avoid the parity matrices when absorbing the weights, though at the cost that the index order of the weights are no longer always (west, east) and (south, north).

lkdvos: Ah I see, that makes sense. Can I think about this as well (intuitively, not literally) as being the consequence/equivalent of always doing the updates/svds along the direction of the arrow, rather than along the fixed direction of the lattice?

Yue-Zhengyuan: It can be mentally regarded as this, though not exactly implemented in this way. For example, the A, B tensors in `_su_xbond!` are still the left and right ones connected to the bond.

src/environments/suweight.jl
Yue-Zhengyuan: If we directly store the square root of weights, we may want to use a more suggestive name, as `SUWeight` naturally make people think it is the bond weights themselves, which is a change that I don't quite want to do...

`inv` of `SUWeight` does not make much sense outside the SU algorithm (especially if its output is still an `SUWeight`), so I'd refrain from defining it.

In addition, I think BPEnv is more like the bond weights instead of their square root? If `SUWeight` is the square root, will it be a bit cumbersome to compare these two?

lkdvos: I'm totally fine with just keeping it as is, I'm guessing the square root of a diagonal matrix isn't really the bottleneck anyways

Yue-Zhengyuan: > the square root of a diagonal matrix isn't really the bottleneck

Exactly. Then let's keep `SUWeight` unchanged.

docs/src/examples/j1j2_su/index.md
```diff
@@ -165,96 +165,96 @@ symmetry can be advantageous for obtaining lower energies.)
...
-[ Info: LBFGS: iter   72, time  423.98 s: f = -1.977721141239, ‖∇f‖ = 9.4541e-03, α = 1.00e+00, m = 20, nfg = 1
-[ Info: LBFGS: iter   73, time  429.73 s: f = -1.977775043848, ‖∇f‖ = 1.1036e-02, α = 1.00e+00, m = 20, nfg = 1
-[ Info: LBFGS: iter   74, time  435.11 s: f = -1.977818026910, ‖∇f‖ = 1.2406e-02, α = 1.00e+00, m = 20, nfg = 1
-[ Info: LBFGS: iter   75, time  440.76 s: f = -1.977848105077, ‖∇f‖ = 1.1495e-02, α = 1.00e+00, m = 20, nfg = 1
-[ Info: LBFGS: iter   76, time  446.06 s: f = -1.977873438780, ‖∇f‖ = 5.9210e-03, α = 1.00e+00, m = 20, nfg = 1
-[ Info: LBFGS: iter   77, time  451.54 s: f = -1.977888581900, ‖∇f‖ = 6.3137e-03, α = 1.00e+00, m = 20, nfg = 1
-[ Info: LBFGS: iter   78, time  456.92 s: f = -1.977922415140, ‖∇f‖ = 1.1600e-02, α = 1.00e+00, m = 20, nfg = 1
-[ Info: LBFGS: iter   79, time  462.53 s: f = -1.977946751759, ‖∇f‖ = 1.2523e-02, α = 1.00e+00, m = 20, nfg = 1
-┌ Warning: LBFGS: not converged to requested tol after 80 iterations and time 468.13 s: f = -1.977972943430, ‖∇f‖ = 7.7636e-03
+[ Info: LBFGS: iter    1, time  898.29 s: f = -1.912844704482, ‖∇f‖ = 6.1629e-01, α = 1.00e+00, m = 0, nfg = 1
```
Yue-Zhengyuan: I don't know why LBGFS is taking much longer time to run on my machine. Maybe @pbrehmer can do a double-check? I was using 2 CPU cores to run it, with OPENBLAS_NUM_THREADS = 2.

docs/src/examples/j1j2_su/index.md
```diff
@@ -55,85 +55,85 @@ Therefore, we shall gradually increase $J_2 / J_1$ from 0.1 to 0.5, each time in
...
 ````
-[ Info: Space of x-weight at [1, 1] = Rep[U₁](0=>2, 1=>1, -1=>1)
-[ Info: SU iter 1      :  dt = 1e-03,  weight diff = 4.447e-04,  time = 0.030 sec
-[ Info: Space of x-weight at [1, 1] = Rep[U₁](0=>2, 1=>1, -1=>1)
-[ Info: SU conv 3236   :  dt = 1e-03,  weight diff = 9.998e-10,  time = 97.398 sec
-[ Info: Space of x-weight at [1, 1] = Rep[U₁](0=>2, 1=>1, -1=>1)
-[ Info: SU iter 1      :  dt = 1e-04,  weight diff = 4.436e-05,  time = 0.031 sec
-[ Info: Space of x-weight at [1, 1] = Rep[U₁](0=>2, 1=>1, -1=>1)
-[ Info: SU conv 796    :  dt = 1e-04,  weight diff = 9.999e-10,  time = 24.018 sec
+[ Info: Space of x-weight at [1, 1] = Rep[TensorKitSectors.U₁](0=>2, 1=>1, -1=>1)
```
Yue-Zhengyuan, line 109: `Rep[TensorKitSectors.U₁]` should be `Rep[U₁]`.

Yue-Zhengyuan (commented):
I find some issues when regenerating the docs.

docs/src/examples/j1j2_su/index.md
lkdvos: I think this happens when `U1` is not loaded into the namespace, which confuses me a bit since it really should be exported. Perhaps this wasn't loaded into the namespace in the docs maker, but I am still confused about why that would affect this.

docs/src/examples/j1j2_su/index.md
lkdvos: It might really just be using a larger machine, or MKL with more cores?, or a different operating system having a different random seed, therefore having a different initial state?

pbrehmer: I suspect that precompilation might again be the culprit. I think OptimKit times the first step such that it includes the precompilation time of the first differentiation through CTMRG. When I was generating the docs I had lots of things already precompiled in the session I was running it in.

pbrehmer: But it's true, even with the precompilation, the timings between the respective LBFGS steps are shorter when I generated it. I was probably using more OpenBLAS cores and also the PEPSKit parallelization (and even then, as Lukas pointed out, different machines can lead to all sorts of performance differences). In any case, the final results match so there shouldn't be any issue.

Yue-Zhengyuan: I tried again with 4 cores, and the speed has improved a lot (although still a bit slower than Paul's past run). Then I think things are fine.

docs/src/examples/j1j2_su/index.md
Yue-Zhengyuan: Anyway it's not a big deal. So we dismiss it for now?

lkdvos: ok for me. In any case, it is probably not something caused by PEPSKit, but rather an upstream thing.

lkdvos (approved):

PR #240, opened 2025-08-04
Bump v0.6.1

lkdvos:
Since it has been a while since we had a new release, and we did add a bunch of optimizations and improvements, I would suggest registering a new patch release before we do any further changes.

Release notes:

- This version contains a number of performance improvements, specifically for the dominant contractions within CTMRG for PEPS. 
- We now support constructing reduced density matrices (`reduced_densitymatrix`), which also has a more efficient implementation than before. This is now used for obtaining expectation values as well.
- There now is built-in support for computing correlators along the horizontal or vertical directions (`correlator`).
- There is a new truncation mode (`SiteDependentTruncation`) to allow specifying site-dependent truncation schemes
- Several small internal changes were made to improve type stability.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -10,3 +10,4 @@ TensorKit = "07d1fe3e-3e46-537d-9eac-e9e13d0d4cec"
 
 [compat]
 Documenter = "1.0"
+DocumenterInterLinks = "~1.0"


Yue-Zhengyuan (commented):
I suggest to be more specific on where the new features are in the release notes. 

* We now support constructing reduced density matrices (`reduced_densitymatrix`), which also have a more efficient implementation than before, and is now used for obtaining expectation values as well.
* There now is built-in support for computing correlators along the horizontal or vertical directions (`correlator`).
* There is a new truncation mode (`SiteDependentTruncation`) to allow specifying site-dependent truncation schemes.

I'm still a bit uncomfortable about normalization in SU in this version, and it is affecting some of my works. Other than this, 0.6.1 is OK for me.

lkdvos:
Linking this here: https://github.com/JuliaDocs/DocumenterInterLinks.jl/issues/18

lkdvos:
Ok, should be good to go once tests pass, and one person approves the PR. Feel free to immediately tag a release as well, if not I will probably do this tomorrow.

leburgel (approved):

PR #245, opened 2025-08-15
[perf] avoid `@autoopt` for partition function

ogauthe:
This PR is a follow-up to #229 and #237. It replaces `@autoopt` by explicit contraction scheme in CTMRG partition function contractions. I assumed the optimal permutation was the same as for a wavefunction and reproduced the same order.

I do not really understand which constraints are imposed by planar non-braiding categories, so I may be doing illegal permutations. I do not know how to check for these, I am happy to learn.

### DIFF
--- a/src/algorithms/contractions/ctmrg_contractions.jl
+++ b/src/algorithms/contractions/ctmrg_contractions.jl
@@ -96,9 +96,9 @@ function enlarge_northeast_corner(
         E_east::CTMRG_PF_EdgeTensor, A::PFTensor,
     )
     return @tensor begin
-        EC[χ_W DN; χ2] := E_north[χ_W DN; χ1] * C_northeast[χ1; χ2]
-        ECE[χ_W χ_S; DN DE] := EC[χ_W DN; χ2] * E_east[χ2 DE; χ_S]
-        corner[χ_W D_W; χ_S D_S] := ECE[χ_W χ_S; DN DE] * A[D_W D_S; DN DE]
+        EC[DN χ_W; χ2] := E_north[χ_W DN; χ1] * C_northeast[χ1; χ2]
+        ECE[DN DE; χ_S χ_W] := EC[DN χ_W; χ2] * E_east[χ2 DE; χ_S]
+        corner[χ_W D_W; χ_S D_S] :=  A[D_W D_S; DN DE] * ECE[DN DE; χ_S χ_W]
     end
 end
 
--- a/src/algorithms/contractions/ctmrg_contractions.jl
+++ b/src/algorithms/contractions/ctmrg_contractions.jl
@@ -98,7 +98,7 @@ function enlarge_northeast_corner(
     return @tensor begin
         EC[DN χ_W; χ2] := E_north[χ_W DN; χ1] * C_northeast[χ1; χ2]
         ECE[DN DE; χ_S χ_W] := EC[DN χ_W; χ2] * E_east[χ2 DE; χ_S]
-        corner[χ_W D_W; χ_S D_S] :=  A[D_W D_S; DN DE] * ECE[DN DE; χ_S χ_W]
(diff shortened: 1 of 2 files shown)


lkdvos (approved):
Looks good to me! Any chance you have some timing results too? 

Considering the planar and braiding things, happy to explain but this isn't super relevant here because they require @planar anyways so we don't really support that in PEPSKit.jl right now. In principle we could do this for the partition functions, but until someone actually needs it I'm fine with just keeping the @tensor. 

(Combining planarity and efficiency would actually be kind of a nightmare: for braided categories at least we could replace the permutations with braidings and their inverses, but for non-braided ones we cannot arbitrarily choose the intermediary permutations so we're more limited in what can be done)

src/algorithms/contractions/ctmrg_contractions.jl
```diff
@@ -54,8 +54,11 @@ function enlarge_northwest_corner(
         E_west::CTMRG_PF_EdgeTensor, C_northwest::CTMRGCornerTensor,
         E_north::CTMRG_PF_EdgeTensor, A::PFTensor,
     )
-    return @autoopt @tensor corner[χ_S D_S; χ_E D_E] :=
-        E_west[χ_S D1; χ1] * C_northwest[χ1; χ2] * E_north[χ2 D2; χ_E] * A[D1 D_S; D2 D_E]
+    return @tensor begin
+        EC[χ_S DW; χ2] := E_west[χ_S DW; χ1] * C_northwest[χ1; χ2]
+        ECE[χ_S χ_E; DW DN] := EC[χ_S DW; χ2] * E_north[χ2 DN; χ_E]
+        corner[χ_S D_S; χ_E D_E] := ECE[χ_S χ_E; DW DN] * A[DW D_S; DN D_E]
```
lkdvos, line 60: I'm not sure I fully understand your comment about never having to permute `A`. For example, here you in any case first have to permute `A[DW DN; D_S D_E]` or vice-versa to contract (we are doing TTGT everywhere), so there would still be a permutation, although hopefully keeping the `DW` leg in place should make that rather contiguous.

I don't think this can be avoided in any case, the only permutation-free directions would have to be SW or NE I think.

ogauthe: Indeed in `enlarge_XXX_corner`, it is not possible to avoid permuting `A` for directions northwest and southeast. But it is possible for `renormalize_XXX_edge`.

lkdvos: I'm sorry, I shouldn't be trying to review PRs that early in the morning, you clearly said that you were talking about the other contractions...

lkdvos:
Thanks a lot for the detailed benchmark! I do have to admit that the results are somewhat surprising to me. Am I reading that wrong or are there actual regressions by making this change too? I think it is indeed expected that the speedup isn't uniform over all the directions, but I would have guessed that we should be able to have an improvement overall, and seemingly right now it is only sometimes true. I realize also that the cases vary quite wildly, since `D` scales differently depending on whether or not it is a squashed quantum case or a partition function, but I expected them both to result in the same outcome.

In particular, what surprised me is some of the choices of putting a `D` leg first to keep a permutation contiguous, rather than a `chi` leg. While I agree that `D` might become as large as `chi` in your typical usecases, I think in all regimes we still expect `D < chi`, so that might be slightly suboptimal? In a similar line of thinking, permuting `A` should be less costly than permuting two edges that are connected, so maybe the focus might be not entirely fair. (correct me if I'm wrong though, this is really something I would have to measure and profile to see which of the steps is actually where the time is spent)

lkdvos:
@leburgel it seems like something did actually go wrong with the last changes, as now the example tests seem to time out. Any idea what could be the cause?

lkdvos:
Ok, in that case I'll merge this since that is unrelated.

PR #246, opened 2025-08-21
Change default fixed-point gradient algorithm to `:eigsolver`

leburgel:
Changes the default fixed-point gradient algorithm to `:eigsolver`, and adds a fallback linsolve in case the eigsolver approach fails.

### DIFF
--- a/src/operators/models.jl
+++ b/src/operators/models.jl
@@ -169,7 +169,7 @@ function j1_j2_model(
         T::Type{<:Number}, S::Type{<:Sector}, lattice::InfiniteSquare;
         J1 = 1.0, J2 = 1.0, spin = 1 // 2, sublattice = true,
     )
-    term_AA = S_xx(T, S; spin) + S_yy(T, S; spin) + S_zz(T, S; spin)
+    term_AA = S_exchange(T, S; spin)
     term_AB = if sublattice
         -S_xx(T, S; spin) + S_yy(T, S; spin) - S_zz(T, S; spin)  # Apply sublattice rotation
     else


src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -391,9 +391,16 @@ function fpgrad(∂F∂x, ∂f∂x, ∂f∂A, x₀, alg::EigSolver)
...
     end
     if norm(vecs[1][2]) < 1.0e-2 * alg.solver_alg.tol
-        @warn "Fixed-point gradient computation using Arnoldi failed: auxiliary component should be finite but was $(vecs[1][2]). Possibly the Jacobian does not have a unique eigenvalue 1."
+        @warn "Fixed-point gradient computation using Arnoldi failed:\n\tauxiliary component should be finite but was $(vecs[1][2])\n\tpossibly the Jacobian does not have a unique eigenvalue 1"
+        @info "Falling back to linear solver for fixed-point gradient computation."
+        backup_ls_alg = GMRES(; tol = alg.solver_alg.tol, maxiter = alg.solver_alg.maxiter, verbosity = alg.solver_alg.verbosity)
+        y, info = reallinsolve(∂f∂x, ∂F∂x, x₀, backup_ls_alg, 1, -1)
+        if alg.solver_alg.verbosity > 0 && info.converged != 1
+            @warn("backup gradient fixed-point iteration reached maximal number of iterations:", info)
+        end
```
lkdvos: How do you feel about making a backup alg which is `LinSolver`, and just returning `fpgrad(args..., alg::LinSolver)`? It removes a tiny bit of code, but it might also be slightly more readable.

pbrehmer: I agree with Lukas suggestion but perhaps we could just use the existing default algorithm machinery? That way we make sure that we consistently use defaults in places where `EigSolver` and `LinSolver` don't share common parameters.

```suggestion
        @warn "Fixed-point gradient computation using Arnoldi failed:\n\tauxiliary component should be finite but was $(vecs[1][2])\n\tpossibly the Jacobian does not have a unique eigenvalue 1"
        @info "Falling back to linear solver for fixed-point gradient computation."
        backup_ls_alg = _alg_or_nt(GradMode, (; alg = :linsolver, tol = alg.solver_alg.tol, maxiter = alg.solver_alg.maxiter, verbosity = alg.solver_alg.verbosity, iterscheme = iterscheme(alg))
        return fpgrad(∂F∂x, ∂f∂x, ∂f∂A, x₀, backup_ls_alg)
```

Other than that I like this a lot, thanks!

PR #249, opened 2025-08-25
Fix gauge fixing in `:fixed` mode for non-uniform unit cells from full SVD

pbrehmer:
I noticed that when using `:fixed` mode differentiation the gauge fixing routine would error for non-uniform unit cells when using a full SVD (as opposed a `IterSVD`). This came down the fact that the gauge signs on the full space were initialized with the wrong spaces. Now we explicitly initialize the correct spaces which is also tested in the `test/ctmrg/unitcell.jl` tests.

### DIFF
--- a/src/algorithms/optimization/fixed_point_differentiation.jl
+++ b/src/algorithms/optimization/fixed_point_differentiation.jl
@@ -281,7 +281,7 @@ function _fix_svd_algorithm(alg::SVDAdjoint, signs, info)
     # embed gauge signs in larger space to fix gauge of full U and V on truncated subspace
     signs_full = map(Iterators.product(1:4, 1:size(signs, 2), 1:size(signs, 3))) do (dir, r, c)
         σ = signs[dir, r, c]
-        space = if dir == NORTH # take unit cell interdependency of signs into account
+        extended_space = if dir == NORTH # take unit cell interdependency of signs into account
             domain(info.U_full[dir, r, _prev(c, end)]) ← codomain(info.V_full[dir, r, _prev(c, end)])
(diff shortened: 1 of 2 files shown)


src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -279,8 +280,18 @@ end
 
 function _fix_svd_algorithm(alg::SVDAdjoint, signs, info)
     # embed gauge signs in larger space to fix gauge of full U and V on truncated subspace
-    signs_full = map(zip(signs, info.S_full)) do (σ, S_full)
-        extended_σ = zeros(scalartype(σ), space(S_full))
+    signs_full = map(Iterators.product(1:4, 1:size(signs, 2), 1:size(signs, 3))) do (dir, r, c)
+        σ = signs[dir, r, c]
+        extended_space = if dir == NORTH # take unit cell interdependency of signs into account
```
lkdvos: This might read slightly easier if you do something like 
```julia
r_ext, c_ext = if ...

else
...
end
extended_space = domain(info.U_full[dir, r_ext, c_ext]) \leftarrow codomain(info.V_full[dir, r_ext, c_ext])
```

lkdvos (approved):
Suggested a slight reorganization of the code, otherwise happy to merge once tests pass (also happy to merge like this)

src/algorithms/optimization/fixed_point_differentiation.jl
pbrehmer: That does improve readability, so I added that suggestion :-)

Issue #247, opened 2025-08-25
[Discussion] Finite temperature imaginary time evolution


Yue-Zhengyuan:
I plan to add the feature to calculate ρ = exp(-βH) via simple/full update. The idea is that we start from a product state with an "ancilla" space attached to each PEPS tensor:
```math
|\psi(0)\rangle = \prod_j \sum_s {|s^{phy}_j, s^{anc}_j\rangle}
```
Then we perform imaginary time evolution on the physical space only (i.e. we apply identity gates to the ancilla space) 
```math
|\psi(\beta)\rangle = \exp(-\frac{1}{2}\beta H) {|\psi(0)\rangle}
```
Finally, exp(-βH) is obtained by tracing out the ancilla
```math
\rho(\beta) = \exp(-\beta H) = \mathsf{tr}_{anc} {|\psi(\beta)\rangle} {\langle\psi(\beta)|}
```
The issue is whether it is worthwhile to keep the physical and the ancilla spaces separated during the evolution, which will need a new struct similar to InfinitePEPS but with more than one codomain spaces, and duplicate lots of code.

A compromised version is we only split the physical and the ancilla spaces when applying the gate. Then they are fused for the rest of an SU/FU step. But from what I know, TensorKit does not provide a "efficient" way to split/fuse indices of a tensor (without contraction with a fuser tensor). So maybe it won't affect performance too much if we just always keep the physical and the ancilla spaces fused all the time (this will require the ancilla space to be fused into the gates, LocalOperators etc)?

Yue-Zhengyuan:
@sanderdemeyer I also wonder how SU/FU complement the cluster expansion method you are developing.

One thing I noticed in your and @ogauthe's code is that you choose the ancilla space to be the *dual* of the physical space. But intuitively, the purification idea should lead to ancilla space = physical space, without the need to take the dual?

lkdvos:
I don't think these approaches are actually all that different, and they just constitute different algorithms for the exact same thing. The starting point is mostly just a density matrix $\ket{\psi}\bra{\psi}$, typically the infinite temperature one where this is effectively the identity operator.
We can represent this object as a PEPO (although the infinite temperature one is a "product PEPO"), or a PEPS with two physical legs. These representations are equivalent, and we can go between the two representations at any point, simply by bending the PEPO leg down, or the PEPS leg up. (See also how mapping an MPO to an MPS with two physical legs has that same idea).

[...]

PR #250, opened 2025-08-29
Simple update for PEPO

Yue-Zhengyuan:
This PR adds the ability of simple update to evolve a PEPO. The most well-known application is calculating exp(-βH) at finite temperature. 

- Overloaded `MPSKit.infinite_temperature_density_matrix` to creates the trivial PEPO (i.e. ρ = exp(-βH) at infinite temperature) given a Hamiltonian (as a LocalOperator).

- `simpleupdate` gets a new kwarg `gate_bothsides::Bool`, controlling how the evolution of PEPO is done:
    - `true`: (default) exp(-H dt/2) ρ exp(-H dt/2). This is also how TeNeS does it.
    - `false`: exp(-H dt) ρ (corresponding to the purified approach, without fusing the physical and the ancilla spaces)
    The total evolution time β is always `maxiter * alg.dt`. In the former (default) approach the time slice for each Trotter gate is actually `dt / 2`. This kwarg has no effect for PEPS evolution.

- The kwarg `bipartite` does not apply to finite temperature evolution. 

### Things to do in follow-up PRs

- [ ] Implement the 3-site version to allow Hamiltonians with next-nearest neighbor terms (e.g. J1-J2).
- [x] Add `reduced_densitymatrix` and `expectation_value` support for calculating tr(ρΟ) / tr(ρ). (Done in #252 and #254)


Yue-Zhengyuan:
Here are benchmark results on the transverse field Ising model (J = 1 and g = 2), using PEPO D = 8 and CTMRG χ = 16, and evolution time step dt = 0.001. The non-purified approach smears out the transition compared to HOTRG, and produce a smaller x-magnetization in the high-T phase, so it is indeed bad as mentioned by @sanderdemeyer in #247. But the smearing out is much reduced after purification. 

Screenshot 2025-08-29 at 17 04 49
Screenshot 2025-08-29 at 17 04 59

I still need to compare with TeNeS (which also uses simple update) to see if they also suffer.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -28,24 +28,24 @@ Simple update of the x-bond between `[r,c]` and `[r,c+1]`.
...
-        gate::AbstractTensorMap{T, S, 2, 2},
-        env::SUWeight,
-        row::Int,
-        col::Int,
-        trscheme::TruncationScheme,
-    ) where {T <: Number, S <: ElementarySpace}
-    Nr, Nc = size(peps)
+        state::P, gate::AbstractTensorMap{T, S, 2, 2}, env::SUWeight,
+        row::Int, col::Int, trscheme::TruncationScheme; gate_ax::Int = 1
+    ) where {P <: InfiniteState, T <: Number, S <: ElementarySpace}
```
lkdvos:
```suggestion
        state::InfiniteState, gate::AbstractTensorMap{T, S, 2, 2}, env::SUWeight,
        row::Int, col::Int, trscheme::TruncationScheme; gate_ax::Int = 1
    ) where {T <: Number, S <: ElementarySpace}
```
This doesn't really need a where clause, and I think this reads slightly better

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -28,24 +28,24 @@ Simple update of the x-bond between `[r,c]` and `[r,c+1]`.
...
+    Nr, Nc = size(state)[1:2]
     @assert 1 <= row <= Nr && 1 <= col <= Nc
     cp1 = _next(col, Nc)
     # absorb environment weights
-    A, B = peps.A[row, col], peps.A[row, cp1]
+    A, B = if state isa InfinitePEPS
+        state.A[row, col], state.A[row, cp1]
+    else
+        state.A[row, col, 1], state.A[row, cp1, 1]
```
lkdvos: This might be worth double-checking, but I think that when an array has size 1 in one of its axes, you can simply discard this, so both `InfinitePEPS` and `InfinitePEPO` can be indexed `state.A[row, col], state.A[row, cp1]`. (But then probably better to put `@assert state isa InfinitePEPS || size(state, 3) == 1` as well)

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -71,24 +75,24 @@ Simple update of the y-bond between `[r,c]` and `[r-1,c]`.
...
 function _su_ybond!(
-        peps::InfinitePEPS,
-        gate::AbstractTensorMap{T, S, 2, 2},
-        env::SUWeight,
-        row::Int,
-        col::Int,
-        trscheme::TruncationScheme,
-    ) where {T <: Number, S <: ElementarySpace}
-    Nr, Nc = size(peps)
+        state::P, gate::AbstractTensorMap{T, S, 2, 2}, env::SUWeight,
```
lkdvos: Same comment about where clause here

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -98,73 +102,113 @@ function _su_ybond!(
...
+    else
+        state.A[row, col, 1], state.A[rm1, col, 1] = A, B
+    end
     env.data[2, row, col] = s
     return ϵ
 end
 
 """
-    su_iter(peps::InfinitePEPS, gate::LocalOperator, alg::SimpleUpdate, env::SUWeight; bipartite::Bool=false)
+    su_iter(peps::P, gate::LocalOperator, alg::SimpleUpdate, env::SUWeight; kwargs...) where {P <: InfiniteState}
```
lkdvos: Given that `InfiniteState` is a non-exported constant, I'm not sure it helps to put this in the docstring like this. I think I would probably do something like:

```julia
"""
    su_iter(state::InfinitePEPS, gate::LocalOperator, alg::SimpleUpdate, env::SUWeight; bipartite=false)
    su_iter(densitymatrix::InfinitePEPO, gate::LocalOperator, alg::SimpleUpdate, env::SUWeight; biparatite=false, gate_side::Symbol = :both)
...
"""
```

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -98,73 +102,113 @@ function _su_ybond!(
...
+When the input `state` has a unit cell size of (2, 2), one can set `bipartite = true` to enforce the bipartite structure. 
 """
 function su_iter(
-        peps::InfinitePEPS,
-        gate::LocalOperator,
-        alg::SimpleUpdate,
-        env::SUWeight;
-        bipartite::Bool = false,
+        state::InfinitePEPS, gate::LocalOperator, alg::SimpleUpdate, env::SUWeight;
+        bipartite::Bool = false, gate_side::Symbol = :both
```
lkdvos: Does it make sense to have the default `gate_side` be `:both`? I might be mistaken but I think that one is the least common approach.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -98,73 +102,113 @@ function _su_ybond!(
...
+            state2.A[rm1, cm1] = deepcopy(state2.A[r, c])
+            state2.A[r, cm1] = deepcopy(state2.A[rm1, c])
             env2.data[2, rm1, cm1] = deepcopy(env2.data[2, r, c])
         end
     end
-    return peps2, env2
+    return state2, env2
+end
+
+function su_iter(
```
lkdvos: Similar comment about the indexing behavior here as well, I think these implementations can be joined that way, to reduce code duplication

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -181,42 +225,44 @@ function _simpleupdate2site(
...
         converge && break
     end
-    return peps, env, wtdiff
+    return state, env, wtdiff
 end
 
 """
-    simpleupdate(peps::InfinitePEPS, ham::LocalOperator, alg::SimpleUpdate, env::SUWeight;
+    simpleupdate(
+        state::P, ham::LocalOperator, alg::SimpleUpdate, env::SUWeight;
```
lkdvos: Same comment here about the `InfiniteState` in a docstring

src/environments/suweight.jl
```diff
@@ -125,6 +125,25 @@ function SUWeight(peps::InfinitePEPS)
...
+"""
+    SUWeight(pepo::InfinitePEPO)
+
+Create an SUWeight for a given one-layer InfinitePEPO.
+The weights are initialized as identity matrices of element type `Float64`.
+"""
+function SUWeight(pepo::InfinitePEPO)
+    @assert size(pepo, 3) == 1
+    Nspaces = map(pepo.A[:, :, 1]) do t
+        V = domain(t, NORTH)
```
lkdvos:
```suggestion
        V = north_virtualspace(t)
```

src/environments/suweight.jl
```diff
@@ -125,6 +125,25 @@ function SUWeight(peps::InfinitePEPS)
...
+        isdual(V) ? V' : V
+    end
+    Espaces = map(pepo.A[:, :, 1]) do t
+        V = domain(t, EAST)
```
lkdvos:
```suggestion
        V = east_virtualspace(t)
```

src/environments/suweight.jl
```diff
@@ -125,6 +125,25 @@ function SUWeight(peps::InfinitePEPS)
...
 
+"""
+    SUWeight(pepo::InfinitePEPO)
+
+Create an SUWeight for a given one-layer InfinitePEPO.
+The weights are initialized as identity matrices of element type `Float64`.
+"""
+function SUWeight(pepo::InfinitePEPO)
+    @assert size(pepo, 3) == 1
+    Nspaces = map(pepo.A[:, :, 1]) do t
```
lkdvos:
```suggestion
    Nspaces = map(@view(pepo.A[:, :, 1])) do t
```

Here you have to be a little bit careful because slicing arrays, even in this trivial case, allocates a new array, which isn't necessary. This is not a huge impact, but might as well avoid it here.

src/environments/suweight.jl
```diff
@@ -125,6 +125,25 @@ function SUWeight(peps::InfinitePEPS)
...
+        V = domain(t, NORTH)
+        isdual(V) ? V' : V
+    end
+    Espaces = map(pepo.A[:, :, 1]) do t
```
lkdvos:
```suggestion
    Espaces = map(@view(pepo.A[:, :, 1])) do t
```

src/environments/suweight.jl
```diff
@@ -210,11 +230,12 @@ absorb_weight(t, weights, 2, 3, 2; inv=true)
 ```
 """
 function absorb_weight(
-        t::PEPSTensor, weights::SUWeight, row::Int, col::Int, ax::Int; inv::Bool = false
-    )
+        t::PT, weights::SUWeight, row::Int, col::Int, ax::Int; inv::Bool = false
```
lkdvos: Same comment about where clause here

src/environments/suweight.jl
```diff
@@ -228,23 +249,23 @@ function absorb_weight(
...
     else
-        [-axp1, 1] # t → wt
+        [-ax′, 1] # t → wt
     end
-    return permute(ncon((t, wt), (t_idx, wt_idx)), ((1,), Tuple(2:5)))
+    return permute(ncon((t, wt), (t_idx, wt_idx)), (Tuple(1:nout), Tuple((nout + 1):ntol)))
 end
 function absorb_weight(
-        t::PEPSTensor,
+        t::PT,
```
lkdvos:
```suggestion
        t::Union{PEPSTensor,PEPOTensor},
```

src/operators/ancilla.jl
```diff
@@ -0,0 +1,34 @@
+"""
+Attach the ancilla space `Va` and merge with physical space `Vp` of
+the physical operator `op`, so the new physical space becomes `Vp ⊗ Va`.
+"""
+function attach_ancilla(
+        op::AbstractTensorMap{T, S, N, N}; Va0::Union{ElementarySpace, Nothing} = nothing
+    ) where {T, S <: ElementarySpace, N}
+    Vp = first(codomain(op))
+    Va = isnothing(Va0) ? Vp' : Va0
+    id_Va = TensorKit.id(Va)
```
lkdvos: I think you can create the identity for all spaces at the same time, to avoid the reduce

src/operators/ancilla.jl
```diff
@@ -0,0 +1,34 @@
...
+        vcat([collect(1:4N)], f_idxs),
+        vcat(fill(false, N + 1), fill(true, N))
+    )
+    return permute(op_a, (Tuple(1:N), Tuple((N + 1):2N)))
+end
+
+"""
+Attach the ancilla space to all terms in the LocalOperator `O`
+"""
+function attach_ancilla(O::LocalOperator; Va0 = nothing)
```
lkdvos: Note that this function is, if I'm not mistaken, equivalent to https://github.com/QuantumKitHub/PEPSKit.jl/blob/df7a2c582591d6e225ad459a8040335b92d641a0/src/operators/localoperator.jl#L192

src/operators/infinitepepo.jl
```diff
@@ -123,6 +123,21 @@ function initializePEPS(
     return InfinitePEPS(Pspaces, Nspaces, Espaces)
 end
 
+"""
+    trivial_InfinitePEPO([T::Type{<:Number},] Vp::ElementarySpace, unitcell::NTuple{3, Int})
+
+Create the trivial InfinitePEPO with physical space `Vp ⊗ Vp'` and virtual bond dimension 1,
+which is the product of identity operators on `Vp`.
+"""
+function trivial_InfinitePEPO(T::Type{<:Number}, Vp::ElementarySpace, unitcell::NTuple{3, Int})
```
lkdvos: It might be nice to repurpose the equivalent MPSKit construction: https://github.com/QuantumKitHub/MPSKit.jl/blob/df76b90f5d570cfae7ee2cf0a505d296f2e50e98/src/algorithms/toolbox.jl#L30

src/operators/infinitepepo.jl
```diff
@@ -221,3 +236,47 @@ function _stack_tuples(A::Matrix{NTuple{N, T}}) where {N, T}
...
+"""
+    InfinitePartitionFunction(pepo::InfinitePEPO)
+
+Creates an InfinitePartitionFunction by tracing out the physical spaces of a one-layer InfinitePEPO.
+"""
+function InfinitePartitionFunction(pepo::InfinitePEPO)
+    if size(pepo, 3) != 1
+        throw(ArgumentError("To convert a PEPO to partition function, it must have only 1 layer."))
+    end
+    A = map(pepo.A[:, :, 1]) do t
```
lkdvos:
```suggestion
    A = map(@view(pepo.A[:, :, 1])) do t
```

lkdvos (commented):
Looks great, thanks for the work!

I left some comments throughout, but there are some additional things I'm not fully convinced about that we could discuss:

I'm not super convinced by the `gate_side` keyword arguments.
It's a bit strange to me to use a symbol at the top level, and then suddenly change it to an integer a bit further down the callstack. In this case, it might make more sense to have either the symbols all the way through, or `@enum GateSide left right both` if you prefer to work with integers to begin with. There should probably also be some checking that the symbol is one of the allowed symbols to begin with, since this sometimes gets the compiler confused about return types because in principle neither of the branches could be chosen, and this is easy to have subtle type stability bugs.

Additionally, some of these functions to me feel a bit like they are better left as separate functions, rather than using a keyword to distinguish between them (especially the lower down functions). This might be a bit subjective though, feel free to disagree!

Am I correct in assuming that `:domain` and `:codomain` side could also be implemented by correctly permuting the PEPO/gate, so really there is only one of them that is relevant?
Additionally, for the `:both` option, it might be more efficient to really map it to a PEPS and do the update there, as this would require less overhead from the permutations etc? The interesting part is that this would really apply both at the same time, rather than sequentially.

Overall, this definitely looks great and I think the results are as expected from simple update results for finite temperature. It woudl definitely be nice to have the expectation values/reduced density matrices working in the near future as well, and have some tests for fermionic systems, but I'm okay with adding that in later.

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: I just checked, this is indeed allowed. In addition, the check `size(state, 3) == 1` is unnecessary, since Julia already won't allow it otherwise.
```julia
julia> a = rand(2, 2, 2);

julia> a[1, 2]
ERROR: BoundsError: attempt to access 2×2×2 Array{Float64, 3} at index [1, 2]
Stacktrace:
 [1] throw_boundserror(A::Array{Float64, 3}, I::Tuple{Int64, Int64})
   @ Base ./essentials.jl:14
 [2] checkbounds
   @ ./abstractarray.jl:699 [inlined]
 [3] getindex(::Array{Float64, 3}, ::Int64, ::Int64)
   @ Base ./array.jl:929
 [4] top-level scope
   @ REPL[14]:1
```

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: In the next fix I shall temporarily set it to `:codomain` before we finally decide on the interface.

Yue-Zhengyuan:
> Am I correct in assuming that `:domain` and `:codomain` side could also be implemented by correctly permuting the PEPO/gate, so really there is only one of them that is relevant?

Their difference lies in which physical axis of the PEPO tensor is transferred to the "reduced tensor" when performing QR/LQ in `_qr_bond` (currently specified by `gate_ax` there). So in principle `:domain` is the same as `:codomain` with a PEPO from dual spaces to dual spaces (which is something I don't feel comfortable with). What do you think if I just remove the `:domain` mode?

> Additionally, for the `:both` option, it might be more efficient to really map it to a PEPS and do the update there, as this would require less overhead from the permutations etc? The interesting part is that this would really apply both at the same time, rather than sequentially.

I did consider applying the gate on both sides at the same time. But in this way to bond dimension truncation will be from $d^2 D$ (instead of $d D$) back to $D$, which use more memory and may result in a larger truncation error (though I'm not sure about the latter point). Anyway the sequential approach is just simpler to implement, and tensor permutation is never not the bottleneck of the algorithm.

src/operators/infinitepepo.jl
Yue-Zhengyuan: It will correspond to `infinite_temperature_density_matrix(H::LocalOperator)` here. However, it seems that in PEPSKit `LocalOperator` does not enforce the same scalartype of all of its terms, so I can't decide the scalartype of the produced PEPO without an additional argument `T` to the function.

lkdvos: You could promote the scalartype of all terms at runtime though?

src/operators/ancilla.jl
Yue-Zhengyuan: Now I directly use the existing `_fuse_id`, so `attach_ancilla` is no longer needed.

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: Now I changed it to `gate_bothsides::Bool`, which defaults to `true` (following TeNeS). A side effect is that `:domain` mode is removed. For InfinitePEPS I just further set it to `false` inside the SU function, which does not look very elegant. Is there a better way to make this kwarg applicable to InfinitePEPO only?

src/operators/ancilla.jl
Yue-Zhengyuan: I added a version of it without `Ps`. In this case `Ps` is set to the domain of `op` (converted to a tuple). I removed `attach_ancilla` for LocalOperators, though.

src/operators/infinitepepo.jl
Yue-Zhengyuan: I found a way to promote the datatypes. Thanks!

src/algorithms/time_evolution/evoltools.jl
```diff
@@ -1,3 +1,16 @@
+const InfiniteState = Union{InfinitePEPS, InfinitePEPO}
+
+function MPSKit.infinite_temperature_density_matrix(H::LocalOperator)
+    T = promote_type((scalartype(t.second) for t in H.terms)...)
```
lkdvos: Can we define this as `scalartype(H)` and factor out that function?

lkdvos (commented):
I think I'm okay with your proposed interface, I can't really come up with something better myself right now either, so I think this is just what it has to be

src/algorithms/time_evolution/evoltools.jl
Yue-Zhengyuan: I have moved this to `VI.scalartype(O::LocalOperator)`. 

I see that the scalartypes for other PEPSKit structs are defined as `VI.scalartype(::Type{Something})`, but I feel this is impossible for LocalOperators.

lkdvos (approved):

src/operators/localoperator.jl
```diff
@@ -144,6 +144,13 @@ end
 Base.:-(O::LocalOperator) = -1 * O
 Base.:-(O1::LocalOperator, O2::LocalOperator) = O1 + (-O2)
 
+# VectorInterface
+# ---------------
+
+function VI.scalartype(O::LocalOperator)
+    return promote_type((scalartype(t.second) for t in O.terms)...)
+end
```
lkdvos:
```suggestion
function VI.scalartype(::Type{<:LocalOperator{T}}) where {T}
    return promote_type((scalartype(last(fieldtypes(p))) for p in fieldtypes(T))...)
end
```

lkdvos (changes requested):
Can you check if the following works in the type domain?

src/operators/localoperator.jl
Yue-Zhengyuan: This works. Thanks!

PR #251, opened 2025-09-02
Actually use scheduler in reverse rules

lkdvos:
This is a small change that ensures that we are actually using the correct scheduler in the reverse rules. I don't really understand why or how this is happening, but I noticed that the current way the `rrule` was implemented seems to intercept the callstack before the default keyword gets filled in, therefore always using the default scheduler, even if set to `:serial`.


lkdvos:
I’m definitely okay with having both separately, I’ll see what I can do

PR #252, opened 2025-09-02
Add double-layer-pepo reduced densitymatrix

lkdvos:
Follow-up for #250 and #184:

I want to write reduced density matrices for `densitymatrix * densitymatrix'` sandwiches, assuming that we'll obtain the CTMRG environments via fusing the InfinitePEPO into a InfinitePEPS construction.

- [x] Write double-layer reduced densitymatrix contraction
- [ ] Write `InfinitePEPO` to fused `InfinitePEPS` converter
- [ ] Add tests

@Yue-Zhengyuan if you have test cases laying around that would be wonderful


Yue-Zhengyuan:
We can test the functions using [`test/examples/tf_ising_finiteT.jl`](https://github.com/Yue-Zhengyuan/PEPSKit.jl/blob/d1683b3cc083c942af546b209a1f195e555754da/test/examples/tf_ising_finiteT.jl) in #250; just replace the current "manual" measurements with `expectation_value`. And energy measurement can also be added. (Though fermions haven't been covered)

The `densitymatrix * densitymatrix'` case corresponding to the purified approach to get exp(-βH). But if we don't care about it being Hermitian, it is also useful to have the reduced density matrix constructed from a *single* layer of PEPO (`densitymatrix` itself). The cost for the latter approach is smaller because the PEPO bond dimension is D instead of D^2.

lkdvos:
Yeah, I would definitely like to add both but this one was easier since it was more similar to the peps case. I’ll try generalize in a follow up PR if that’s ok?

Yue-Zhengyuan:
OK. For the tests, besides the physical example of Ising model, we can just generate a random PEPO and CTMRGEnv (no need to converge the env), and check the measurement results are the same as what we would obtain if we fuse the virtual legs and the physical legs respectively of the PEPO.

src/algorithms/contractions/localoperator.jl
```diff
@@ -251,25 +303,40 @@ end
...
+""" reduced_densitymatrix
+
 function reduced_densitymatrix(
         inds::NTuple{N, CartesianIndex{2}}, ket::InfinitePEPS, bra::InfinitePEPS, env::CTMRGEnv
     ) where {N}
     static_inds = Val.(inds)
     return _contract_densitymatrix(static_inds, ket, bra, env)
 end
+function reduced_densitymatrix(
+        inds::NTuple{N, CartesianIndex{2}}, ket::InfinitePEPO, bra::InfinitePEPO, env::CTMRGEnv
```
Yue-Zhengyuan, line 327: Maybe we add some docstrings explaining why the two PEPOs are still called `ket` and `bra` from the purification point of view?

lkdvos:
That's a great point about the tests, thanks a lot, this really sped up the dev time for me.
I think this is now working and ready, as discussed I'll add the single-layer algorithm in a separate PR, to avoid having a huge PR here.

Yue-Zhengyuan:
Do you mind if I make the tests a bit more stricter (and some other minor changes)? I find it helpful not to set the dimension of all sectors the same, especially when the charge of a sector is changed after flipping / taking the dual of the space. Anyway I have checked and the tests still pass.

src/networks/tensors.jl
```diff
@@ -123,3 +123,23 @@ function physicalspace(t::PEPOTensor)
...
+    F = isomorphism(Int, fuse(codomain(O)), codomain(O))
+    return F * O, F
+end
+
+"""
+    $(SIGNATURES)
+
+Trace out the physical indices of a PEPO tensor, obtaining a partition function tensor.
+"""
+function contract_physicalspaces(O::PEPOTensor)
```
Yue-Zhengyuan: I prefer to name it `trace_physicalspaces`.

test/toolbox/densitymatrices.jl
```diff
@@ -0,0 +1,43 @@
...
+    D = Ds[I]
+    χ = χs[I]
+
+    ρ = InfinitePEPO(d, D)
+
+    ρ_peps = @constinferred InfinitePEPS(ρ)
+    env = CTMRGEnv(ρ_peps, χ)
+
+    O = rand(d, d)
+    O += O'
```
Yue-Zhengyuan: I don't think we need to make `O` Hermitian? Things should work for arbitrary `O` which does not need to be a physical observable.

test/toolbox/densitymatrices.jl
```diff
@@ -0,0 +1,43 @@
+using TensorKit
+using PEPSKit
+using Test
+using TestExtras
+
+ds = Dict(Trivial => ℂ^2, U1Irrep => U1Space(i => 1 for i in -1:1), FermionParity => Vect[FermionParity](0 => 1, 1 => 1))
+Ds = Dict(Trivial => ℂ^3, U1Irrep => U1Space(i => 2 for i in -1:1), FermionParity => Vect[FermionParity](0 => 2, 1 => 2))
+χs = Dict(Trivial => ℂ^4, U1Irrep => U1Space(i => 3 for i in -2:2), FermionParity => Vect[FermionParity](0 => 4, 1 => 4))
```
Yue-Zhengyuan: I find it helpful not to set the dimension of all sectors the same, especially when the charge of a sector is changed after flipping / taking the dual of the space. Anyway I have checked and the tests still pass.

test/toolbox/densitymatrices.jl
lkdvos: Oh you are right, this is a remnant from some of my tests with the fermions that was initially failing, and I wanted to get a real number out but that's not even true when the environments aren't converged either

PR #253, opened 2025-09-08
Fix `add_physical_charge` for fermionic operators

### LINKED ISSUE #259: Bug in `add_physical_charge` for fermionic operators
When switching the convention for `MPSKit.add_physical_charge` in #183 to take in charges that we want to insert into the domain of the local state tensors, we introduced a sneaky bug. With this change the auxiliary physical space naturally became a dual space, but this was not taken into account when applying the isomorphisms that fuse the identity on the auxiliary space into the Hamiltonian in the `MPSKit.add_physical_charge` implementation for a `PEPSKit.local_operator`.

The reason this never came up is because we include the one-site operators directly into the two-site terms in `MPSKitModels.hubbard_model`, giving two missing twists which always cancelled out.

A minimal example demonstrating the issue:

```julia
lattice = InfiniteSquare(1, 1)
U = 30.0
S = FermionParity ⊠ U1Irrep

# construct the onsite term of the Fermi-Hubbard model
H_U = U * nꜛnꜜ(ComplexF64, U1Irrep, Trivial)
spaces = fill(space(H_U, 1), (lattice.Nrows, lattice.Ncols))

# wrap it in a trivial LocalOperator
H = LocalOperator(spaces, ((1, 1),) => H_U)

# store trace as characteristic of the operator
tr_before = tr(last(only(H.terms)))

# shift physical charge
caux = S((1, 1))
H_shifted = add_physical_charge(H, fill(caux, size(H.lattice)...))

# recompute the trace and see if it's the same
tr_after = tr(last(only(H_shifted.terms)))

# missing twist causes implicit minus sign
@show abs(tr_before - tr_after) / abs(tr_before)
```
```
abs(tr_before - tr_after) / abs(tr_before) = 2.0
```

leburgel:
Fixes #259.

Instead of applying a `flip` to the auxiliary space to get rid of the dual (which would also be correct I think), I decided to change the underlying contraction such that it no longer makes implicit assumptions on the auxiliary space.

### DIFF
--- a/src/operators/localoperator.jl
+++ b/src/operators/localoperator.jl
@@ -176,10 +176,10 @@ TensorKit.spacetype(::Type{T}) where {S, T <: LocalOperator{<:Any, S}} = S
     end
     f_dag_es = map(1:N) do i
         j = 3 * (i - 1) + 1
-        return tensorexpr(:(fs[$i]), -(N + i), (j + 1, j + 2))
+        return tensorexpr(:(twistdual(fs[$i]', 1:2)), (j + 1, j + 2), -(N + i))
     end
     multiplication_ex = Expr(
-        :call, :*, op_e, f_es..., map(x -> Expr(:call, :conj, x), f_dag_es)...
+        :call, :*, op_e, f_es..., f_dag_es...
     )
     return macroexpand(@__MODULE__, :(return @tensor $op_out_e := $multiplication_ex))
 end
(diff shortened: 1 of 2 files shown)


lkdvos (dismissed):
Do we have some kind of test coverage for this, or any way to actually check that things keep working properly in the meantime?
I'm definitely happy to keep things like this if that is just too much work for now, just wanted to ask.

PR #254, opened 2025-09-09
Add single-layer-pepo reduced densitymatrix

lkdvos:


src/algorithms/contractions/localoperator.jl
```diff
@@ -230,7 +235,7 @@ end
 @generated function _contract_local_operator(
         inds::NTuple{N, Val},
         O::AbstractTensorMap{T, S, N, N},
-        ket::InfinitePEPS, bra::InfinitePEPS,
+        state::Tuple{InfinitePEPS, InfinitePEPS},
```
Yue-Zhengyuan, line 236: After this change, L36 should be `_contract_local_operator(static_inds, O, (ket, bra), env)`, which hasn't been changed yet.

Moreover, I find that `contract_local_operator` and `contract_local_norm` are used *nowhere*, both within the package or in any of the tests or examples. This may be why the tests didn't catch the above issue. I think these two they are just remnants of the old implementation of `expectation_value`. Besides, for partition functions we have `contract_local_tensor` which has nothing to do with these two either. Since they were never exported, we don't need to worry about backward compatibility either. Can we simply remove them, because they appear to be already replaced by newer functions utilizing the RDMs? 

BTW, are there any way to automatically detect unused code in Julia?

src/algorithms/contractions/localoperator.jl
```diff
@@ -171,35 +159,52 @@ function _contract_state_expr(rowrange, colrange, cartesian_inds = nothing)
...
+                    physicallabel(:O, side, inds_id)
+                end
+                tensor_name = if side == 2
+                    :(state[2][mod1($(rmin + i - 1), end), mod1($(cmin + j - 1), end)])
+                else
+                    :(twistdual(state[1][mod1($(rmin + i - 1), end), mod1($(cmin + j - 1), end)], (1, 2)))
+                end
             else
-                physicallabel(:O, side, inds_id)
+                error("tba")
```
Yue-Zhengyuan: Maybe it's better to clearly specify which is unimplemented, so other people know what's left out, like 
```julia
error("Contraction with PEPO with more than 2 layers is unimplemented.")
```

src/algorithms/contractions/localoperator.jl
```diff
@@ -534,29 +551,43 @@ end
 end
 
 @generated function _contract_densitymatrix(
-        inds::NTuple{N, Val}, ket::InfinitePEPO, bra::InfinitePEPO, env::CTMRGEnv
-    ) where {N}
+        inds::NTuple{N, Val}, state::Tuple{InfinitePEPO, Vararg{InfinitePEPO, M}}, env::CTMRGEnv
```
Yue-Zhengyuan, line 552: Is `Varargs` the Julian way to accept a variable number of InfinitePEPOs in the tuple `state`? 

I prefer we only accept one InfinitePEPO. Then to calculate tr(O ρ1 ρ2...), we simply ask users to combine ρ1, ρ2, ... into one single PEPO. I think this can help avoid causing confusion between `tr(O ρ1 ρ2)` (where `ρ1 ρ2` is treated as a single mixed state) and `<ρ1 | O | ρ2>` (where `ρ1`, `ρ2` are actually two separate pure states, just with an ancilla leg on each site). We should also make this clear in the docstring.

To stack InfinitePEPOs together, we can overload `Base.:*` which simply does the stacking, without fusing virtual legs or truncating them. But if `Base.:*` give people the impression that it should be commutative, we can define something else.

lkdvos: I agree with the sentiment, though this was more a programming trick to obtain whether or not it is 1 or 2 pepos, since they are handled differently (the 2-option explicitly conjugates the bra). I didn't want to copy the entire function for both, therefore I just merged the signatures and use the value M to distinguish between both cases. I agree that we don't really want to use this for more than 2, but this function will error anyways in that case, since it's asserting the height to be 1-2. TLDR, it might look like this accepts multiple, but it's really only meant to distinguish between one or two. Since it's an internal function I believe this is fine. 

Considering the other functions, I'm not entirely sure. As @leburgel was mentioning yesterday, for larger operators (plaquette terms etc), it seems like always using the density matrices may become quite expensive, and while I would have to work out the expressions to really convince myself, I can see how for 4-body operators, explicitly multiplying in the operators while contracting could be more efficient than first constructing the reduced density matrix.

src/algorithms/contractions/localoperator.jl
Yue-Zhengyuan: @lkdvos If we are keeping these two functions (so that one day we may automatically switch between the two ways to calculate `expectation_value` according to the cost), we should add a test (using random states and envs) to check that they agree.

src/algorithms/contractions/localoperator.jl
Yue-Zhengyuan: If we don't intend to support PEPO with more than 2 layers, maybe the error message shouldn't be "TBA"? 😁

lkdvos: Very good point 😀 To be completely honest, I definitely didn't think it through too much, and I do agree that defining `Base.*` for PEPOs would be reasonable, but for this implementation I'm also assuming that everything is single layer anyways so perhaps I should really just rethink the names of everything...

The main idea is to distinguish between two cases: the `tr(rho*O)` one and the `dot(rho, O, rho)` (with conjugation) one. This is what I'm calling height here. Separetely, there is the height of the individual PEPOs themselves, which for now we are also just assuming to be 1. I'm not so sure if there is that much use for allowing more layers anyways, mostly since the computational costs really grow quickly then, but I don't want to explicitly exclude it.

Yue-Zhengyuan:
I've added tests on `contract_local_operator` and `contract_local_norm`, and fixed some space mismatch errors caused by wrong generated `@tensor` indices.

lkdvos (commented):
Thanks for putting that together and fixing this. I don't have too much time this week, so if you want to change the error messages to be more informative and merge this, I'm definitely happy with that.

Yue-Zhengyuan (dismissed):
I have made some minor changes to the docstring and error messages. The format check error seems to be an issue with GitHub. OK to merge for me now.

Yue-Zhengyuan (dismissed):
I changed `physicalspace` for PEPO to return the physical codomain space in the bottom layer directly as a Nrow x Ncol matrix (no 3rd dimension). The original implementation returns something strange (e.g. for an 2 x 2 x 1 PEPO, it returns a length-2 vector). Things should be fine now. @lkdvos You may want to take a final look before merging.

src/operators/infinitepepo.jl
```diff
@@ -153,7 +153,7 @@ domain_physicalspace(T::InfinitePEPO, r::Int, c::Int) = domain_physicalspace(T[r
 function codomain_physicalspace(T::InfinitePEPO, r::Int, c::Int)
     return codomain_physicalspace(T[r, c, end])
 end
-physicalspace(T::InfinitePEPO) = physicalspace.(Ref(T), 1:size(T, 1), 1:size(T, 2))
+physicalspace(T::InfinitePEPO) = physicalspace.(@view(unitcell(T)[:, :, end]))
```
lkdvos: I agree that the previous implementation was wrong, although I dont think this is fully right either. Previously we were verifying that the `codomain_physicalspace` of the last layer and the `domain_physicalspace` of the first layer match, now this only verifies the last layer, so if we would have two layers that alternate between `ComplexSpace(2)` and `ComplexSpace(3)`, this would now error.

```suggestion
physicalspace(T::InfinitePEPO) = [physicalspace(T, row, col) for row in axes(T, 1), col in axes(T, 2)]
```

Yue-Zhengyuan (approved):

Issue #255, opened 2025-09-10
Issue with `SimultaneousCTMRG` convergence starting from random initial environment


leburgel:
I ran into an issue when running `leading_boundary` on a $2 \times 2$ `InfinitePartitionFunction` using `SimultaneousCTMRG`.

Usually, when running `leading_boundary` at a fixed environment bond dimension $\chi$ using either `trscheme = FixedSpaceTruncation()` or `trscheme = truncdim(χ)`, I would initialize a CTMRG environment with a total bond dimension $\chi$ first and then plug this into the `leading_boundary` call. However, it turns out that the default randomly initialized environments can give very bad convergence if you directly intialize them with a nontrivial virtual space. Initializing with a trivial virtual space and growing it solves this, but in this case you can't use `FixedSpaceTruncation`.

For example, for the 2D classical Ising model $2 x 2$ unit cell at criticality:

```julia
using TensorKit
using PEPSKit
using Random

using MPSKitModels: classical_ising

Random.seed!(1234) # same problem for many seeds...

# toggle symmetry, but same issue for both
S = Z2Irrep
# S = Trivial

χ = 20
tol = 1.0e-4
maxiter = 1000
verbosity = 2

# initialize
T = classical_ising(S)
O = T[1]
n = InfinitePartitionFunction([O O; O O])
Venv = S == Z2Irrep ? Z2Space(0 => χ / 2, 1 => χ / 2) : ℂ^χ
P = space(O, 2)
trscheme = FixedSpaceTruncation()

# simultaneous CTMRG: very sensitive to initial guess
simultaneous_alg = (;
    alg = :simultaneous,
    tol,
    verbosity,
    trscheme,
    maxiter,
)
env0 = CTMRGEnv(n, Venv)
env_sim, = leading_boundary(env0, n; simultaneous_alg...)
```
gives
```
[ Info: CTMRG init:     obj = -2.222242140629e+01 +1.949978774954e+01im err = 1.0000e+00
┌ Warning: CTMRG cancel 1000:   obj = +3.739239477627e+01 -5.522835410994e-01im err = 2.5066866523e-01  time = 11.77 sec
└ @ PEPSKit ~/ctmbench/lib/PEPSKit.jl/src/algorithms/ctmrg/ctmrg.jl:152
```
Initializing with a trivial virtual space and growing until `truncdim(χ)` to grow the environment first seems to avoid this,
```julia
env0 = CTMRGEnv(n, oneunit(P))
[...]

PR #257, opened 2025-09-12
Add PEPS initialization based on fidelity maximization

pbrehmer:
This PR adds tools to initialize a PEPS based on maximizing the fidelity with respect to another PEPS. As we discussed on Tuesday, this would be useful e.g. to extract a single-site approximation of a SU evolved PEPS on a $2\times2$ unit cell. This would be great to have since SU/FU evolved initial PEPS can then be used for single-site variational optimization.

Note that I use TensorKit's embed branch in order to use the `embed!` function:
```
] add TensorKit#ld-embed
```


src/algorithms/time_evolution/fidelity_maximization.jl
```diff
@@ -0,0 +1,134 @@
...
+    peps_single = noise_amp * InfinitePEPS(randn, scalartype(peps₀), physspace, bondspace) # single-site unit cell with random noise
+
+    peps_uc = InfinitePEPS(fill(only(unitcell(peps_single)), size(peps₀))) # fill peps₀ unit cell with peps_singles
+    maximize_fidelity!(peps_uc, peps₀, envspace; kwargs...) # modifies peps_single in-place
+
+    return peps_single
+end
+
+# maximal virtual space over unit cell
+_spacemax(peps::InfinitePEPS) = argmax(dim, map(p -> argmax(dim, domain(p)), unitcell(peps)))
```
lkdvos: I think you want to use `reduce(supremum, spaces)` to determine the maximal space, since this might be from a different site per sector.

src/algorithms/time_evolution/fidelity_maximization.jl
```diff
@@ -0,0 +1,134 @@
...
+        pepsdst::InfinitePEPS, pepssrc::InfinitePEPS, envspace;
+        maxiter = 5, tol = 1.0e-3, boundary_alg=(; verbosity=1)
+    )
+
+Iteratively maximize the fidelity of `pepssrc` and `pepsdst` where the contents of `pepssrc`
+are embedded into `pepsdst`. To contract the respective networks, the specified `envspace`
+is used on the environment bonds and kept fixed. The CTMRG contraction algorithm is specified
+via the `boundary_alg` `NamedTuple`.
+"""
+function maximize_fidelity!(
```
lkdvos: We could also just use `MPSKit.approximate!` here

lkdvos (commented):
I'll try and go over this in a bit more detail later, just wanted to leave these small comments before I forget

src/algorithms/time_evolution/fidelity_maximization.jl
pbrehmer: Yes, I like that better :-)

I'll try to finish the rest over the next few days, but time is limited at the moment since I'm at a summer school so probably it's still gonna take a bit

src/algorithms/time_evolution/fidelity_maximization.jl
pbrehmer: I gave this a try now and I think it does the right thing, see the test

src/algorithms/optimization/fidelity_initialize.jl
```diff
@@ -0,0 +1,145 @@
...
+function MPSKit.approximate!(
+        pepsdst::InfinitePEPS, pepssrc::InfinitePEPS, envspace;
+        maxiter = 10, tol = 1.0e-3, boundary_alg = (; verbosity = 1)
+    )
+    @assert size(pepsdst) == size(pepssrc) "incompatible unit cell sizes"
+
+    # absorb src PEPS tensors into dst tensors in-place
+    for (pdst, psrc) in zip(unitcell(pepsdst), unitcell(pepssrc))
+        absorb!(pdst, psrc)
+    end
```
leburgel: I think it would be better to have this in `single_site_fidelity_initialize` instead of inside the actual `approximate!` implementation. Now we're really throwing away the initial guess, which is not really what we want if we start from a very good guess already for example.

src/algorithms/optimization/fidelity_initialize.jl
```diff
@@ -0,0 +1,145 @@
...
+    )
+
+Approximate `pepssrc` with `pepsdst` by iteratively maximizing their fidelity where the
+contents of `pepssrc` are embedded into `pepsdst`. To contract the respective networks, the
+specified `envspace` is used on the environment bonds and kept fixed. The CTMRG contraction
+algorithm is specified via the `boundary_alg` `NamedTuple`.
+"""
+function MPSKit.approximate!(
+        pepsdst::InfinitePEPS, pepssrc::InfinitePEPS, envspace;
+        maxiter = 10, tol = 1.0e-3, boundary_alg = (; verbosity = 1)
```
leburgel: Would it be better to also allow an actual algorithm instance instance here, just in case people have this lying around? I think using `boundary_alg = _alg_or_nt(CTMRGAlgorithm, boundary_alg)` and then passing `boundary_alg` as a positional argument to the `leading_boundary_calls` wouldn't be that much more complicated.

src/algorithms/optimization/fidelity_initialize.jl
```diff
@@ -0,0 +1,145 @@
...
+    peps /= sqrt(abs(_local_norm(peps, peps, env)))
+
+    nw₀ = InfiniteSquareNetwork(peps₀, peps)
+    envnw, = leading_boundary(CTMRGEnv(nw₀, envspace), nw₀; boundary_alg...)
+    peps′ = _∂local_norm(peps₀, envnw)
+    fid = 0.0
+    for i in 1:maxiter
+        # compute fidelity from ∂norm
+        fid′ = abs2(_local_norm(peps, peps′))
+        @info @sprintf("Fidmax. iter %d:   fid = %.4e   Δfid = %.4e", i, fid′, fid′ - fid)
```
leburgel: Maybe better to use `LoggingExtras.withlevel(; verbosity)` and `@infov 2 ...` here? Or maybe even a proper `MPSKit.IterLog` eventually, but I think it's good to at least adopt the verbosity levels already.

test/toolbox/fidelity_initialize.jl
```diff
@@ -0,0 +1,34 @@
+using TensorKit
+using PEPSKit
+using PEPSKit: _spacemax
+using Random
+using Test
+
+@testset "Fidelity initialization with approximate!" begin
+    Random.seed!(18092025)
```
leburgel: Did this need a lot of finetuning, or does it mostly just work out of the box?

pbrehmer: Not really, the reason I fixed the seed was that the CTMRG contractions will not converge in the beginning of the fidelity maximization because it seems that the $3\times3$ unit cell and random spaces makes this hard to contract. Fixing the seed doesn't really change this but I wanted to at least make it deterministic.

As a general question: Do we have a feeling on how well this fidelity approximation should be able to converge? Because for the case where we approximate the $3\times3$ PEPS with a single-site PEPS, the infidelity will not drop below 0.1 but it does plateau somewhat.

src/algorithms/optimization/fidelity_initialize.jl
pbrehmer: Initially was too lazy to implement this but I added it now :-)

src/algorithms/optimization/fidelity_initialize.jl
pbrehmer: Good point, added it now

src/algorithms/optimization/fidelity_initialize.jl
pbrehmer: I instead now added a non-mutating `approximate` method because that seems closer to MPSKit; removing `absorb!` from `approximate!` would make it non-mutating anyway. Are you also fine with that?

leburgel: My concern wasn't so much the fact that it was in-place, but rather that doing the absorption step by default pretty much discards the initial guess. I would prefer to be able to actually use a really good initial guess for `pepsdst`.

So definitely agree with having a  non-mutating `approximate`, but even there much of the initial guess is thrown away by the `absorb` step. Similar to the MPSKit algorithms, it seems better to really start from the initial guess that was given as is, and keep the step that prepares a good initial guess for the routine calling `approximate`, i.e., `single_site_fidelity_initialize`.

test/toolbox/fidelity_initialize.jl
leburgel: I see, then I guess using a specific seed is fine for now until we have some tools to properly address #255, which should then also remove the need to finetune here.

I don't think this simple power method here should converge very well, as long as it gets in the right ballpark it should be enough for an initialization. From the point of view of proper fidelity maximization, the gradient used as computed by `_∂local_norm` is really only the partial gradient of the fidelity contraction with respect to the PEPS tensor being optimized over. To get the full gradient we would have to take into account the contributions that come from the environment tensors in the fidelity contraction. We could do this by backpropagating through the entire contraction step just like we do for energy minimizations, but that's probably overkill here. It might be good to verify that using the full gradient gives a clean maximization of the fidelity, but that would be more expensive and likely unnecessary when just getting a good initial state.

If this approach doesn't solve the issues then we can always switch to using the full fidelity gradient instead.

src/PEPSKit.jl
```diff
@@ -96,6 +97,7 @@ export correlator, correlation_length
 export leading_boundary
 export PEPSOptimize, GeomSum, ManualIter, LinSolver, EigSolver
 export fixedpoint
+export single_site_fidelity_initialize
```
leburgel: What do we think about not exporting this for now? The only reason is I'm not too sure about the name, but I don't necessarily want to block this until we agree on a better name, in particular since this is quite an experimental feature.

Not exporting means we can still change the name without breaking things, although this might be bad practice. I think we should anyway settle on a name and export once we start using it in published examples, but keeping it internal before that doesn't seem that bad to me.

src/algorithms/optimization/fidelity_initialize.jl
pbrehmer: Okay I see, then I misunderstood yesterday - I'll fix that up now

src/PEPSKit.jl
pbrehmer: I'd be definitely fine with that. I'm not yet satisfied with the name, I just didn't come up with anything more fitting yet, so I'm happy about any suggestions

pbrehmer:
@leburgel @Yue-Zhengyuan Quick update on running SU with staggered charges on the physical spaces (as discussed yesterday): This does seem to work indeed and I obtain SU energies which are quite similar to the previous runs. Note that I first had to fix the index rotation functions to make this work, see #267. I haven't run variational optimizations on these states yet, but I'll keep you posted.

leburgel:
@pbrehmer, how would you feel about splitting off and adding the `approximate` part of this at least? I think the single-site replacement was a bit tricky conceptually, but the crude fidelity optimization by itself could be a very useful tool, so I think it might be good to revive that part.

pbrehmer:
> @pbrehmer, how would you feel about splitting off and adding the `approximate` part of this at least? I think the single-site replacement was a bit tricky conceptually, but the crude fidelity optimization by itself could be a very useful tool, so I think it might be good to revive that part.

I agree, the crude fidelity `approximate` would indeed be helpful. Then I will remove the single-site initialization and just leave the `approximate` part as is.

pbrehmer:
I removed the single-site initialization and fixed the test. Of course one could add more test cases but I think this is also fine as is. Would be good to go for me.

src/algorithms/approximate.jl
```diff
@@ -0,0 +1,211 @@
...
+    return approximate(ψ₀, ψ, alg)
+end
+function MPSKit.approximate(ψ₀::InfinitePEPS, ψ::InfinitePEPS, alg::FidelityMaxCrude)
+    @assert size(ψ₀) == size(ψ) "incompatible unit cell sizes"
+    @assert all(map((p₀, p) -> space(p₀, 1) == space(p, 1), unitcell(ψ₀), unitcell(ψ))) "incompatible physical spaces"
+
+    log = MPSKit.IterLog("Approximate")
+    return LoggingExtras.withlevel(; alg.verbosity) do
+        # normalize reference PEPS
+        peps_init = ψ # smaller bond spaces
```
Yue-Zhengyuan: I think ψ has larger bond spaces?

src/algorithms/approximate.jl
```diff
@@ -0,0 +1,211 @@
...
+"""
+abstract type ApproximateAlgorithm end
+
+const APPROXIMATE_SYMBOLS = IdDict{Symbol, Type{<:ApproximateAlgorithm}}()
+
+"""
+$(TYPEDEF)
+
+PEPS approximation algorithm maximizing the fidelity by successively applying the fidelity
+derivative with respect to the approximator PEPS.
```
Yue-Zhengyuan, line 14: Can we add a more detailed description of the iterative process?

test/toolbox/approximate.jl
```diff
@@ -0,0 +1,30 @@
...
+@testset "approximate methods with FidelityMaxCrude" begin
+    peps1 = InfinitePEPS(Vphys, Vbond)
+    peps2 = InfinitePEPS(Vphys, Vbond)
+    Random.seed!(sd)
+    peps_approx1, = approximate(
+        peps1, peps2;
+        maxiter = maxiter_approx,
+        boundary_alg = (; maxiter = maxiter_boundary, trunc = truncrank(6), verbosity = 1)
+    )
+    Random.seed!(sd) # reseeding with the same seed is required for the test to pass
```
Yue-Zhengyuan: Is this test merely testing if the algorithm selection works, which is why the two runs need exactly the same `peps1` and `peps2`?

src/algorithms/approximate.jl
```diff
@@ -0,0 +1,211 @@
...
+        env, = leading_boundary(CTMRGEnv(peps, envspace), peps, alg.boundary_alg)
+        peps /= sqrt(abs(_local_norm(peps, peps, env)))
+
+        approximate_loginit!(log, one(real(scalartype(peps))), zero(real(scalartype(peps))))
+        nw₀ = InfiniteSquareNetwork(peps_init, peps) # peps₀ has different virtual spaces than peps
+        envnw, = leading_boundary(CTMRGEnv(nw₀, envspace), nw₀, alg.boundary_alg)
+        peps′ = _∂local_norm(peps_init, envnw)
+        for iter in 1:(alg.maxiter)
+            # compute fidelity from ∂norm
+            fid = abs2(_local_norm(peps, peps′))
```
Yue-Zhengyuan, line 128: I suggest we describe in the docstring how the "crude fidelity" we are using is defined, and how it is related to the exact `⟨ψ₀|ψ⟩`.

src/algorithms/approximate.jl
```diff
@@ -0,0 +1,211 @@
...
+    @assert size(ψ₀) == size(ψ) "incompatible unit cell sizes"
+    @assert all(map((p₀, p) -> space(p₀, 1) == space(p, 1), unitcell(ψ₀), unitcell(ψ))) "incompatible physical spaces"
+
+    log = MPSKit.IterLog("Approximate")
+    return LoggingExtras.withlevel(; alg.verbosity) do
+        # normalize reference PEPS
+        peps_init = ψ # smaller bond spaces
+        envspace = domain(ψ₀[1])[1]
+        env₀, = leading_boundary(CTMRGEnv(peps_init, envspace), peps_init, alg.boundary_alg)
+        peps_init /= sqrt(abs(_local_norm(peps_init, peps_init, env₀))) # normalize to ensure that fidelity is bounded by 1
```
Yue-Zhengyuan: In Julia, doing `ψ /= nrm` is exactly equivalent to `ψ = ψ / nrm` (@lkdvos correct me if I'm wrong, or the behavior is defined differently for PEPS/TensorMaps). After this assignment (rebinding), any changes to `ψ` does not affect the input variable. Then the function is not mutating, and we can avoid introducing new names like `peps_init`, `peps`.

Also, `env₀` being the environment for `ψ`, and `env` being the environment for `ψ₀` looks off to me.

src/algorithms/approximate.jl
pbrehmer: I think in the original context where I implemented this we were treating the case where we wanted to initialize a PEPS with larger $D$ from a smaller $D$ PEPS. (I guess one frequently encounters this situation when optimizing PEPS where larger $D$ optimization runs are hard to initialize properly.) But in principle `ψ` could of course also have a larger bond space - this would correspond to a lower rank approximation of a large $D$ PEPS. In principle both cases should be allowed (I think?), so I'll just remove the comment.

test/toolbox/approximate.jl
pbrehmer: Yes indeed. It doesn't hurt testing that since the test itself is quite cheap.

src/algorithms/approximate.jl
pbrehmer: I introduced new names on purpose since I felt that was a bit cleaner and made it easier to read to iterative process. Note that `env₀` is the environment of `peps₀` (see updated name) and `env` of `peps`, i.e. the `₀` index refers to the zeroth iteration step of the `approximate` routine, not the initial guess. Instead, the first iteration step is determined by the initial guess `ψ₀` since a perfect initial guess would immediately terminate the iteration. This is all slightly confusing so I wanted to work with different names.

src/algorithms/approximate.jl
pbrehmer: I added some text describing the algorithm.

src/algorithms/approximate.jl
pbrehmer: I added some words on this in the algorithm description.

test/toolbox/approximate.jl
Yue-Zhengyuan, line 1: What do you think about adding a more elementary test?
```julia
using Test
using Random
using TensorKit
using PEPSKit
using PEPSKit: _∂local_norm, _local_norm

sd = 123456
Random.seed!(sd)
Vphys, Vbond, Venv = ℂ^2, ℂ^3, ℂ^6

ket = InfinitePEPS(Vphys, Vbond)
bra = InfinitePEPS(Vphys, Vbond)
network = InfiniteSquareNetwork(ket, bra)
env = CTMRGEnv(network, Venv)

nrm1 = _local_norm(ket, bra, env)
∂norm = _∂local_norm(ket, env)
nrm2 = _local_norm(bra, ∂norm)
@test nrm1 ≈ nrm2
```
It helps me (and maybe others) understand what the underlying functions are doing.

src/algorithms/approximate.jl
```diff
@@ -0,0 +1,218 @@
...
+const APPROXIMATE_SYMBOLS = IdDict{Symbol, Type{<:ApproximateAlgorithm}}()
+
+"""
+$(TYPEDEF)
+
+PEPS approximation algorithm maximizing the fidelity by successively applying the fidelity
+derivative with respect to the approximator PEPS.
+
+Starting from an initial guess, the algorithm first computes the CTMRG environment of the ⟨ψᵢ|ψᵢ₊₁⟩
+network. From that, the fidelity derivative ∂F/∂ψᵢ₊₁ is computed for each PEPS tensor in the unit cell,
```
Yue-Zhengyuan: Please explain what is `F` before introducing the notation.

src/algorithms/approximate.jl
```diff
@@ -0,0 +1,218 @@
...
+corresponding to the norm network of ψᵢ where the respective bra PEPS tensor is missing. Then we update
+ψᵢ = ψᵢ₊₁ and ψᵢ₊₁ = ∂F/∂ψᵢ₊₁, and compute the fidelity |⟨ψᵢ|ψᵢ₊₁⟩|². This is iterated until the
```
Yue-Zhengyuan: When describing an iterative process I would write something like `ψᵢ₊₁ = f(ψᵢ)`. So `ψᵢ = ψᵢ₊₁` feels a bit unnatural. Again, please explain what is `F`.

src/algorithms/approximate.jl
```diff
@@ -0,0 +1,218 @@
...
+$(SIGNATURES)
+
+Compute the `InfinitePEPS` resulting from removing the bra PEPS tensors in `_local_norm`.
+"""
+function _∂local_norm(peps::InfinitePEPS, env::CTMRGEnv)
+    return InfinitePEPS(map(ind -> _∂contract_site(ind, peps, env), eachcoordinate(peps)))
+end
+
+# contract CTMRG environment leaving open the bra-PEPS virtual and physical bonds
+function _∂contract_site(ind::Tuple{Int, Int}, peps::InfinitePEPS, env::CTMRGEnv)
```
Yue-Zhengyuan: For clarity, maybe we just rename `peps` to `ket`?

src/algorithms/approximate.jl
```diff
@@ -0,0 +1,218 @@
...
+    end
+end
+
+
+"""
+$(SIGNATURES)
+
+Compute the `InfinitePEPS` resulting from removing the bra PEPS tensors in `_local_norm`.
+"""
+function _∂local_norm(peps::InfinitePEPS, env::CTMRGEnv)
```
Yue-Zhengyuan: How about renaming `peps` to `ket`?

Yue-Zhengyuan (commented):
Some more nitpicky comments

test/toolbox/approximate.jl
```diff
@@ -0,0 +1,31 @@
...
+    peps_approx2, = approximate(
+        peps1, peps2, FidelityMaxCrude(;
+            maxiter = maxiter_approx,
+            boundary_alg = (; maxiter = maxiter_boundary, trunc = truncrank(χ), verbosity = 1)
+        )
+    )
+
+    @test peps_approx1 ≈ peps_approx2
```
Yue-Zhengyuan, line 47: As a less trivial test, can we do two approximations with two different seeds (leading to different CTMRGEnv during the run), and finally check if the fidelity of the two results is close enough to 1?

src/algorithms/approximate.jl
```diff
@@ -0,0 +1,218 @@
...
+
+"""
+$(SIGNATURES)
+
+Sum over `contract_local_norm` values of all unit cell entries.
+"""
+function _local_norm(ket::InfinitePEPS, bra::InfinitePEPS, env::CTMRGEnv)
+    return sum(ind -> contract_local_norm((ind,), ket, bra, env), eachcoordinate(ket))
+end
+function _local_norm(peps::InfinitePEPS, ∂norm::InfinitePEPS)
```
Yue-Zhengyuan: Also, I suggest renaming `peps` to `bra` here. Maybe also a more descriptive name for `∂norm` to emphasize that it is the derivative *with respect to the bra tensors*?

test/toolbox/approximate.jl
pbrehmer: Sure, I really do like that this illustrates how the internal functions are used.

test/toolbox/approximate.jl
pbrehmer: Yes, I'll add that to the existing test set.

src/algorithms/approximate.jl
pbrehmer: I now omit the use of `F` (which was technically wrong anyway).

src/algorithms/approximate.jl
pbrehmer: I find that `∂norm` is fine as a name, especially since it does have a docstring where it is explicitly stated that the derivative is with respect to the bra.

test/toolbox/approximate.jl
Yue-Zhengyuan: On my Mac this new test fails miserably...

pbrehmer: How miserably does it fail? In general I think we can't expect the fidelities to approach 1 very closely since the algorithm is rather crude but I thought that being within 10% should be possible. Does is get better for other seeds?

pbrehmer: Okay, trying around with different seeds this test will always fail. Maybe the tolerance is too strict: If one evaluates the fidelity of two random PEPS, it will in general be very small. So perhaps it is enough if the fidelity is already of the order of $10^{-1}$ (as opposed to $10^{-7}$ for two random PEPS).

Yue-Zhengyuan: For me the fidelity between two runs with different seeds can be ad low as 0.3. If random states are hard, how about using SU to get a physical state and try to approximate it?

pbrehmer: I don't think the problem is the initial state, it's rather the algorithm itself which will only bring two states in the same ballpark. One can already see in the convergence itself that it is hard to converge the fidelity below an error of $10^{-2}$. But it would be interesting to see what happens if one initializes with SU.

Yue-Zhengyuan:
If I take a random PEPS `ψ1`, and obtain another `ψ2` by performing a gauge transformation (inserting `X inv(X)` on virtual bonds), will the crude fidelity defined here evaluates to 1? This is not obvious to me by looking at the code.

test/toolbox/approximate.jl
```diff
@@ -57,5 +57,5 @@ end
     envnw, = leading_boundary(env1, InfiniteSquareNetwork(peps_approx1, peps_approx3); maxiter = maxiter_boundary)
     ∂norm = _∂local_norm(peps_approx1, envnw)
     fid = abs2(_local_norm(peps_approx3, ∂norm))
-    @test fid ≈ 1 atol = 0.8 # can't expect much more than 10⁻¹
+    @test fid ≈ 1 atol = 0.9 # judging from different seeds, one generally can't expect much more than 10⁻¹ fidelity
```
Yue-Zhengyuan: I think we just test `fid > 0.1` if it is so bad

pbrehmer:
> If I take a random PEPS `ψ1`, and obtain another `ψ2` by performing a gauge transformation (inserting `X inv(X)` on virtual bonds), will the crude fidelity defined here evaluates to 1? This is not obvious to me by looking at the code.

I don't think it will evaluate to 1 due to the way CTMRG (without spatial symmetries) is not gauge-invariant if you insert a gauge transformation on the virtual bonds since the north and east edge will absorb $X$ whereas the south and west edge will absorb $X^{-1}$. It would only be one if that inner product contraction environment of `InfiniteSquareNetwork(ψ1, ψ2)` would produce the same network value as the respective `ψ1` and `ψ2` norm networks - which it generally won't due to the gauge transformation, I believe.

Edit: If that is true, then that also kind of answers why the fidelity test won't approach 1, since two different initializations won't end up in the same gauge.

Yue-Zhengyuan:
Then I suggest we also explain in docstring how the fidelity is crudely approximated so that users can have the right expectations.

Yue-Zhengyuan:
Just remembered this discussion #316 that is very relevant.

Yue-Zhengyuan:
~I think that's all I wanted to say.~ @leburgel do you have additional comments?

Yue-Zhengyuan:
Oh, can we maybe use BP gauging fixing during the iteration to improve the robustness?

pbrehmer:
> Oh, can we maybe use BP gauging fixing during the iteration to improve the robustness?

I gave this a quick try. I tried both gauging before the power-method iteration and during the iteration. This didn't really help to stabilize the algorithm from what I found. In case you had something more specific in mind, let me know. I would rather be in favor of leaving the algorithm as is because I think it already is useful for the original use case we had in mind (optimization initialization). Maybe we can try to come up with a more refined algorithm in the future but I also don't want to spend much more time on this PR right now.

PR #262, opened 2025-09-23
Simple update for PEPO (3-site version)

Yue-Zhengyuan:
This PR adds the 3-site version of simple update for PEPO, which deals with Hamiltonians containing up to next-nearest neighbor terms. The J1-J2 model is used to test the algorithm by comparing the calculated finite-temperature energy (both tr(ρH) and ⟨ρ|H|ρ⟩) with high-temperature expansion results in https://doi.org/10.1103/PhysRevB.67.014416 (as @ogauthe did in #215).

### DIFF
--- a/src/algorithms/time_evolution/evoltools.jl
+++ b/src/algorithms/time_evolution/evoltools.jl
@@ -189,7 +189,7 @@ Apply 2-site `gate` on the reduced matrices `a`, `b`
 ```
 """
 function _apply_gate(
-        a::AbstractTensorMap{T, S}, b::AbstractTensorMap{T, S},
+        a::AbstractTensorMap, b::AbstractTensorMap,
         gate::AbstractTensorMap{T, S, 2, 2}, trscheme::TruncationScheme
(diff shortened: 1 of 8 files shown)


src/algorithms/time_evolution/evoltools.jl
```diff
@@ -189,7 +189,7 @@ Apply 2-site `gate` on the reduced matrices `a`, `b`
 ```
 """
 function _apply_gate(
-        a::AbstractTensorMap{T, S}, b::AbstractTensorMap{T, S},
+        a::AbstractTensorMap, b::AbstractTensorMap,
```
Yue-Zhengyuan, line 192: This change is to allow application of real gates on complex states.

lkdvos (approved):
Overall looks good to me. I see you changed from `AbstractTensorMap{<:Any,<:Any,1,N}` to `AbstractTensorMap{<:Any,<:Any,N,1}`, is this just more convenient, or was there any special reason for that?

PR #264, opened 2025-09-26
Environment initialization for CTMRG + `ProductStateEnv`

### LINKED ISSUE #136: Initial enviroment tensors from the local tensors
Currently, PEPSKit initializes the environment’s corner and edge tensors randomly. However, I have found that this random initialization often leads to poor convergence—especially when working with symmetric tensors. In my experience, initializing these tensors by contracting the extra virtual legs from the bra and ket local tensors significantly improves convergence. It would be very useful if PEPSKit offered an option to use this improved initialization method, so I wouldn’t have to study the code in detail and implement it myself.

### LINKED ISSUE #255: Issue with `SimultaneousCTMRG` convergence starting from random initial environment

leburgel:
Attempt at some better environment initialization routines for CTMRG. Should solve #255 once it's finished and documented.

This implements an `initialize_environment` method which initializes a `CTMRGEnv` for a given network according to a specific `InitializationStyle`. I specified three kinds of `InitializationStyle`s: `RandomInitialization` for initializing a random environment, `ProductStateInitialization` for initializing a (potentially embedded) product state environment, and `ApplicationInitialization` which grows an initial environment from a product state according to a given truncation scheme. Names, arguments and which ones are optional for discussion, but at least this already handles the test case from #255 in a fairly straightforward way.


src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,56 @@
...
+        )
+    )
+    boundary_alg = (; boundary_alg..., trscheme) # merge trscheme with optional alg definition
+    env = initialize_environment(elt, n, ProductStateInitialization())
+    env, = leading_boundary(env, n; boundary_alg...)
+    return env
+end
+
+function initialize_environment(n::InfiniteSquareNetwork, args...; kwargs...)
+    return initialize_environment(ComplexF64, n, args...; kwargs...)
```
lkdvos:
```suggestion
    return initialize_environment(scalartype(n), n, args...; kwargs...)
```

src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,56 @@
...
+end
+function initialize_environment(A::Union{InfinitePEPS, InfinitePartitionFunction}, args...; kwargs...)
+    return initialize_environment(ComplexF64, A, args...; kwargs...)
```
lkdvos:
```suggestion
    return initialize_environment(scalartype(A), A, args...; kwargs...)
```

src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,56 @@
...
+end
+
+function initialize_environment(
+        elt::Type{<:Number},
+        n::InfiniteSquareNetwork,
+        ::ApplicationInitialization,
+        trscheme::TruncationScheme;
+        boundary_alg = (;
+            alg = :sequential, tol = 1.0e-5, maxiter = 10, verbosity = -1,
+        )
```
lkdvos: I think I would imagine this being a part of the `ApplicationInitialization` struct, and the same for the `trscheme`?

src/algorithms/ctmrg/initialization.jl
lkdvos, line 1: Having a look at the argument order, I think something feels a little off with the `virtual_spaces` as final argument to me.
(Disclaimer that everything that follows is subjective and therefore please do tell me to shut up if you disagree)

In general, most of our "algorithm functions" actually follow a very similar type of interface, based on the purpose of the arguments:
```julia
algorithm_function(init, operator, algorithm, [additional_optional_structures])
```

With this in mind, I think I would try to make something like this work:
```julia
initialize_environment([spaces], network, alg)
initialize_environment([(T, spaces)], network, alg)
```
but I wonder if there is a more elegant solution for the `T`, possibly just putting it in the keyword arguments. I don't think there are many occasions where `T != scalartype(network)` is desired, and we can try to make this at least `@constinferred`?

src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,56 @@
+abstract type InitializationStyle end
+struct ProductStateInitialization <: InitializationStyle end
+struct RandomInitialization <: InitializationStyle end
```
lkdvos: Should this one encapsulate which initialization function we are using, so it can switch between `randn` and `rand`?

src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,56 @@
...
+struct ApplicationInitialization <: InitializationStyle end
+
+function initialize_environment(
```
Yue-Zhengyuan: How about we name the function as `initialize_CTMRGEnv` to be more specific?

src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,56 @@
...
+        ::RandomInitialization,
+        virtual_spaces... = oneunit(spacetype(n)),
+    )
+    return CTMRGEnv(randn, elt, n, virtual_spaces...)
+end
+
+function initialize_environment(
+        elt::Type{<:Number},
+        n::InfiniteSquareNetwork,
+        ::ProductStateInitialization,
```
Yue-Zhengyuan: This is just an idea, which doesn't need to be addressed here. 

As I realized in the last meeting, `SUWeight` is a special kind of product-state CTMRG environment; maybe it's useful to add an SUWeight-to-CTMRGEnv conversion. 

For PEPS environment, one can use "trivial simple update" (repeatedly applying identity gates) to get the SUWeight, and use this result to initialize a product-state CTMRG environment. Though I'm not sure how much speed up in CTMRG convergence it can bring. 

For partition functions, things are a bit more subtle. Maybe it can be regarded as an iPEPS (or iPEPO?) with a trivial physical space in order to perform the trivial-SU.

src/algorithms/ctmrg/initialization.jl
leburgel: For the `trscheme`, I kind of think of this as equivalent to the virtual space specification in the other schemes, which is why I kept it separate. See the comment below for some more explanation on that.

For the algorithm being part of the `ApplicationInitialization` struct, that was also my first thought, but it really wasn't clear to me how to do this in a simple way. Part of the reason is that I would like to reuse this for other kinds of initializations (e.g. PEPS fixed point initialization for Hamiltonians or PEPOs, using different kinds of algorithms such as `SimpleUpdate` or variational truncation). So it didn't really make sense to have constructors for `ApplicationInitialization` which put in a default algorithm since the possible algorithm really depend on the objects we're dealing with. But at the same time I really wanted a simple calling signature with a default algorithm that users wouldn't have to manually construct first. So I kind of like the black box behavior now, but I'm also not too happy with the implementation

If we make the algorithm a part of the struct,
```julia
struct ApplicationInitialization{A} <: InitializationStyle
    alg::A
end
```
we could get in the algorithm based on the object here through something like
```julia
function ApplicationInitialization(network::InfiniteSquareNetwork; kwargs...)
    return ApplicationIntialization(CTMRGAlgorithm(; kwargs...))
end
```
Then I just move the defaults there, and it cleans up the `initialize_environment` signature. Would that be better?

src/algorithms/ctmrg/initialization.jl
leburgel: Yes, that makes a lot of sense.

src/algorithms/ctmrg/initialization.jl
leburgel: I wouldn't mind changing `T` to a keyword argument, the only reason I added it like this was to stay closer to the environment and state constructor signatures.

The choice of position for the `virtual_spaces` argument was purely pragmatic, as this is basically the only way I could make it compatible with the way the space specification in the `CTMRGEnv` constructor works. Since the virtual space specification for the different directions works with optional positional arguments, putting `virtual_spaces` as the last positional argument was the only way to allow for more elaborate virtual space specifications as far as I could tell.

If we want to move it, we would probably need some kind of `CTMRGEnvSpaces` structure which contains all the information on the virtual spaces in a single object that we can then put wherever we want. This is anyway something we might want to get rid of signatures like `CTMRGEnv(ComplexF64, randn, network, virtual_spaces...)` in favor of something like `randn(ComplexF64, spaces::CTMRGEnvSpaces)`. That's also part of the reason I introduced these `_fill_environment_virtual_spaces` methods in #266, since these would come in handy when defining `CTMRGEnvSpaces(network, virtual_spaces...)`. We can decide to do this first, and at the same time change how we internally specify all other space specification.

Starting from the assumption that `virtual_spaces` needs (at least right now) to be the last positional argument, my reasoning for the argument order was to have something like
```julia
initialize_environment([T,] network, alg, virtual_space_specification)
```
The way you would specify a space depends on the algorithm. For product state or random initialization, the virtual space specification is just an actual space specification as it would be passed to the constructor, hence the whole reason the space specification is positioned last. For an `ApplicationInitialization`, a full space specification doesn't really make sense, what matters then is how the virtual spaces are truncated during the application iterations. That's why I chose to specify spaces here using a `TruncationScheme`, and also why the `trscheme` was the last positional argument in the `initialize_environment` signature for `ApplicationInitialization`.

src/algorithms/ctmrg/initialization.jl
leburgel: In the ideal case I think we would like to use an `initialize_environment` method quite generally across different objects and algorithms, or at least that was my initial thought. Depending on how much sense that makes or if we actually manage to use the same signature in different contexts, we can rename to be more specific.

Unless we already think using the same method name won't really work, then we can rename now.

src/algorithms/ctmrg/initialization.jl
leburgel: I think the `SUWeight` to contraction environment conversion would be very nice to have. Regardless of whether this is useful to speed up CTMRG convergence, being able to play around with a product state contraction environment would be very handy.

This is also very much related to #223, which is based around the notion of a particular product state contraction environment, there for the purpose of bringing a PEPS into a specific gauge. In particular, I think the `CTMRGEnv` obtained from a set of `SUWeights` should essentially be equivalent to the `BPEnv` computed there. In that sense, it would be particularly nice if we could reuse the same routine for the determination of a set of `SUWeights` for a 'trivial operator' and finding a product state `BPEnv` contraction environment if this is possible. For now, an `SUWeight` to `CTMRGEnv` by itself would already be great in practice.

Yue-Zhengyuan: I'll try to make an PR on this soon, but only restricted to the case where the SUWeight is for an iPEPS. The iPEPO case can be added later.

Yue-Zhengyuan: To be handled in #275.

Yue-Zhengyuan:
I just noticed one more way to initialize the environment (I already have some code on this in my neighborhood tensor update branch), which we may call `Neighbo(u)rhoodInitialization`. This was mentioned in [1104.5463](https://arxiv.org/abs/1104.5463):

> Initially, the corner and edge tensors at position [x,y] are constructed similarly as a reduced tensor a[x,y], by multiplying the tensor A[x,y] to its conjugate and fusing the bond indices,9 where we trace over the legs directed toward an open boundary.

For example, an edge tensor is constructed like this:
```
           2    
         ╱      
┌-----ket----- 4
|    ╱ |        
|   6  |   1    
|      | ╱      
└-----bra----- 3
     ╱          
    5           
```
In this way, the initialized environment directly have boundary dimension $\chi = D^2$. 

Recently I realized that having a good initialization is also important during full update. Right now, I'm using the environment from the last iteration as initialization for the next step. But sometimes even this doesn't work well. So I kind of hope that we can finish this PR soon.

src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,59 @@
...
+        virtual_spaces... = oneunit(spacetype(n)),
+    )
+    i = one(sectortype(n))
+    env = CTMRGEnv(ones, elt, n, virtual_spaces...)
+    for (dir, r, c) in Iterators.product(axes(env)...)
+        @assert i in blocksectors(env.corners[dir, r, c])
+        for (c, b) in blocks(env.corners[dir, r, c])
+            b .= 0
+            c == i && (b[1, 1] = 1)
```
Yue-Zhengyuan: We definitely need a test with fermionic iPEPS on this, in case parity-odd elements become -1 due to twists.

leburgel:
Given the fact that this PR might be useful for #321, I'll try to pick this up again now. However, considering the original purpose, the review comments, and the new use case in #321, it's not entirely clear to me where to go with this. Let me try to summarize.

The original motivation for this PR was to solve #255, by making it 'easy' to initialize a CTMRG environment with a given bond dimension/virtual space by growing it from a smaller environment (with a product state being the default starting point. Basically, package up what people do all the time in a single method. The easiest way to do this would be to write a specialized `initialize_ctmrg_environment`.

However, since initializing environments is something very common, I tried to make it a bit more generic so that the same signature and initialization flavors could be reused in other places. This came out a bit awkward right away, as can be seen from the review comments. A first obvious problem is that the specification of virtual environment spaces really depends on what exactly we're doing, which gives rise to some awkward argument orderings. In addition, to actually 'grow' an environment we need to apply a few iterations of an algorithm (see `ApplicationInitialization`), which requires some algorithm settings to be specified. Again, this all made sense in the context of the original application to CTMRG, but looks very arbitrary when seen in a generic context.

Now in #321 there's an additional aspect, where we actually want to initialize a different kind of environment depending on the contraction algorithm we use. For now the C4v implementation uses the normal `CTMRGEnv` struct, but it would probably make more sense to use a custom `C4vCTMRGEnv`. Anticipating that we also want to support boundary MPS contractions soon™, this gives more different environment types for more contraction algorithms.

Trying to consider all of this feels a bit annoying, which is also why this PR got completely stalled in the first place. To get this moving, I think the first thing to decide is whether we want to solve the whole (potentially still growing) problem here, or we first just address the original issue #255.

To address the original problem, I would switch to just writing an `initialize_ctmrg_environment` with the minimal functionality required to address the issue.

To go for a full generic `initialize_environment`, I think it becomes inevitable to include an actual contraction algorithm instance in the arguments. Either this, or an equivalent named tuple that can identify multiple different contraction algorithms (and not just flavors of CTMRG like we support now). That would lead to something like
```julia
initialize_environment([T,] network, contraction_algorithm, initialization_style, virtual_space_specification)
```
for the generic case withing PEPSKit.jl alone. The question then becomes if this is compatible with other kinds of environment initializations (e.g. in MPSKit.jl), and if we still want it to be.

Asking for any and all opinions from @lkdvos, @pbrehmer and @Yue-Zhengyuan, after which I'll have a go at getting this to a point where it's usable.

src/environments/product_state_environments.jl
```diff
@@ -0,0 +1,141 @@
...
+
+Construct a product state environment by specifying a corresponding [`InfiniteSquareNetwork`](@ref).
+"""
+function ProductStateEnv(f, T, network::InfiniteSquareNetwork)
+    Ds_north = _north_edge_physical_spaces(network)
+    Ds_east = _east_edge_physical_spaces(network)
+    return ProductStateEnv(f, T, Ds_north, Ds_east)
+end
+function ProductStateEnv(network::InfiniteSquareNetwork)
+    return ProductStateEnv(ones, scalartype(network), network) # TODO: do we want to use a different default function?
```
Yue-Zhengyuan: I suggest we default to creating a "trivial" environment: with the trivial 1D chargeless environment spaces removed, the edge tensors should be an identity map. Currently it can be done with `CTMRGEnv(SUWeight(peps))` for InfinitePEPS. But here your use of `one` fills all tensor elements, not only the diagonal elements.

We may consider adding a keyword argument to control whether we want to fill only the diagonal elements or all elements for the edge tensors.

src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,56 @@
...
+    f::F
+    ApplicationInitialization(f::F = ones) where {F} = new{F}(f)
+end
+
+# initialize randomly, using same virtual space specification as the CTMRGEnv constructor
+function initialize_ctmrg_environment(
+        elt::Type{<:Number},
+        n::InfiniteSquareNetwork,
+        alg::RandomInitialization,
+        virtual_spaces... = oneunit(spacetype(n)),
```
Yue-Zhengyuan, line 11: How about calling it `environment_spaces` (or `env_spaces` for short) to avoid confusion with the virtual spaces in `n`? (Unless you want to be consistent with `src/environments/ctmrg_environments.jl`)

test/ctmrg/initialization.jl
```diff
@@ -0,0 +1,51 @@
...
+maxiter = 1000
+verbosity = 2
+trunc = truncrank(χ)
+boundary_alg = (;
+    alg = :simultaneous,
+    tol,
+    verbosity,
+    trunc,
+    maxiter,
+)
```
Yue-Zhengyuan:
```suggestion
boundary_alg = (; alg = :simultaneous, tol, verbosity, trunc, maxiter)
```
Just want to condense it into a single line.

Yue-Zhengyuan:
I applied `ProductStateInitialization` to finite-T SU tests.

src/algorithms/ctmrg/initialization.jl
leburgel: I was indeed trying to be consistent with `src/environments/ctmrg_environments.jl`, but I'm definitely open to renaming both instances it this would be less confusing.

src/environments/product_state_environments.jl
leburgel: The problem is that this way of initializing only really makes sense for `InfiniteSquareNetwork{<:PEPSSandwich}`, while here I was trying to come up with something that works in general, in particular also for partition functions. To initialize a product state environment consisting of identities between two factors of the PEPS virtual space, we could call `ProductStateEnv` with a custom initializing function. Something like `ProductStateEnv(bipartite_id, T, n)`, where `bipartite_id` is a custom function for initializing an identity between the two factors of a two-component `ProductSpace`.

So I fully agree with your suggestion, but it's not that straightforward to make this a default for all `InfiniteSquareNetwork`s.

leburgel:
Kind of starting over on this one, hopefully I can work toward something useful that can be merged soon.

I'm giving up on trying to fit everything into a single `initialize_environment` method that works for all algorithms, and focusing on first writing a specific `initialize_ctmrg_environment` that solves the issue that actually motivated this PR in the first place.

Currently I have three initialization styles in mind that are useful:
- `RandomInitialization`: initializes a random `CTMRGEnv` with gives virtual spaces, essentially coinciding with the current default constructor.
- `ProductStateInitialization`: initializes a `CTMRGEnv` with trivial one-dimensional virtual spaces, corresponding to a product state environment to which we add trivial virtual legs.
- `ApplicationInitialization`: initializes a `CTMRGEnv` by applying a single CTMRG iteration to a (optionally) given product state environment, giving an environment with a bond dimension $D^2$.

These should solve both #136 and #255 already, without considering how to use this in optimization for now. In particular, none of these take a contraction algorithm as an input, which means we can avoid the nightmare of plumbing in all possible configurations. Any additional steps needed for initialization should then just be handled by the user.

These last two initialization styles make use of a product state environment as an intermediate step, where especially for the `ApplicationInitialization` one might want to manually choose a specific product state as a boundary condition. Therefore, it seemed useful to me to introduce a dedicated `ProductStateEnv` to represent product state environments. In particular, I think we want this to work for general `InfiniteSquareNetwork`s. This means I can't just repurpose `BPEnv`, which is just exactly a product state environment with some extra structure.

To push this over the line, I'm planning to
- Replace `BPEnv` with a more general `ProductStateEnv`, and update all corresponding code.
- Add additional initialization methods specifically for PEPS overlaps, in particular an initialization from a product state where the tensors are identities between the "ket" and "bra" PEPS virtual spaces.

leburgel:
Forgot about this, again.

With another remarkable delay, I added an option to use `ProductStateInitialization(bipartite_id)` specifically for two-layer `InfiniteSquareNetwork`s (i.e. PEPS overlaps) which does exactly what @Yue-Zhengyuan suggested [above](https://github.com/QuantumKitHub/PEPSKit.jl/pull/264#issuecomment-3514782347) and was already suggested in #136.

I picked the name `bipartite_id` because it kind of sounds like what it should do, even though it returns a (2, 0) tensor and not a (1, 1) linear map. I also exported it, because this seems like something people will want to use. Any suggestions on the name and whether or not it's a good idea to export are very welcome.

This definitely needs an example showcasing the different options. But I don't really have time to write one now, maybe it's best to just add the implementation already.

The final thing I still wanted to do was to try and replace `BPEnv` by the more general `ProductStateEnvironment` completely, and adding dispatch blockers on anything that requires a two-layer structure. It seemed a bit silly to me to keep both of these around, since they are essentially just the same thing. I had a look at this, and while it's definitely possible it feels like quite a bit of work so I didn't get to that. However, removing `BPEnv` is very much breaking, and we're just about to release v0.8.0. So I wanted to check opinions:
- Do we spend the time completely rewrite BP in terms of `ProductStateEnv`, and remove `BPEnv` before tagging v0.8.0?
- Or do we just keep both around for now, and save a rewrite for later? This choice could be relevant for #372 as well, since we might be able to implement everything in terms of a single environment type?

Any thoughts, @Yue-Zhengyuan @lkdvos?

src/utility/util.jl
```diff
@@ -208,3 +208,17 @@ function _permute_to_last(axes::NTuple{N, Int}, ax::Int) where {N}
     new_axes = (ntuple(i -> axes[biperm[1][i]], N - 1)..., ax)
     return new_axes, biperm
 end
+
+"""
+    bipartite_id([T::Type=Float64], V::ProductSpace{S, 2}) where {S}
+
+Constructs a tensor corresonding to the (permutation of) the identity operator between
+two componens of a bipartite product space.
```
Yue-Zhengyuan: Typo: `componens`

The word "bipartite" is a bit ambiguous here. Until now it means the checkerboard lattice structure with 2 independent tensors. 

Here, by "between two componens of a bipartite product space" do you mean "between two layers of a two-layer network"?

src/algorithms/ctmrg/initialization.jl
Yue-Zhengyuan, line 1: Better to add docstrings to describe what each InitializationStyle is actually doing, especially `ApplicationInitialization` whose behavior is not obvious by just looking at the name.

src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,56 @@
...
+    )
+    env = CTMRGEnv(ProductStateEnv(alg.f, elt, n))
+    return env
+end
+
+function initialize_ctmrg_environment(
+        elt::Type{<:Number},
+        n::InfiniteSquareNetwork,
+        alg::ApplicationInitialization,
+        env0::ProductStateEnv = ProductStateEnv(alg.f, elt, n)
```
Yue-Zhengyuan: It might be helpful not to restrict `env0` to be a ProductStateEnv. For example, one may start from any CTMRGEnv with a small bond dimension, and do one step of CTMRG. But the current approach is also fine if you don't want too much freedom.

Yue-Zhengyuan (commented):
Some small comments.

> Do we spend the time completely rewrite BP in terms of ProductStateEnv

I'm in favor of this, since I feel that `BPEnv` and `ProductStateEnv` are (exactly?) the same thing, with the renaming `BPEnv.messages` to `ProductStateEnv.edges`.

src/algorithms/ctmrg/initialization.jl
lkdvos: agreed on this!

lkdvos (commented):
Overall looks good to me, I think I still have a few questions/comments about what we were discussing on Wednesday too:

- It seems to me like the `ProductStateEnvironment` really is only ever used as a means to then immediately be converted into a `CTMRGEnv`. Just trying to wrap my head around this: is the goal to have an easy way to construct a CTMRG environment with trivial chi, or is the goal to be able to supply specific (2,0) tensors for this?

In the first case, how would you feel about simply adding `product_ctmrg_environment` as a function instead of the full struct? I think this would remove quite a bit of the needed overloads you now have for this, and would consist mostly of just filling in the correct `unitspace`s and then dispatching to the already existing `CTMRGEnv` constructors.

- I can't really see where the `bipartite_id` function is being used immediately. If this is for initializing a trivial product environment state, I think this is what the `BPEnv` does already (?), so in that sense this would be equivalent to `CTMRGEnv(BPEnv(state))`. Is this something we want to expose/export on top of that, or should we really just an example in the docs on how to actually do this?

Keep in mind that I definitely might still be missing something, but that does seem to also bypass the issue about naming and duplicating the `BPEnv`

leburgel:
> It seems to me like the `ProductStateEnvironment` really is only ever used as a means to then immediately be converted into a `CTMRGEnv`. Just trying to wrap my head around this: is the goal to have an easy way to construct a CTMRG environment with trivial chi, or is the goal to be able to supply specific (2,0) tensors for this?

I don't think I did a good job at making this clear, but one of the main goals of this PR at this point is to address the second case. The only real reason I saw the need to introduce a `ProductStateEnvironment` is because we want to be able to start from a specific product state made up of (N, 0) tensors, where N is not necessarily 2. For example, for a standard partition function (N = 1) corresponding to a model with symmetry breaking, we might want to start from a product state with a specific magnetization to end up with the proper fixed point corresponding to that magnetization. This functionality was explicitly requested, and I think it's a very reasonable thing to do.

The fact that we want to be able to supply specific tensors makes it convenient to have a struct for this, which aside from just storing the tensors also documents the indexing and unit cell conventions for a product state environment (which is quite important when you want to do a non-trivial initialization for systems with a non-trivial unit cell.

The fact that this should work for N != 2 makes it such that we can't really use the existing `BPEnv` type for this, even though it really is exactly the right structure. It just didn't make sense to me to use `BPEnv` for anything that doesn't have the two-layer bra-ket structure that is assumed in almost all of the code that actually works with `BPEnv` instances. For example, we could just naively use the `BPEnv` constructor directly on a set of (1, 0) tensors to encode a product state environment, but this would immediately run into trouble since even the `CTMRGEnv(::BPEnv)` constructor assumes N = 2 as well as a very specific leg ordering of the edge tensors.

> In the first case, how would you feel about simply adding `product_ctmrg_environment` as a function instead of the full struct? I think this would remove quite a bit of the needed overloads you now have for this, and would consist mostly of just filling in the correct `unitspace`s and then dispatching to the already existing `CTMRGEnv` constructors.

This is an option, but to address the goal above we would have to be able to pass in specific product state tensors, according to a consistent convention. I thought about this when first starting on the initialization, but including this would mean we end up with almost all of the same code. And I personally liked being able to separately instantiate the product state environment to start from as an actual object.

>  I can't really see where the `bipartite_id` function is being used immediately. If this is for initializing a trivial product environment state, I think this is what the `BPEnv` does already (?), so in that sense this would be equivalent to `CTMRGEnv(BPEnv(state))`. Is this something we want to expose/export on top of that, or should we really just an example in the docs on how to actually do this?

This I didn't actually think of, good point. I thought `BPEnv(state)` constructed random initial messages, I never noticed the `TensorKit.id!(...)` call afterwards. Since this one really is meant only for two-layer networks, it does make more sense to use that than to introduce a new exported thing. Just an example of this would be enough in principle, although it might be nice to have a plug-and-play version that is more insensitive to how `BPEnv(state)` works. Let me add something separate for this to see how it looks, and if we don't like it I'll remove it in favor of a straight `CTMRGEnv(BPEnv(state))` call.

Slightly irrelevant here, but I do find the manual `TensorKit.id!(...)` call for `BPEnv(state::Union{InfinitePEPS, InfinitePEPO}, args...; kwargs...)` a bit strange. Why not just use `isomorphism` as the default initialization function then instead of `randn`? Currently this means that `BPEnv(state)` and `BPEnv(InfiniteSquareNetwork(state))` give a very different result, which seems a bit strange to me.

---

So in the end it really comes down to the question of how we can best supply a specific product state as an initial environment for the contraction of a general no-2-layer network. I liked using a struct for this, since this could in the end absorb/replace the `BPEnv` struct which already does the same thing for a specific case.

lkdvos:
That makes sense, thanks for explaining (I also didn't know about the feature requests, so it is not always easy to keep up with all goals :)).

Probably the thing that threw me off the most is what you also noticed, which is that if we make this a struct, there is a bit of boilerplate that creeps in, at which point the question very quickly becomes why not use this also for the `BPEnv`. However, since you already have it here, might as well just continue with it if this feels more natural to you.

I also think that the specific case where you actually supply tensors is not actually covered in the code, since you don't have a constructor for that. It might be a good idea to explicitly make that constructor, and have it check that the spaces are actually consistent, i.e. the north and south match, the east and west match?

Ultimately, I think we will have to relax the `BPEnv` anyways, since you probably do also want to be able to run BP on partition functions, but I do also agree that we don't want to wait for that. I dont think there was a specific reason for the weird `id!` behavior by the way, so I'm happy to just make that always `id` immediately?

src/algorithms/ctmrg/initialization.jl
```diff
@@ -0,0 +1,93 @@
...
+        elt::Type{<:Number},
+        n::InfiniteSquareNetwork,
+        alg::ProductStateInitialization,
+    )
+    env = CTMRGEnv(ProductStateEnv(alg.f, elt, n))
+    return env
+end
+
+_CTMRGEnv(env) = CTMRGEnv(env)
+_CTMRGEnv(env::CTMRGEnv) = env
```
lkdvos: Can we just define this constructor directly? I think by convention `CTMRGEnv(::CTMRGEnv)` should construct a newly constructed object, but I don't think it hurts to have that work?

src/environments/product_state_environments.jl
```diff
@@ -0,0 +1,116 @@
...
+
+Construct a CTMRG environment with a trivial virtual space of bond dimension χ = 1 
+from the product state environment `prod_env`.
+"""
+function CTMRGEnv(prod_env::ProductStateEnv)
+    edges = map(eachindex(IndexCartesian(), prod_env)) do idx
+        return insertleftunit(insertleftunit(prod_env[idx]), 1)
+    end
+    corners = map(eachindex(IndexCartesian(), prod_env)) do _
+        return TensorKit.id(scalartype(prod_env), oneunit(spacetype(prod_env)))
```
lkdvos:
```suggestion
        return TensorKit.id(storagetype(prod_env), oneunit(spacetype(prod_env)))
```

src/environments/product_state_environments.jl
```diff
@@ -0,0 +1,116 @@
...
+# conversion to CTMRGEnv
+"""
+    CTMRGEnv(prod_env::ProductStateEnv)
+
+Construct a CTMRG environment with a trivial virtual space of bond dimension χ = 1 
+from the product state environment `prod_env`.
+"""
+function CTMRGEnv(prod_env::ProductStateEnv)
+    edges = map(eachindex(IndexCartesian(), prod_env)) do idx
+        return insertleftunit(insertleftunit(prod_env[idx]), 1)
```
lkdvos, line 111: I might be wrong here, but somehow this looks confusing to me. Does this work for both partition functions and peps? It cannot really do that right?

lkdvos (commented):
Except for the final comments, good to go for me!

src/environments/product_state_environments.jl
leburgel: I think it does work for both? Just from the fact that the tests run through it should, but if I run `insertleftunit(insertleftunit(t, 1))` on an `(N, 0)` `TensorMap` I get a `(N + 1, 1)` `TensorMap`, so this just seems to do what it should.

src/algorithms/ctmrg/initialization.jl
leburgel: Adding `CTMRGEnv(::CTMRGEnv)` kind of felt like bad practice for some reason, so I wanted to hide the bad practice with a `_`. But I don't really know why, I'm also happy to just add the constructor.

leburgel: Would we want this to essentially be a `copy`, meaning it copies the corners and edges?

src/environments/product_state_environments.jl
lkdvos: yeah I think I'm probably just tripping...

lkdvos (approved):

Issue #265, opened 2025-09-26
[Discussion] Remove support for specifying `ComplexSpace`s using integers


leburgel:
I think I brought this up at some point already, but the fact that we allow users to specify the PEPS and environment spaces as integers which are then interpreted as the dimension of a `ComplexSpace` gives quite a bit of clutter and overhead in the `InfinitePEPS` and `CTMRGEnv` constructors. Especially with the aim of adding some initialization routines that wrap the basic constructors, I would like to drop support for this and always require spaces to be specified as (matrices of) `ElementarySpace` or `ProductSpace` instances.

As far as I can tell this is barely used anywhere except in the README example, and I doubt that people really strongly rely on this. I think that we can assume that using PEPSKit means that people are (or should become) familiar enough with TensorKit spaces that this shouldn't be an issue.

Any strong opinions against this @lkdvos, @pbrehmer, @Yue-Zhengyuan?

Yue-Zhengyuan:
I also support this 👍

pbrehmer:
I feel relatively indifferent about the constructors but they do add a lot of clutter and extra maintenance work so I'm also happy to get rid of them

PR #266, opened 2025-09-26
Remove support for integer space specifiers in state and environment constructors

leburgel:
See #265.

I also tried to simplify the constructors a bit in the process by splitting out the filling of the physical and virtual spaces depending on how were provided. I at least removed a bunch of things, not sure if this actually made things simpler.

### DIFF
--- a/src/environments/ctmrg_environments.jl
+++ b/src/environments/ctmrg_environments.jl
@@ -128,19 +128,7 @@ end
 function _fill_edge_physical_spaces(
         D_north::S, D_east::S = D_north; unitcell::Tuple{Int, Int} = (1, 1)
     ) where {S <: VectorSpace}
-    return fill(D_north, unitcell), fill(D_east, unitcell)
-end
-function _fill_edge_physical_spaces(
-        Ds_north::M, Ds_east::M = Ds_north; unitcell::Tuple{Int, Int} = (1, 1)
-    ) where {M <: Matrix{<:VectorSpace}}
-    @assert size(Ds_north) == size(Ds_east) == unitcell "Incompatible size"
-    return map(ProductSpace, Ds_north), map(ProductSpace, Ds_east)
-end
-function _fill_edge_physical_spaces(
-        Ds_north::M, Ds_east::M = Ds_north; unitcell::Tuple{Int, Int} = (1, 1)
-    ) where {M <: Matrix{<:ProductSpace}}
-    @assert size(Ds_north) == size(Ds_east) == unitcell "Incompatible size"
-    return Ds_north, Ds_east
+    return fill(ProductSpace(D_north), unitcell), fill(ProductSpace(D_east), unitcell)
 end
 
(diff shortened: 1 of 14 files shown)


src/environments/ctmrg_environments.jl
```diff
@@ -76,32 +61,27 @@ of the corresponding edge tensor for each direction. Specifically, for a given s
...
 
 Each entry of the `Ds_north` and `Ds_east` matrices corresponds to an effective local space
-of the partition function, represented as a tuple of elementary spaces encoding a product
-space. This can either contain a single elementary space for the case of a partition
-function defined in terms of local rank-4 tensors, or a tuple of elementary spaces
-representing a product space for the case of a partition function representing overlaps of
-PEPSs and PEPOs.
+of the partition function, and can be represented as an `ElementarySpace` (e.g. for the case
+of a partition function defined in terms of local rank-4 tensors) or a `ProductSpace` (e.g.
+for the case of a partition function representing overlaps of PEPSs and PEPOs).
```
lkdvos:
```suggestion
for the case of a network representing overlaps of PEPSs and PEPOs).
```
Is it fair to call the PEPS overlap a partition function? Maybe this is slightly better

src/states/infinitepeps.jl
```diff
@@ -90,25 +87,30 @@ function InfinitePEPS(A::T; unitcell::Tuple{Int, Int} = (1, 1)) where {T <: PEPS
     return InfinitePEPS(fill(A, unitcell))
 end
 
+# expand PEPS spaces to unit cell size
```
lkdvos, line 90: Is there a reason to introduce these functions instead of simply using `fill(Pspace, unitcell)` directly?

lkdvos (commented):
Overall definitely looks good, I like the simplification quite a bit

src/states/infinitepeps.jl
leburgel: Specifically this one, probably not, since there's no optional arguments in the physical space specification. For the other ones, I managed to convince myself that it was a good idea to split off the expansion to a complete space specification because I kind of anticipate having to reuse this for the initialization routines for states that I plan to add in the near future. Also, since we might switch to representing sets of spaces as dedicated objects that we can dispatch on to move away from signatures like `CTMRGEnv(randn, ComplewF64, spaces...)` it kind of made sense to already split off that step.

We can always keep the plain `fill`s until we properly rework things, I just thought this was already a little bit nicer.

src/environments/ctmrg_environments.jl
leburgel: I think it's fair in principle, but it could definitely be confusing so I agree 'network' is better.

src/states/infinitepeps.jl
lkdvos: Ah I see, I missed the subtlety of the optional arguments!

PR #267, opened 2025-10-01
Fix `rotl90`, `rotr90` and `rot180` for `LocalOperator`

pbrehmer:
I stumbled upon this bug when I was trying to run SU for Hamiltonians with non-uniform physical spaces. It turns out that the rotation functions for `LocalOperator`s would only work for $1\times1$ unit cells or uniform lattices - this should fix it. (I hope my cold-ridden brain actually did the right thing to fix it but I added some tests so it should be fine, hopefully.) I also renamed the functions for index rotations to remove the type piracy.

### DIFF
--- a/src/operators/localoperator.jl
+++ b/src/operators/localoperator.jl
@@ -155,26 +155,29 @@ end
 # ----------------------
 
 # rotation of a lattice site
-# TODO: type piracy
-Base.rotl90(site::CartesianIndex{2}) = CartesianIndex(2 - site[2], site[1])
-Base.rotr90(site::CartesianIndex{2}) = CartesianIndex(site[2], 2 - site[1])
-Base.rot180(site::CartesianIndex{2}) = CartesianIndex(2 - site[1], 2 - site[2])
+# (copy logic from Base.rotl90, Base.rotr90, Base.rot180)
+siterotl90(site::CartesianIndex{2}, n::Int) = CartesianIndex(n - site[2], site[1])
+siterotr90(site::CartesianIndex{2}, n::Int) = CartesianIndex(site[2], n - site[1])
+siterot180(site::CartesianIndex{2}, nr::Int, nc::Int) = CartesianIndex(nr - site[1], nc - site[2])
 
 function Base.rotr90(H::LocalOperator)
+    nr = size(H.lattice, 1)
     lattice2 = rotr90(H.lattice)
-    terms2 = ((Tuple(rotr90(site) for site in sites) => op) for (sites, op) in H.terms)
+    terms2 = ((Tuple(siterotr90(site, nr + 1) for site in sites) => op) for (sites, op) in H.terms)
     return LocalOperator(lattice2, terms2...)
 end
 
 function Base.rotl90(H::LocalOperator)
+    nc = size(H.lattice, 2)
     lattice2 = rotl90(H.lattice)
-    terms2 = ((Tuple(rotl90(site) for site in sites) => op) for (sites, op) in H.terms)
+    terms2 = ((Tuple(siterotl90(site, nc + 1) for site in sites) => op) for (sites, op) in H.terms)
     return LocalOperator(lattice2, terms2...)
 end
 
 function Base.rot180(H::LocalOperator)
+    nr, nc = size(H.lattice)
     lattice2 = rot180(H.lattice)
-    terms2 = ((Tuple(rot180(site) for site in sites) => op) for (sites, op) in H.terms)
+    terms2 = ((Tuple(siterot180(site, nr + 1, nc + 1) for site in sites) => op) for (sites, op) in H.terms)
     return LocalOperator(lattice2, terms2...)
 end
 
--- a/src/operators/localoperator.jl
+++ b/src/operators/localoperator.jl
@@ -161,16 +161,16 @@ siterotr90(site::CartesianIndex{2}, n::Int) = CartesianIndex(site[2], n - site[1
 siterot180(site::CartesianIndex{2}, nr::Int, nc::Int) = CartesianIndex(nr - site[1], nc - site[2])
 
 function Base.rotr90(H::LocalOperator)
-    nr = size(H.lattice, 1)
+    nr = size(H.lattice, 1) + 1
     lattice2 = rotr90(H.lattice)
-    terms2 = ((Tuple(siterotr90(site, nr + 1) for site in sites) => op) for (sites, op) in H.terms)
+    terms2 = ((Tuple(siterotr90(site, nr) for site in sites) => op) for (sites, op) in H.terms)
     return LocalOperator(lattice2, terms2...)
 end
 
 function Base.rotl90(H::LocalOperator)
-    nc = size(H.lattice, 2)
+    nc = size(H.lattice, 2) + 1
     lattice2 = rotl90(H.lattice)
-    terms2 = ((Tuple(siterotl90(site, nc + 1) for site in sites) => op) for (sites, op) in H.terms)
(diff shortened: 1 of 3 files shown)


src/operators/localoperator.jl
```diff
@@ -155,26 +155,29 @@ end
 # ----------------------
 
 # rotation of a lattice site
-# TODO: type piracy
-Base.rotl90(site::CartesianIndex{2}) = CartesianIndex(2 - site[2], site[1])
-Base.rotr90(site::CartesianIndex{2}) = CartesianIndex(site[2], 2 - site[1])
-Base.rot180(site::CartesianIndex{2}) = CartesianIndex(2 - site[1], 2 - site[2])
+# (copy logic from Base.rotl90, Base.rotr90, Base.rot180)
```
lkdvos, line 169: Maybe as an additional test we should have:
```julia
@test siterotl90.(CartesianIndices((rows, cols)), rows, cols) == rotl90(collect(CartesianIndices((rows, cols)))
```

src/operators/localoperator.jl
```diff
@@ -155,26 +155,29 @@ end
...
+siterotl90(site::CartesianIndex{2}, n::Int) = CartesianIndex(n - site[2], site[1])
+siterotr90(site::CartesianIndex{2}, n::Int) = CartesianIndex(site[2], n - site[1])
+siterot180(site::CartesianIndex{2}, nr::Int, nc::Int) = CartesianIndex(nr - site[1], nc - site[2])
```
Yue-Zhengyuan: I was defining `rotl90` etc for 2D sites in such a way that you can move the rotated LocalOperator back to the correct place by a lattice translation, so it should be correct. The function `get_gateterm` exists to cancel the lattice translation. I'll try to see what goes wrong with the old implementation...

Yue-Zhengyuan: Oh I figured it out... My "lattice translation" only works for 1x1 unit cell. For larger unit cell, the translation is no longer integer multiples of the lattice basis vector.

src/operators/localoperator.jl
```diff
@@ -155,26 +155,29 @@ end
...
 
 function Base.rotr90(H::LocalOperator)
+    nr = size(H.lattice, 1) + 1
```
Yue-Zhengyuan: Can we move these `+1`'s into the siterot functions? So the meaning of the `n`'s will be more clear.

Yue-Zhengyuan: Also, I think our convention is to use capital `N` to denote the number of rows (Nr, Nrow) / columns (Nc, Ncol). But that's a minor suggestion.

src/operators/localoperator.jl
lkdvos: I would prefer to have these functions all take the number of rows and number of columns as an input, I think I'm to stupid to remember if `siterotl90` takes the rows or columns. 😉

Yue-Zhengyuan: Or we can pass `unitcell::NTuple{2, Int}` if we want to avoid "unused variables" (Julia doesn't check this, but when I was using Python the linter will complain).

pbrehmer: I now added Yue's variant where we supply the `unitcell::NTuple{2, Int}` to avoid unused variables, I like that a bit bit better :-)

src/operators/localoperator.jl
pbrehmer: Good point, the +1s are now included in the function definitions

src/operators/localoperator.jl
pbrehmer: I added similar tests which I think are doing what you meant here :)

pbrehmer:
> If we are changing the name, I would also be happy to do something like: rotate_left_90, rotate_right_90, rotate_180, and use that everywhere, but obviously that's not a requirement.

If we really want to change that everywhere, then I'd do that in a separate PR since we use `rotl90` and `rotr90` a lot throughout the code. In principle I would be open for that but I'm also fine with keeping a separate function for the `CartesianIndex` rotations. Reasons being:
- Using `rotl90` and `rotr90` for all other purposes makes a lot of sense matching the logic from `Base` and we want to avoid type piracy
- The `CartesianIndex` rotations require a different function signature where the unit cell size needs to be supplied so I feel that this is in spirit slightly different than the other `rotl90` and `rotr90` methods

test/utility/localoperator.jl
```diff
@@ -86,3 +87,48 @@ end
     tr_after = tr(last(only(H_shifted.terms)))
     @test abs(tr_before - tr_after) / abs(tr_before) < 1.0e-12
 end
+
+unitcells = [(1, 1), (2, 3), (3, 3), (4, 3)]
+@testset "Site rotations on $uc unitcell" for uc in unitcells
+    unrotated_inds = collect(CartesianIndices(uc))
+    # use reverse(uc) to account for transposing when using rotl90, rotr90 on non-square unit cells
+    rr_rotated_inds = rotr90(siterotr90.(collect(CartesianIndices(reverse(uc))), Ref(reverse(uc))))
+    ll_rotated_inds = rotr90(siterotr90.(collect(CartesianIndices(reverse(uc))), Ref(reverse(uc))))
```
lkdvos: I think these are the same, did you mean to:

```suggestion
    rr_rotated_inds = rotr90(siterotl90.(collect(CartesianIndices(reverse(uc))), Ref(reverse(uc))))
    ll_rotated_inds = rotl90(siterotr90.(collect(CartesianIndices(reverse(uc))), Ref(reverse(uc))))
```

pbrehmer: Whoops, I almost meant that. I think now it should be corrected. (`siterot` and `rot` should have to rotate in the same direction in order to cancel each other I believe)

lkdvos: Ah I see, passive vs active transformations are confusing...

test/utility/localoperator.jl
```diff
@@ -86,3 +87,48 @@ end
...
+
+    charges = [
+        U1Irrep(-1 // 2) U1Irrep(1 // 2)
+        U1Irrep(1 // 2) U1Irrep(-1 // 2)
+    ] # staggered charges to make physical spaces non-uniform
+    H₀ = j1_j2_model(ComplexF64, U1Irrep, InfiniteSquare(2, 2))
+    op_2x2 = add_physical_charge(H₀, charges)
+    @test rotl90(op_2x2) isa LocalOperator
+    @test rotr90(op_2x2) isa LocalOperator
+    @test rot180(op_2x2) isa LocalOperator
```
lkdvos: probably also a good idea to check:
```julia
@test rot180(rot180(x)) == x
@test rotl90(rotl90(x)) == rot180(x) == rotr90(rotr90(x))
@test physicalspace(rot180(x)) == rot180(physicalspace(x))
...
```
to expand a bit on the tests :)
One other thing could be to refactor the `LocalOperator` spacechecks for the lattice vs operators that are done in the constructor into a separate function and run that here?

pbrehmer: Ah very good point about using the constructor spacechecks in the tests. I'll do that and add the remaining tests tomorrow!

lkdvos: Thanks for the effort, I didn't mean to make you do a bunch of extra work for this minor fix

pbrehmer: I updated the tests but I didn't refactor the spacechecks into a separate function since that felt a bit awkward. When calling the rotation function we anyways always implicitly run these checks since we call the `LocalOperator` constructor. Putting the checks into a separate function and running them again wouldn't really test any additional property since we always call the constructor before. This is good to go from my side now :)

PR #269, opened 2025-10-02
Refactor `ctmrg_leftmove`

Yue-Zhengyuan:
This PR refactors `ctmrg_leftmove` by passing only the `ProjectorAlgorithm` instead of the entire `SequentialCTMRG` algorithm struct to the function. In addition, a corresponding `ctmrg_rightmove` is added, which is just the left move on a 180-degree-rotated network, but it's not tested here. These changes will be useful for full update #222.


Yue-Zhengyuan:
There are two places where CTMRG is needed in fast full update:

1. After updating a column of bonds, the CTMRG tensors around this column are updated with a single left/right move (this is what the "fast" is referring to).
2. After fast updating the CTMRG environments for a few rounds of evolution (updating all bonds), one may need to reconverge the environment with the complete CTMRG routine until convergence. 

The `ctm_alg` attribute in `FullUpdate` struct there is actually intended for (2). The reason I just pass `ctm_alg.projection_alg` to (1) is to allow people to use either SimultaneousCTMRG or SequentialCTMRG to reconverge the environment in (2), without the need to specify two CTMRG algorithm structs in `FullUpdate` (I did this some time ago, and find it a bit too repetitive). Though I agree that restricting `ctm_alg` to SequentialCTMRG might be better for FU, since the environment converged by the two approaches are not quite the same. In this case then indeed I can just pass the sequential `ctm_alg` to (1).

Anyway, I just want to be a bit more flexible. The name `leftmove`, `rightmove` may already give enough hint that they are an intermediate step in SequentialCTMRG.

lkdvos:
The issue is that if we get to the C4v version of CTMRG, we can then not use the `left_move` function since we can't really specialize on the CTMRG algorithm anymore. I'm happy to rethink the structure and names of these functions, but in my experience that tends to be a big undertaking that becomes a mess very quickly. I think here the algorithm really is useful for dispatch reasons so I'd keep it, even if the only thing it does is passing on the projector.

PR #274, opened 2025-10-11
Add correlator for InfinitePEPO (mixed state)

Yue-Zhengyuan:
This PR generalizes `correlator` to calculate correlation functions ⟨O(i, j)⟩ between two sites (i, j) in the same row/column in a 1-layer InfinitePEPO ρ (mixed state). Only tr(ρO) is implemented. ⟨ρ|O|ρ⟩ is not intended to be treated here. 

### To do:

- [x] Figure out the twists required for fermions.

### DIFF
--- a/src/algorithms/ctmrg/gaugefix.jl
+++ b/src/algorithms/ctmrg/gaugefix.jl
@@ -56,6 +56,9 @@ end
 
 # this is a bit of a hack to get the fixed point of the mixed transfer matrix
 # because MPSKit is not compatible with AD
+# NOTE: the action of the transfer operator here is NOT the same as that of
+# MPSKit.transfer_right; to match the MPSKit implementation, it would require a twist on any
+# index where spaces(Abar, 2:end) is dual (which we can just ignore as long as we use a random Abar)
 @generated function mps_transfer_right(
         v::AbstractTensorMap{<:Any, S, 1, N₁},
         A::GenericMPSTensor{S, N₂}, Abar::GenericMPSTensor{S, N₂},


Yue-Zhengyuan:
I can move the redefinition `MPSKit.correlator` to `src/algorithms/toolbox.jl` without changing other functions. But I don't quite want to scatter things around. BTW there are also many things in `src/algorithms/toolbox.jl` that appear worth putting into `src/algorithms/contractions`.

src/algorithms/contractions/correlator/pepo_1layer.jl
```diff
@@ -0,0 +1,161 @@
...
+    Vn, Vo = start_correlator(i, ρ, O[1], env)
+    i += CartesianIndex(0, 1)
+    for (k, j) in enumerate(js)
+        # transfer until left of site j
+        while j > i
+            Atop = env.edges[NORTH, _prev(i[1], end), mod1(i[2], end)]
+            Amid = trace_physicalspaces(ρ[mod1(i[1], end), mod1(i[2], end)])
+            Abot = env.edges[SOUTH, _next(i[1], end), mod1(i[2], end)]
+            T = TransferMatrix(Atop, Amid, _dag(Abot))
+            Vn = _transfer_left(Vn, T)
```
lkdvos: For these calls, why do you not use the same `Vn * T`? I feel like that is both a bit easier to read, while also being a bit easier to generalize. `_transfer_left` is somewhat of an implementation detail of a specific subset of the more general transfer functionality that is defined here in PEPSKit, that I would prefer not to depend on.

src/algorithms/contractions/correlator/pepo_1layer.jl
```diff
@@ -0,0 +1,161 @@
+function _transfer_left(
+        vec::GenericMPSTensor{S, 2}, d::MPSKit.SingleTransferMatrix
```
lkdvos: Can you be more concrete with the type for the transfer matrix? I don't think this is the typical code path that the transfer functions take, and I would rather stick to the other strategy of using `T * x` which lowers to `transfer_left` or `transfer_right`, where you can then intercept and specialize appropriately

src/algorithms/contractions/correlator/pepo_1layer.jl
Yue-Zhengyuan: `Vn * T` calls the MPSKit definition which uses `@plansor` and doesn't include twists for fermions. It's a bit involved (I thought) to distinguish the two cases on argument types (because they are the same). And I wanted to avoid manually adding the twists back.

src/algorithms/contractions/correlator/pepo_1layer.jl
```diff
@@ -0,0 +1,139 @@
+function MPSKit.transfer_left(
+        vec::AbstractTensorMap{T, S, 1, 2}, 
+        O::MPOTensor{S}, A::MPSTensor{S}, Ab::MPSTensor{S}
+    ) where {T, S}
+    return @tensor y[-1; -2 -3] := vec[1 2; 4] *
+        A[4 5; -3] * O[2 3; 5 -2] * conj(Ab[1 3; -1])
+end
```
Yue-Zhengyuan: Here I used a very inelegant approach: to avoid to dispatching to the following [MPSKit function](https://github.com/QuantumKitHub/MPSKit.jl/blob/69b2494b94c6b4ef4f442ab2c9ebae1e9137bcf9/src/transfermatrix/transfer.jl#L125-L127)
```julia
function transfer_left(x::MPSTensor, O::MPOTensor, A::MPSTensor, Ab::MPSTensor)
    return @plansor y[-1 -2; -3] := x[1 2; 4] * A[4 5; -3] * O[2 3; 5 -2] * conj(Ab[1 3; -1])
end
```
I changed `vec` to a ~(1, 2)~ (3, 0) tensor (instead of a (2, 1) MPS tensor). But this avoidable if we can manually (Edit: is it worth it?) resolve the needed fermion twists missing in `@plansor`, but this contraction graph is a bit convoluted for me. @lkdvos @leburgel Do you know how to do it?

src/algorithms/contractions/correlator/pepo_1layer.jl
```diff
@@ -0,0 +1,139 @@
...
+
+function MPSKit.transfer_left(
+        v::GenericMPSTensor{S, 3}, O::MPOTensor{S}, A::MPSTensor{S}, Ab::MPSTensor{S}
+    ) where {S}
+    return @tensor t[d_string -1 -2; -3] := v[d_string 1 2; 4] *
```
Yue-Zhengyuan: After figuring out the needed twists, I can change `@tensor` to `@plansor` here as well.

Yue-Zhengyuan (commented):
I have moved `correlator_horizontal`, `correlator_vertical` and the extension of `MPSKit.correlator` back to `src/algorithms/correlators.jl`. 

To really using the `V * T` syntax, I need some help in figuring out the fermion twists to be added to the planar `transfer_left`.

src/algorithms/contractions/vumps_contractions.jl
```diff
@@ -32,19 +32,6 @@ function _transfer_left(
         A[χ_NW D_N_above D_N_below; χ_NE]
 end
 
-# transfer with excited GL
-function MPSKit.transfer_left(
-        GL::GenericMPSTensor{S, 4}, O::PEPSSandwich,
-        A::GenericMPSTensor{S, 3}, Ā::GenericMPSTensor{S, 3},
-    ) where {S}
```
Yue-Zhengyuan, line 39: I moved this `transfer_left` to `src/algorithms/contractions/correlator/peps.jl` because it is not intended to be used in VUMPS, and has gave me some headaches in #232.

src/algorithms/contractions/correlator/pepo_1layer.jl
leburgel: I'm a bit confused now. Since this contraction is actually just nicely planar, I would assume `@plansor` just gives the desired result and there are no extra fermionic twists needed. I thought this was the whole point of using `@plansor`, that there's no need for manually putting extra twists to compensate for what `@tensor` is doing under the hood versus what we want it to do.

I thought the only reason for not using `@plansor` for planar contractions in PEPSKit is that it's not compatible with Zygote (for example in the gauge fixing for CTMRG). Not because it needs extra manual twists to be added. So I would think that here you can just use the MPSKit transfer functions with their `V * T` syntax without issues (unless we want to differentiate through this at some point).

Or maybe I'm totally misunderstanding, really not sure.

src/algorithms/contractions/vumps_contractions.jl
leburgel: This isn't really the point of this PR, but since everywhere else the name `MPSKit.transfer_left` really means a transfer function to be used in VUMPS, i.e. with extra twists if they're evaluated using `@tensor`, I think it's maybe not entirely natural to call this one `MPSKit.transfer_left`. Shouldn't be a blocker here, but I think this could keep confusing us in the future so it might be good to separate this from the VUMPS contraction by name as well.

src/algorithms/contractions/correlator/pepo_1layer.jl
Yue-Zhengyuan: I did some test. `@plansor` and `@tensor` are only equal in the following two cases:
```
┌-←- 4 -←-A--- -3   ┌-→- 4 -→-A--- -3
|         ↓         |         ↓
|         5         |         5
|         ↓         |         ↓
v-←- 2 -←-O--- -2   v-→- 2 -→-O--- -2
|         ↓         |         ↓
|         3         |         3
|         ↓         |         ↓
└-→- 1 -→Ab†-- -1   └-←- 1 -←Ab†-- -1
```

leburgel: They won't give the same result, that's true. I guess my question here is rather, why exactly do you want the result that `@tensor` would give here, rather than the one that `@plansor` would give? Sorry if I'm making things more confusing, but that's the part I don't fully understand here.

Since the expression you're evaluating is an overlap, I would expect that evaluating with `@tensor` (i.e. the contraction map) is the one that would require additional twists, whereas `@plansor` would give the desired result as is. But again, I'm not sure, and I might be misunderstanding what you're trying to do.

Yue-Zhengyuan: The overlap is already accounted when taking the trace over the physical indices:
```julia
Amid = trace_physicalspaces(ρ[mod1(i[1], end), mod1(i[2], end)])
```
Indeed, `trace_physicalspaces` uses `@plansor`. The remaining contraction is with the CTMRGEnv, and by our convention it does not need parity tensors to cancel out the fermion signs produced by ket-in-front-of-bra contraction.

leburgel: Wait, never mind, I think I understand what you mean. The bottom MPS tensor really isn't an adjoint of an MPS tensor here, it's just a south CTMRG edge (that's conjugated twice in total). So then I agree, this should be evaluated just with the `@tensor` contraction map (since this is what's actually used in the CTMRG algorithm), and using `@plansor` instead gives the wrong result.

I think a large part of my (and maybe also Lukas') confusion is that this really looks like an MPS transfer matrix where the bottom tensor is an adjoint and `@plansor` gives the physically desired result. I think a better solution to this in the long term is to completely separate transfer functions for CTMRG edges, where the bottom edge is not actually ever conjugated. While this approach with a `_dag` call in between seemed like a convenient shortcut at some point, I think it's really catching up with us and making things more confusing now.

The same reasoning applies to the `MPSKit.transfer_left` definition for an 'excited' left vector referenced below I think, where the quantity we want to evaluate actually just involves regular CTMRG edges, but it's forced into the same signature of an MPS-style transfer function where I would expect `@plansor` to give the desired result.

[EDIT] I didn't see the reply above before commenting. Thanks for the clarification, I indeed misunderstood what was happening here. Still, I think it might be best to just introduce separate transfer functions for expressions involving only CTMRG edge tensors without adjoints, which can then be evaluated using `@tensor` without additional twists. That would then move us away from @lkdvos' suggestion to use the `V * T` transfer matrix syntax, but forcing ourselves to go through the MPSKit routines while these are really meant for a different purpose isn't really a good idea I think.

Yue-Zhengyuan: I agree with @leburgel that here `V * T` looks nice but can easily confuse people.

lkdvos: I think I agree with most of the statements above, but I would maybe slightly alter the answer: I would much rather have a custom `<: AbstractTransferMatrix` type that embodies this difference in meaning, rather than bypassing that machinery entirely. There might be a point to make for changing the contractions in MPSKit to not include the adjoint and going from there, but that sounds like a lot of painful work

src/algorithms/contractions/vumps_contractions.jl
lkdvos: I am not sure I agree with this, how is this not the exact same transfer function? the twists relate to gauge choices on the inputs I think, but the function is still the same?

src/algorithms/contractions/correlator/pepo_1layer.jl
leburgel: Sure, it would be really nice to bring this all under the umbrella of an `AbstractTransferMatrix` approach. What I mean mostly is that right now, since `TransferMatrix` multiplication calls `MPSKit.transfer_left` which does something that's not quite what we need here, right now there's no point in forcing it. Since there no way to define a new `MPSKit.transfer_left` for what we need here based on the argument types (which are exactly the same), the only real option is to write a new implementation with a new name, which is what I mean by splitting off the implementations for transfer functions involving CTMRG edges. Of course it would be best to plug those implementations into a unified `AbstractTransferMatrix` approach.

How about we add the implementations for what we need here already and call those in the correlator implementation (so not overloading `MPSKit.transfer_left`, but maybe also not just having another `_transfer_left` method like it originally was since that name doesn't say much). The exact name doesn't matter much for now I think, as long as it's clear what it's meant to do. Then once this is working an no longer confusing here, we can work on handling CTMRG edge transfer functions in general (which also includes the excited `MPSKit.transfer_left` referenced below). How does that sound?

src/algorithms/contractions/vumps_contractions.jl
Yue-Zhengyuan: This `transfer_left` does not involve any additional twists. For comparison, the ones used by VUMPS are
```julia
function MPSKit.transfer_left(
        GL::GenericMPSTensor{S, N}, O::Union{PEPSSandwich, PEPOSandwich},
        A::GenericMPSTensor{S, N}, Ā::GenericMPSTensor{S, N},
    ) where {S, N}
    Ā = twistdual(Ā, 2:N)
    return _transfer_left(GL, O, A, Ā)
end
```
where the various `_transfer_left`'s only use `@tensor`.

leburgel: Yes, the twists just relate gauge choices on the inputs, that's true. What I mean mostly is that MPSKit usually assumes a certain gauge choice for twists on its MPS tensors, like the convention in `MPSKit.transfer_left` for a regular VUMPS on a PEPS sandwich. Making a different assumption here is very weird to me, since that means you couldn't just input the same MPS into both methods of `MPSKit.transfer_left`. To me it would make more sense to add twists here to 'adhere to the MPSKit convention', and then compensate for those explicitly when using this function on CTMRG edges which have a different gauge choice regarding twists.

But the most obvious solution to me would just be to not use this at all for CTMRG edges, since this inevitably involves taking an adjoint twice and is just needlessly confusing in retrospect.

lkdvos: Sure, but can we keep the transfer matrix object construction? I don't mean it has to be the exact same struct, I just want to retain the skeleton of the machinery, also to lower the mental load on knowing how everything is defined

Yue-Zhengyuan: Keeping a transfer matrix struct is definitely convenient. We just need to be able to easily distinguish between the ones constructed from CTM edge tensors from the ones constructed from <MPS|(MPO)|MPS>.

leburgel:
Okay, to move forward with this. While it does kind of stall progress here, I think it's not too much of a detour to properly address the transfer function confusion between MPS and CTMRG environments here. I would suggest to
- Introduce dedicated methods for transfer functions using CTMRG edge tensors (i.e. without adjoints)
   - My first idea would be to call these `edge_transfer_left` and `edge_transfer_right`, but I would be happy with any reasonable name
- Replace all instances which call `PEPSKit._dag` before `MPSKit.transfer_left/right` with direct calls to the edge transfer methods
- Double check to only call `MPSKit.transfer_left/right` on objects which adhere to the MPSKit convention for parity flips, to avoid any possible future confusion
- [Optional, at least to me for merging this PR] Wrap the application of `edge_transfer_left/right` into multiplication syntax for a `EdgeTransferMatrix <:MPSKit.AbstractTransferMatrix` (name up for debate) which would be the equivalent of `MPSKit.SingleTransferMatrix`
  - I don't think this is strictly necessary for this specific PR, but it's probably not much extra work so if it turns out to be easy we might as well add it here.
  - Unfortunately, here we can't really use the generic constructor `TransferMatrix(...)`, since the objects we would call it on are type-indistinguishable from the ones used in MPS transfer matrices. So this part would require messing around a bit with syntax and signatures

I can give this a go tomorrow morning. Having originally introduced the `_dag` shortcut, I'm happy to help dealing with the resulting fallout :)

As a side note, having to think about all of this again made me realize that probably our correlation length implementation is most likely just wrong for fermionic systems. We currently compute a transfer matrix spectrum using (in the end) `MPSKit.transfer_left`, but this is actually not compatible with the twist convention we use for the south CTMRG edges. So also there, we need to either pre-twist the south edges before passing them to `MPSKit.transfer_spectrum`, or just compute the spectrum of a proper edge transfer function in the first place.

src/algorithms/contractions/correlator/pepo_1layer.jl
```diff
@@ -0,0 +1,56 @@
...
+    ) where {T, S}
+    (size(ρ, 3) == 1) ||
+        throw(ArgumentError("The input PEPO ρ must have only one layer."))
+    r, c = Tuple(j)
+    E_north = env.edges[NORTH, _prev(r, end), mod1(c, end)]
+    E_east = env.edges[EAST, mod1(r, end), _next(c, end)]
+    E_south = env.edges[SOUTH, _next(r, end), mod1(c, end)]
+    C_northeast = env.corners[NORTHEAST, _prev(r, end), _next(c, end)]
+    C_southeast = env.corners[SOUTHEAST, _next(r, end), _next(c, end)]
+    t = twistdual(ρ[mod1(r, end), mod1(c, end)], 1:2) # TODO: is this still needed?
```
Yue-Zhengyuan: This twist is on physical indices, and necessary in the "tr" of tr(ρO).

Yue-Zhengyuan: Similarly, for PEPS correlators, I think should also be a `twistdual(ket(sandwich))` when the physical space is a dual space (which hardly happens in practice). I can modify the tests later to catch this case.

leburgel: That's a good point. I would think the same goes for any energy expectation value in principle then? So in practice we really don't support any kind of dual physical spaces at all then I guess, I never thought of that.

Yue-Zhengyuan:
There are problems when the physical space is a dual space. I'll try to solve them later.

Yue-Zhengyuan:
@lkdvos I'm now sure that either `str` or `trmul` is wrong when the physical spaces include dual spaces
```julia
using Random
using TensorKit
using PEPSKit
Random.seed!(0)
V1 = Vect[FermionParity](0 => 2, 1 => 2)
V = V1' ⊗ V1' ⊗ V1
# create a positive operator
ρ = randn(V → V)
ρ = ρ * ρ'
A = id(V)
PEPSKit.trmul(A, ρ) / PEPSKit.str(ρ) # not equal to 1
```

src/algorithms/contractions/transfer.jl
```diff
@@ -0,0 +1,178 @@
...
+    ) where {S}
+    @autoopt @tensor v´[χ_SE D_E_above D_E_below; χ_NE] :=
+        v[χ_SW D_W_above D_W_below; χ_NW] *
+        Etop[χ_NW D_N_above D_N_below; χ_NE] *
+        Ebot[χ_SE D_S_above D_S_below; χ_SW] *
+        ket(O)[d; D_N_above D_E_above D_S_above D_W_above] *
+        conj(bra(O)[d; D_N_below D_E_below D_S_below D_W_below])
+
+    return v´
+end
```
leburgel, line 86: The left (and similarly right below) transfer function for edge tensors. It mirrors `MPSKit.transfer_left` as closely as possible in terms of argument order.

We still need some decision on the naming. I'm personally fine with `edge_transfer_left`. In particular, I'm not really in favor of adding any reference to 'CTMRG' in the name. The same kind of operations also show up for different ways of computing a contraction environment (e.g. in boundary MPS for networks without spatial symmetries), so I think it should be clear it works for edges of any general contraction environment.

src/algorithms/contractions/vumps_contractions.jl
```diff
@@ -7,20 +7,20 @@ function MPSKit.transfer_left(
         A::GenericMPSTensor{S, N}, Ā::GenericMPSTensor{S, N},
     ) where {S, N}
     Ā = twistdual(Ā, 2:N)
-    return _transfer_left(GL, O, A, Ā)
+    return mps_transfer_left(GL, O, A, Ā)
```
leburgel, line 10: I systematically renamed this to `mps_transfer_left` to indicate the fact that the bottom edge is conjugated, not rotated. This should be a bit more clear than before

src/algorithms/correlators.jl
```diff
@@ -11,159 +15,68 @@ function correlator_horizontal(
...
         while j > i
-            Atop = env.edges[NORTH, _prev(i[1], end), mod1(i[2], end)]
-            Abot = env.edges[SOUTH, _next(i[1], end), mod1(i[2], end)]
+            Etop = env.edges[NORTH, _prev(i[1], end), mod1(i[2], end)]
+            Ebot = env.edges[SOUTH, _next(i[1], end), mod1(i[2], end)]
             sandwich = (
                 ket[mod1(i[1], end), mod1(i[2], end)], bra[mod1(i[1], end), mod1(i[2], end)],
             )
-            T = TransferMatrix(Atop, sandwich, _dag(Abot))
+            T = edge_transfermatrix(Etop, sandwich, Ebot)
```
leburgel, line 38: We now avoid the double adjoint on the bottom edge.

src/algorithms/ctmrg/gaugefix.jl
```diff
@@ -56,7 +56,7 @@ end
 
 # this is a bit of a hack to get the fixed point of the mixed transfer matrix
 # because MPSKit is not compatible with AD
-@generated function _transfer_right(
+@generated function mps_transfer_right(
```
leburgel, line 62: This is not quite the same as `MPSKit.transfer_right`, in the sense that it's missing some twists that would be needed to exactly match the implementation here with the `@plansor` implementation in MPSKit. At the moment we only use in the gauge fixing with a completely random bottom MPS, so in practice this doesn't matter at all. However, I can rename this one if it's confusing.

src/algorithms/toolbox.jl
```diff
@@ -240,14 +240,36 @@ end
...
+    transferspace = fuse(MPSKit._firstspace(first(top)) * MPSKit._lastspace(first(bot)))
+    num_vals = min(dim(transferspace, sector), num_vals) # we can ask at most this many values
+    eigenvals, eigenvecs, convhist = eigsolve(
+        flip(edge_transfermatrix(top, bot)), init, num_vals, :LM; tol = tol
+    )
+    convhist.converged < num_vals &&
+        @warn "correlation length failed to converge: normres = $(convhist.normres)"
+
+    return eigenvals
 end
```
leburgel, line 270: This is basically just a copy of `MPSKit.transfer_spectrum`, but using the dedicated edge transfer function instead. The main difference is that the bottom edge is a mandatory argument here (it simply can't be the same as the top since the physical spaces need to fit). There's quite a lot of duplication, but I don't know if it's worth it to really do anything about this.

src/algorithms/toolbox.jl
```diff
@@ -271,31 +293,31 @@ function _correlation_length(
 
     # Horizontal
     λ_h = map(1:n_rows) do r
-        above = InfiniteMPS(env.edges[NORTH, r, :])
-        below = InfiniteMPS(_dag.(env.edges[SOUTH, _next(r, n_rows), :]))
-        vals = MPSKit.transfer_spectrum(above; below, num_vals, sector, kwargs...)
+        top = env.edges[NORTH, r, :]
+        bot = env.edges[SOUTH, _next(r, n_rows), :]
+        vals = edge_transfer_spectrum(top, bot; num_vals, sector, kwargs...)
```
leburgel, line 298: This now uses the edge transfer function, which hopefully fixes correlation lengths for fermionic systems.

src/algorithms/transfermatrix.jl
```diff
@@ -0,0 +1,38 @@
...
+# Edge transfer matrices
+#
+
+# single site transfer
+struct EdgeTransferMatrix{A <: CTMRGEdgeTensor, B, C <: CTMRGEdgeTensor} <: MPSKit.AbstractTransferMatrix
+    top::A
+    mid::B
+    bot::C
+    isflipped::Bool
+end
```
leburgel, line 11: This needs a decision on the name, and whether it's worth it so subtype `MPSKit.AbstractTransferMatrix`. The upside is that we get some free convenience syntax regarding products of and multiplication with transfer matrices. The downside is that there's no way to dispatch to this instead of `MPSKit.SingleTransferMatrix` based on the arguments, which means that we need an explicit constructor (with yet another name) in the end anyway (see below).

src/algorithms/transfermatrix.jl
```diff
@@ -0,0 +1,38 @@
...
+
+Base.:*(tm1::T, tm2::T) where {T <: EdgeTransferMatrix} = ProductTransferMatrix([tm1, tm2])
+
+# TODO: really not sure it TensorKit.flip is the suitable method for this...
+function TensorKit.flip(tm::EdgeTransferMatrix)
```
leburgel, line 16: Not really the right place to bring this up, but I'm not sure if `TensorKit.flip` is really the right name for this. Calling `flip` on a `TensorMap` doesn't actually move any spaces around, while the intended action here is really a proper transpose.

src/algorithms/transfermatrix.jl
```diff
@@ -0,0 +1,38 @@
...
+# constructors
+edge_transfermatrix(a) = edge_transfermatrix(a, nothing, a)
+edge_transfermatrix(a, b) = edge_transfermatrix(a, nothing, b)
+function edge_transfermatrix(a::CTMRGEdgeTensor, b, c::CTMRGEdgeTensor, isflipped = false)
+    return EdgeTransferMatrix(a, b, c, isflipped)
+end
+function edge_transfermatrix(a::AbstractVector, b, c::AbstractVector, isflipped = false)
+    tot = ProductTransferMatrix(convert(Vector, edge_transfermatrix.(a, b, c)))
+    return isflipped ? flip(tot) : tot
+end
```
leburgel, line 38: The analogue of `MPSKit.TransferMatrix` (which should be de-capitalized in the near future). We need a decision on the name, I'm not too fond of this choice. I would prefer to not introduce a new name for this at all, but the calls are really dispatch-wise indistinguishable from the mirrored `MPSKit.TransferMatrix` calls, so I don't see a way around this right now.

src/algorithms/transfermatrix.jl
Yue-Zhengyuan: Is `flip` simply a label of whether the thing to be transferred is on the left (flipped) or right (not flipped) of the transfer matrix? It indeed is quite irrelevant of what `flip` does (on vector spaces) in TensorKit.

src/algorithms/contractions/transfer.jl
Yue-Zhengyuan: > I'm not really in favor of adding any reference to 'CTMRG' in the name.

I'm now OK with this, in case there are other places unknown to me that use this.

src/algorithms/contractions/correlator/pepo_1layer.jl
lkdvos: Can we remove this todo?

src/algorithms/ctmrg/gaugefix.jl
lkdvos: Could you just add this as a comment? I don't mind, but I know that I will forget this and it might be nice to be reminded if this ever becomes an issue

src/algorithms/toolbox.jl
lkdvos: I'm happy to keep this as is.

src/algorithms/transfermatrix.jl
lkdvos: This is correct, but a very annoying breaking change for MPSKit 🙃 I'm not sure why this was added in the first place, but we should indeed change it. I'm pretty sure that `transpose` is closer to what we actually mean, so if anyone feels up for changing that in MPSKit be my guest!

src/algorithms/transfermatrix.jl
lkdvos: Would it help/solve the issue if we have `transfermatrix(coords, top::CTMRGEnv, [sandwich], bot::CTMRGEnv)`?

src/algorithms/transfermatrix.jl
```diff
@@ -0,0 +1,38 @@
...
+
+# constructors
+edge_transfermatrix(a) = edge_transfermatrix(a, nothing, a)
+edge_transfermatrix(a, b) = edge_transfermatrix(a, nothing, b)
+function edge_transfermatrix(a::CTMRGEdgeTensor, b, c::CTMRGEdgeTensor, isflipped = false)
+    return EdgeTransferMatrix(a, b, c, isflipped)
+end
+function edge_transfermatrix(a::AbstractVector, b, c::AbstractVector, isflipped = false)
+    tot = ProductTransferMatrix(convert(Vector, edge_transfermatrix.(a, b, c)))
```
lkdvos, line 36: It would be nice to proactively not depend on `ProductTransferMatrix`, would something like `prod(edge_transfermatrix.(a, b, c))` work?

lkdvos (commented):
Overall looks good to me, I'm fine with the names, I dont think there are that many better options and I also don't think it matters too much. I have some final small comments, but otherwise should be good to merge for me

src/algorithms/transfermatrix.jl
leburgel: I would have been fine with calling it `flip`, just not `TensorKit.flip` :). Anyway, doesn't matter much here, just wanted to check what others thought of this.

src/algorithms/transfermatrix.jl
leburgel: I can't really think of a situation where the top and bottom edges would belong to different environments. Can't say for sure that this could never happen, but for all applications I know of you would just take opposing edges (north-sout/east-west) from the same environment.

Specifying the coordinates might be a good idea, but there's some (arbitrary?) choices to make then. We need a direction part to the coordinate to specify in which direction the transfer operator works, but do we choose this to be the direction of the 'top', or really the direction in which it is applied? Then we need some actual spatial coordinates, one of which needs to be a fixed integer and the other can be a range, depending on the direction. Also then, we would need to choose whether the spatial coordinate denotes the locations of the top, sandwich or bot. All definitely doable and it would actually solve the dispatch issue if we only construct transfer operators from `InfiniteSquareNetwork`s and `CTMRGEnv`s instead of (vectors of) individual tensors.

Since this would require some more work, it's probably better suited for a follow-up PR?

src/algorithms/transfermatrix.jl
leburgel: Not out of the box, only if we define the equivalent of this: https://github.com/QuantumKitHub/MPSKit.jl/blob/666f2d32e3c691d1b812f21ed2c059f4dce63b00/src/transfermatrix/transfermatrix.jl#L26

So that would still require an explicit call to `ProductTransferMatrix`, but if this is a better option I'm happy to change it.

src/algorithms/contractions/correlator/pepo_1layer.jl
Yue-Zhengyuan: Todo removed, twist kept.

src/algorithms/transfermatrix.jl
lkdvos: Yeah, definitely happy to put this off to a follow-up, but I think quite a few of the issues you bring up are things we now need to handle anyways, scattered around at the callsites, so this might in the end help simplifying that. (maybe)

I'm mostly just thinking out loud since this is probably also something we then do in MPSKit, where a similar function really should exist, and it would be nice to think through that a bit so we can immediately have both PEPSKit and MPSKit follow the same signature

src/algorithms/transfermatrix.jl
lkdvos: Ah, I missed that this was not defined for the `AbstractTransferMatrix`. Fine to leave as is then, no need to complicate things if it doesn't work :)

src/algorithms/transfermatrix.jl
leburgel: MPSKit has the additional annoyance that just supplying a top and bottom `InfiniteMPS` and a direction in which to act isn't even enough, because you also need to specify whether you use the left or right gauged MPS tensors from each of them. Do you think it's best to work this out in MPSKit first, with PEPSKit kind of in mind, or the other way around?

lkdvos: Ah you are right, forgot about that. I guess it might be fair to just have a keyword there though. Realistically I wont find time to immediately tackle this anyways, so probably just doing something in either package that works and then re-evaluating is fine?

Issue #277, opened 2025-10-21
Segmentation faults when running variational optimization on Julia v1.12


leburgel:
We've been having systematic CI failures since the release of Julia v1.12. Any test that calls Zygote using Julia v1.12 gives a segmentation fault. The issue seems to be exactly the same as JuliaLang/julia#59138. It was already noted that Zygote currently isn't really compatible with Julia v1.12 (see FluxML/Zygote.jl#1580), but it doesn't seem like anything happened with this since.

We could open another issue in Zygote.jl to try and get this sorted or try to work around it here, but maybe in the meantime we could restrict to Julia <1.12 to ensure users don't get any unexpected segfaults and so we can again properly assess the CI on open PRs right now.

leburgel:
Closed by #295.

PR #278, opened 2025-10-21
Restrict Julia to `<v1.12` due to Zygote incompatibility

leburgel:
Adds an upper bound for the Julia compat, until the incompatibility of Julia v1.12 with Zygote.jl is resolved. See #277.

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -24,7 +24,7 @@ jobs:
       matrix:
         version:
           - 'lts' # minimal supported version
-          - '1.11' # latest released Julia version
+          - '1.11' # TODO: update to '1' when compatible
         group:
           - ctmrg
           - boundarymps


.github/workflows/Documentation.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1' # automatically expands to the latest stable 1.x release of Julia
+          - '1.11' # automatically expands to the latest stable 1.x release of Julia
```
lkdvos:
```suggestion
          - '1.11' # TODO: update to '1' when compatible
```

leburgel:
Somehow there's still a bunch of tests for Julia 1 queued, even though the version in the tests workflow file is definitely set to 1.11 explicitly. Do you have any idea why this is @lkdvos?

PR #280, opened 2025-10-23
Improve time evolution interface

Yue-Zhengyuan:
This PR provides an improved interface for current and upcoming (Trotter-based) time evolution algorithms. 

Time evolution is now performed by an iterator, to allow easier access and finer control of the intermediate steps.
```julia
mutable struct TimeEvolver{TE <: TimeEvolution, G, S, N <: Number}
    "Time evolution algorithm (currently supported: `SimpleUpdate`)"
    alg::TE
    "Trotter time step"
    dt::N
    "The number of iteration steps"
    nstep::Int
    "Trotter gates"
    gate::G
    "PEPS/PEPO and its environment"
    state::S
end
```
- The internal state `state` stores the evolution outcome after each step, and will change during the evolution. 
- `dt` can be any real or complex number or any precision.
- The 2-site (nearest neighbor only) and 3-site (up to next nearest neighbor) version of the update are dispatched based on the type of `gate`. 

For simple update, `state` contains the following data:
```julia
struct SUState{S <: InfiniteState, E <: SUWeight, N <: Number}
    "number of performed iterations"
    iter::Int
    "evolved time"
    t::N
    "PEPS/PEPO"
    psi::S
    "SUWeight environment"
    env::E
end
```

The upper-level interface of time evolution are the following functions:
```julia
# expert mode: directly dealing with the iterator
## creating the iterator (can be used with for-loop syntax)
TimeEvolver(psi0, H, dt, nstep, alg, env0; t0 = 0.0)
## one step of iteration on an externally specified state (wrapper of Base.iterate(it, state))
MPSKit.timestep(it, psi, env) -> (ψ, env, info)
## evolve to the end without interruption
MPSKit.time_evolve(it; tol = 0.0, kwargs...) -> (ψ, env, info)

# normal usage: the iterator is handled internally (wrapper of time_evolve(it))
MPSKit.time_evolve(psi0, H, dt, nstep, alg, env0; tol = 0.0, t0 = 0.0, kwargs...) -> (ψ, env, info)
```
- `t0` records the initial time of the input state. The time of the output state is recorded in `info.t`.
- Real and imaginary time evolution are distinguished by `imaginary_time::Bool` in the `SimpleUpdate` algorithm struct. The default value is `true` (different from MPSKit).
- Convergence check is enabled by setting `tol > 0`:
    - For simple update, convergence is implied by weight difference between two steps < tol.
    - For full update, convergence is implied by energy difference between two steps < tol.
- I still don't quite want to create a separate `ground_state_search` function. (1) In practice one set the `nstep` to a very large number (of order 1e+5 to 1e+6), instead of letting it evolve indefinitely, since sometimes things don't converge well. (2) Except the need to check convergence, it is not too different from other usages. (3) It's usually not actually the ground state, and only serve as initialization for more sophisticated optimizations.

To do:

- [x] Polish the docstring.
- [x] Update the examples with the new time evolution interface. (Docs will be re-generated in a separate PR.)

### DIFF
--- a/test/examples/tf_ising_finiteT.jl
+++ b/test/examples/tf_ising_finiteT.jl
@@ -66,7 +66,7 @@ result_β = measure_mag(pepo, env)
 @test isapprox(abs.(result_β), bm_β, rtol = 1.0e-2)
 
 # continue to get results at 2β, or T = 1.25
-pepo, wts, info = time_evolve(pepo, ham, dt, nstep, alg, wts; t₀ = β)
+pepo, wts, info = time_evolve(pepo, ham, dt, nstep, alg, wts; t0 = β)
 env = converge_env(InfinitePartitionFunction(pepo), 16)
 result_2β = measure_mag(pepo, env)
 @info "tr(σ(x,z)ρ) at T = $(1 / (2β))" result_2β


Yue-Zhengyuan:
I'm going to try making time evolution an iterator (the approach of YASTN) to allow easier user access to the intermediate states.

Yue-Zhengyuan:
I now find it necessary to store Hamiltonian in TimeEvolver, since for full update it will be need to measure the energy of the state and compare with the last iteration.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -134,94 +176,125 @@ function su_iter(
...
-    if state isa InfinitePEPS
-        gate_bothsides = false
-    end
-    dt = gate_bothsides ? (alg.dt / 2) : alg.dt
-    gate = get_expham(ham, dt)
-    wtdiff = 1.0
-    env0 = deepcopy(env)
-    for count in 1:(alg.maxiter)
+function Base.iterate(it::TimeEvolver{<:SimpleUpdate}, state = it.state)
+    return LoggingExtras.withlevel(; it.verbosity) do
```
lkdvos: I think the `withlevel` construction would probably be more appropriate one level higher, in the `time_evolve` function instead. This will still be propagated to this level without explicitly passing it on as it is a scoped value.

lkdvos:
For the logging:

```julia
LoggingExtras.withlevel(f; it.verbosity)
```
runs `f() -> y` in a scope with the modified logger, and returns the output `y` of that function.
The syntax
```julia
LoggingExtras.withlevel(; it.verbosity) do
# body
end
```
is syntactically equivalent to
```
function f()
# body
end

LoggingExtras.withlevel(f; it.verbosity)
```

Therefore, if you don't return anything from within the `do` block of the `withlevel`, this follows the same syntax rules as regular function definitions and anonymous functions in Julia: the result of the last executed statement is returned. Since this is often a bit subtle to actually figure out what is being returned, we tend to favor explicitly adding the `return` statement, even though this is technically not required by the language: See also [this docs page](https://docs.julialang.org/en/v1/manual/functions/#The-return-Keyword).

---
 
I'm happy to exclude a bunch of these functions from the docs if you like, but I'd rather do that by restricting the docs, while keeping the docstrings. I find these actually very useful, and fundamentally disagree with the idea that a private internal function cannot have a docstring. In an organization where multiple people work on different parts of the code, it seems to me almost required to be able to have this.

---

For the `FU`, it is a bit harder without having an actual look at it, but perhaps something like this could work?

```julia
function Base.iterate(it::FUIterator, state = it.state)
    if state.iteration % it.environment_frequency == 0
        leading_boundary
    end
    
    # actual iterator implementation
end
```

Yue-Zhengyuan:
@lkdvos Ready for (hopefully) final review

examples/hubbard_su/main.jl
```diff
@@ -66,26 +66,24 @@ maxiter = 20000
...
 ## Computing the ground-state energy
 
 In order to compute the energy expectation value with evolved PEPS, we need to converge a
-CTMRG environment on it. We first converge an environment with a small enviroment dimension
-and then use that to initialize another run with bigger environment dimension. We'll use
-`trunc=truncrank(χ)` for that such that the dimension is increased during the second CTMRG
-run:
+CTMRG environment on it. We first converge an environment with a small enviroment dimension,
+which is initialized using the simple update bond weights. Next we use it to initialize 
+another run with bigger environment dimension. The dynamic adjustion of environment dimension
```
lkdvos:
```suggestion
another run with bigger environment dimension. The dynamic adjustment of environment dimension
```

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -1,38 +1,96 @@
...
-$(SIGNATURES)
+Internal state of simple update algorithm
+"""
+struct SUState{S <: InfiniteState, E <: SUWeight, N <: Number}
+    "number of performed iterations"
+    iter::Int
+    "evolved time"
+    t::N
+    "PEPS/PEPO"
+    ψ::S
```
lkdvos: Could you rename this property to a non-unicode symbol? I don't mind using unicode throughout the code, but in some terminals and machines this can be really annoying, so I don't want to force anyone to use this, and here the only way to access this property is through a unicode symbol.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -1,38 +1,96 @@
...
+    )
+
+Initialize a TimeEvolver with Hamiltonian `H` and simple update `alg`, 
+starting from the initial state `ψ₀` and SUWeight environment `env₀`.
+
+- The initial time is specified by `t₀`.
+"""
+function TimeEvolver(
+        ψ₀::InfiniteState, H::LocalOperator, dt::Number, nstep::Int,
+        alg::SimpleUpdate, env₀::SUWeight; t₀::Number = 0.0
```
lkdvos: Same comment about the unicode here

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -134,94 +179,139 @@ function su_iter(
...
+    timestep(
+        it::TimeEvolver{<:SimpleUpdate}, ψ::InfiniteState, env::SUWeight;
+        iter::Int = it.state.iter, t::Float64 = it.state.t
+    ) -> (ψ, env, info)
+
+Given the TimeEvolver iterator `it`, perform one step of time evolution
+on the input state `ψ` and its environment `env`.
+
+- Using `iter` and `t` to reset the current iteration number and evolved time
+    respectively of the TimeEvolver `it`.
```
lkdvos: Do we use this functionality anywhere? I'm not sure it adds that much to give `timestep` the responsibility of resetting the iteration number and evolved time.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -134,94 +179,139 @@ function su_iter(
...
+    including the time `info.t` evolved since `ψ₀`.
 """
-function simpleupdate(
-        state::InfiniteState, ham::LocalOperator, alg::SimpleUpdate, env::SUWeight;
-        bipartite::Bool = false, gate_bothsides::Bool = true,
-        force_3site::Bool = false, check_interval::Int = 500
+function MPSKit.time_evolve(
+        ψ₀::InfiniteState, H::LocalOperator, dt::Number, nstep::Int,
+        alg::SimpleUpdate, env₀::SUWeight;
+        tol::Float64 = 0.0, t₀::Number = 0.0, check_interval::Int = 500
```
lkdvos: same comment about unicode kwarg name

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: My logic was that since the initial state `ψ` and `env` are arbitrary in principle, `iter` and `t` should also be changed to match the input state. 

Sometimes I indeed need to reset the internal state. But then I evolve to the end, not only one step, so I didn't reset it via `timestep`. See the [real time full update test](https://github.com/Yue-Zhengyuan/PEPSKit.jl/blob/79ae4ff7fc04b1bcafdbdf059b8255eb51192b25/test/timeevol/tf_ising_realtime.jl) in #222. Besides this, frankly nothing uses `timestep` to reset the iterator.

Yue-Zhengyuan: Do you recommend adding a dedicated function to reset the internal state of TimeEvolver? Or we should just recreate the iterator when needed (the cost is to recreate the Trotter gates).

lkdvos: Is it a matter of just setting `it.t, it.iter = newt, newiter`? I'm definitely okay with adding a function for this, but I'd be fine with just having this be something manual too. I think this is based on me not seeing a large difference in resetting it vs not resetting it, I guess only `mod(iter, check)` is relevant for verifying convergence in some cases but otherwise it doesn't alter too much to just let this increase, right?

Yue-Zhengyuan: Then let's just let users do the resetting manually, e.g. with `Accessors.@set`, or if they want, directly construct the un-exported `PEPSKit.SUState`.

BTW, `iter` and `t` are recorded in `it.state`, instead of directly in `it`.

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: I'll simply change ψ to psi, and the subscript 0 to normal 0.

pbrehmer:
Thanks for the effort, this is really nice! I can regenerate the docs examples tomorrow (unless someone does it before me :)) so that we can tag a new release.

PR #291, opened 2025-11-16
Correct evolved time for bipartite simple update

Yue-Zhengyuan:
This PR fixes the evolved time to `dt` in one iteration of simple update for bipartite states.

Previously, the `exp(-H dt)` gate is applied twice to the each bond in one iteration of bipartite SU, making the evolved time `2 * dt`.

### DIFF
--- a/test/timeevol/cluster_projectors.jl
+++ b/test/timeevol/cluster_projectors.jl
@@ -101,11 +101,12 @@ end
 @testset "Hubbard model with 2-site and 3-site SU" begin
     Nr, Nc = 2, 2
     ctmrg_tol = 1.0e-9
-    Random.seed!(100)
+    Random.seed!(1459)
     # with U(1) spin rotation symmetry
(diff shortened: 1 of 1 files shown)


lkdvos (approved):

Yue-Zhengyuan:
The macOS lts check that time out seems to be just a glitch. Windows and ubuntu checks are normal.

Shall we consider moving the finite-T SU tests (`test/examples/j1j2_finiteT.jl` and `tf_ising_finiteT.jl`) from the `examples` group to `timeevol` group to even out the running time of each test group?

lkdvos:
I'm happy to do so, but I would in the long run like to find some time to re-evaluate the tests in general. While it is definitely useful to include some of the examples and full runs as tests, it would be far superior to have smaller unit tests that don't take this long, or are less flaky/rng dependent. I have some ideas, just haven't found the time yet. I'll manually merge this PR though, since that would be better in a different PR anyways.

PR #295, opened 2025-11-20
Attempt to fix Julia v1.12

leburgel:
Gets rid of the segfaults that Zygote was throwing under Julia v1.12. It seems like the fix was a lot less scary than the error. Based on Jutho's digging here JuliaLang/julia#59138, I just put all the `@warn`s we have inside an `ignore_derviatives`. Turns out there was really only one stray `@warn`, and explicitly ignoring it seems to have solved the problem. Still, Zygote is having all kinds of issues for Julia v1.12 (that we now no longer seem to suffer from), and it doesn't seem like any of these will be fixed soon (if at all).

While this all feels a bit sketchy, since we're still quite a way from moving to a different AD engine I suggest to merge this and re-enable Julia v1.12 if tests pass here. If nothing else it will make tagging a whole lot easier, since the automerge decisions on the General registry are now running on Julia v1.12. In fact, my only real motivation for this PR is to fix the [registration of v0.7.0](https://github.com/JuliaRegistries/General/pull/142685) which can't be automerged because we restricted the Julia version.

If we agree and this is merged, I'll update the registration PR to get v0.7.0 released.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -51,7 +51,6 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 - BoundaryMPS methods now have their own custom transfer functions, avoiding a double conjugation and twist issues for fermions
 - `physicalspace` and related functions now correctly handle periodic indexing for infinite networks
 - Updated compatibility with TensorKit v0.15
-- Restrict Julia to `<v1.12` due to Zygote incompatibility
 - Runic formatter
 
 ### Removed


Yue-Zhengyuan:
The 1.11 tests are still listed in the pending checks as "Expected". Will they automatically go away, or we need some manual clean-up?

Other than this I'm OK to merge the PR.

Yue-Zhengyuan (approved):

PR #311, opened 2025-12-05
Compress 2-layer InfinitePEPO with local truncation

Yue-Zhengyuan:
This PR introduces the *most naive* algorithm to compress two 1-layer iPEPOs into a new 1-layer one with smaller virtual bond dimension:
```
compress((ρ1, ρ2), alg::LocalTruncation)
```
It is a polished version of @sanderdemeyer's work in ClusterExpansions.jl, with the extra support for arbitrary unit cell sizes (ClusterExpansions.jl focused on the trivial unit cell case). At least it can serve as initialization for other advanced approximation methods. 

The algorithm minimizes the local cost function at each virtual bond, which is the norm of
```
        ↓ ╱      ↓ ╱            ↓ ╱                 ↓ ╱
    ----A2---←---B2---      ----A2-←-|╲       ╱|--←-B2---
      ╱ |      ╱ |            ╱ |    | ╲     ╱ |  ╱ |
        ↓        ↓       -      ↓    |P1├-←-┤P2|    ↓
        | ╱      | ╱            | ╱  | ╱     ╲ |    | ╱
    ----A1---←---B1---      ----A1-←-|╱       ╲|--←-B1---
      ╱ ↓      ╱ ↓            ╱ ↓                 ╱ ↓
```
The projectors are found following steps described in Appendix A, [1905.02351](https://arxiv.org/abs/1905.02351) (*Boundary Tensor Renormalization Group*). To handle fermions, the virtual arrows are required to follow the leftwards/downwards directions.

In addition, as suggested by @leburgel, I introduced lazy multiplication of two iPEPOs `ρ1 * ρ2` by stacking layers of `ρ1` *below* `ρ2` without doing any contraction. Since we restrict all codomain physical spaces to be non-dual, we don't need to worry about fermionic twists here.

An immediate generalization is truncating the product of an InfinitePEPO and an InfinitePEPS, but currently I don't have concrete test cases.


sanderdemeyer:
Cool! 

What you could do in the PEPS-PEPO case is to construct the PEPO as $$ exp(-d\tau H) $$, and see whether many applications of this PEPO on a random initial PEPS make it converge to the ground state. That is what I have as one of the tests in my ClusterExpansions code, which you can find [here](https://github.com/sanderdemeyer/ClusterExpansions/blob/main/test/ground_state_search.jl). This PEPO could be made using [the Trotter decomposition](https://github.com/sanderdemeyer/ClusterExpansions/blob/main/src/models/models_Trotter.jl), or via Simple Update, like you did in the previous tests. This should also allow for a fermionic test case.

Yue-Zhengyuan:
@sanderdemeyer I don't know if you knew it before, but recently someone mentioned to me that we can optimize the HOTRG projectors (in your terminology, the "oblique projectors" in "no-environment truncation") with automatic differentiation, which is equivalent to second RG (see [this paper](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.101.220409)). If you haven't tried it before, maybe I can help test this idea (also as an exercise to learn AD for me), which is also a great extension to this PR to actually make it appealing.

Yue-Zhengyuan:
From the experience of [2508.05406](https://arxiv.org/abs/2508.05406) and [2602.22113](https://arxiv.org/abs/2602.22113v1), this simple truncation method can already yield some decent results. So we may just merge this after settling on some technical details like the interface.

src/algorithms/approximate/approx_tools.jl
```diff
@@ -0,0 +1,62 @@
+"""
+$(TYPEDEF)
+
+Abstract super type for the truncation algorithms of
+virtual spaces in InfinitePEPO or InfinitePEPS.
```
leburgel: I don't know if "truncation algorithms for virtual spaces" is the most clear description, since we're really approximating networks. In particular, we would use similar algorithms to enlarge virtual space (approximating a network by another one with a larger bond dimension, for example for initialization).

Maybe we can just say "approximation algorithms for two-dimensional tensor networks"?

src/algorithms/approximate/approx_tools.jl
```diff
@@ -0,0 +1,62 @@
...
+"""
+abstract type ApproximateAlgorithm end
+
+function _check_virtual_dualness(state::Union{InfinitePEPS, InfinitePEPO})
+    Nr, Nc = size(state)
+    if isa(state, InfinitePEPO)
+        @assert size(state, 3) == 1
```
leburgel: I would add an error message here, since there's no intrinsic reason why this should be restricted to 1-layer PEPOs. It's fine to keep the restriction for now, but it's always good to record why in case people run into this.

src/algorithms/approximate/approx_tools.jl
```diff
@@ -0,0 +1,62 @@
...
+    end
+    flip_xs = map(Iterators.product(1:Nr, 1:Nc)) do (r, c)
```
leburgel: I personally prefer "horizontal" over "x" (and similarly "vertical" over "y"), or even "east" and "north" to keep things more in line with virtual space directions. But not a very strong preference, if this is more in line with the time evolution code for example I'm fine with keeping this.

src/algorithms/approximate/approx_tools.jl
```diff
@@ -0,0 +1,62 @@
...
+        return !isdual(virtualspace(state[r, c], EAST))
+    end
+    flip_ys = map(Iterators.product(1:Nr, 1:Nc)) do (r, c)
+        return !isdual(virtualspace(state[r, c], NORTH))
+    end
+    return flip_xs, flip_ys
+end
+
+function _flip_virtual_spaces!(
```
leburgel: Maybe it would be good to also have an out-of-place version?

src/algorithms/approximate/approx_tools.jl
```diff
@@ -0,0 +1,62 @@
...
+        inv::Bool = false
+    )
+    Nr, Nc = size(flip_xs)
+    for r in 1:Nr, c in 1:Nc
+        inds = findall(
+            [
+                flip_ys[r, c], flip_xs[r, c],
+                flip_ys[_next(r, Nr), c], flip_xs[r, _prev(c, Nc)],
+            ]
+        ) .+ (isa(state, InfinitePEPS) ? 1 : 2)
```
leburgel: Can we split off `isa(state, InfinitePEPS) ? 1 : 2` into a dedicated `num_physical_ind` method (or something similar)? That way it's clearer that that's what's being probed here, and also there's probably other places where this an be reused.

src/algorithms/approximate/approx_tools.jl
```diff
@@ -0,0 +1,62 @@
...
+        state.A[r, c] = flip(state.A[r, c], inds; inv)
```
leburgel:
```suggestion
        state[r, c] = flip(state[r, c], inds; inv)
```
`setindex!` should just work, and is nicer than manually accessing the fields I think.

src/algorithms/approximate/approx_tools.jl
```diff
@@ -0,0 +1,62 @@
...
+"""
+Flip all north and east virtual spaces of `state` to non-dual spaces.
+A new state is constructed.
+"""
+function standardize_virtual_spaces(
+        state::Union{InfinitePEPS, InfinitePEPO}; inv::Bool = false
+    )
+    flip_xs, flip_ys = _check_virtual_dualness(state)
+    !all(flip_xs) && !all(flip_ys) && (return state)
+    return _flip_virtual_spaces!(deepcopy(state), flip_xs, flip_ys; inv)
```
leburgel: This could then use the out-of-place version. Also, why the `deepcopy`? Shouldn't `copy` be enough? If not, we should probably just change our `copy` implementation instead.

src/algorithms/approximate/local_approx.jl
```diff
@@ -0,0 +1,127 @@
+struct LocalApprox <: ApproximateAlgorithm
```
leburgel: How about writing `LocalApproximate` in full? Also, I think it would be good to add a docstring here.

src/algorithms/changebonds/local.jl
```diff
@@ -0,0 +1,127 @@
+struct LocalApprox <: ApproximateAlgorithm
+    trunc::TruncationStrategy
+end
+
+"""
+Calculate the QR decomposition of 2-layer PEPO tensor
```
leburgel: This doesn't really say anything about the virtual spaces that are being contracted in the sandwich before computing the QR. Updating the diagram is probably a bit tricky, but maybe we could explain more clearly in words what's actually being decomposed?

src/algorithms/approximate/local_approx.jl
```diff
@@ -0,0 +1,127 @@
...
+function qr_twolayer(A1::PEPOTensor, A2::PEPOTensor)
+    @assert isdual(virtualspace(A1, EAST)) && isdual(virtualspace(A2, EAST))
+    A2′ = twistdual(A2, [2, 3, 5, 6])
+    A1′ = twistdual(A1, [1, 3, 5, 6])
+    @tensoropt MdagM[x2 z z′ x2′] :=
+        conj(A2[z z2; Y2 x2 y2 X2]) * A2′[z′ z2; Y2 x2′ y2 X2]
+    @tensoropt MdagM[x1 x2; x1′ x2′] := MdagM[x2 z z′ x2′] *
+        conj(A1[z1 z; Y1 x1 y1 X1]) * A1′[z1 z′; Y1 x1′ y1 X1]
+    # avoid small negative eigenvalues due to numerical errors
+    D, R = eigh_trunc!(MdagM; trunc = trunctol(; atol = 1.0e-16))
```
leburgel: Probably adding the `trunctol` here is unavoidable, but I think it would be better to control this with some global value in `Defaults` rather than hardcoding it here. Also, would an `rtol` not be safer, for tensors with non-standard normalizations?

src/algorithms/approximate/local_approx.jl
```diff
@@ -0,0 +1,127 @@
...
+function lq_twolayer(A1::PEPOTensor, A2::PEPOTensor)
+    @assert !isdual(virtualspace(A1, WEST)) && !isdual(virtualspace(A2, WEST))
+    A2′ = twistnondual(A2, [2, 3, 4, 5])
+    A1′ = twistnondual(A1, [1, 3, 4, 5])
+    @tensoropt MMdag[x2 z z′ x2′] :=
+        A2[z z2; Y2 X2 y2 x2] * conj(A2′[z′ z2; Y2 X2 y2 x2′])
+    @tensoropt MMdag[x1 x2; x1′ x2′] := MMdag[x2 z z′ x2′] *
+        A1[z1 z; Y1 X1 y1 x1] * conj(A1′[z1 z′; Y1 X1 y1 x1′])
+    # avoid small negative eigenvalues due to numerical errors
+    D, L = eigh_trunc!(MMdag; trunc = trunctol(; atol = 1.0e-16))
```
leburgel: We can then reuse the same default here, makes it easier to change if we ever have to.

src/algorithms/approximate/local_approx.jl
```diff
@@ -0,0 +1,127 @@
...
+    @assert isdual(virtualspace(A1, EAST)) && isdual(virtualspace(A2, EAST))
+    A2′ = twistdual(A2, [2, 3, 5, 6])
+    A1′ = twistdual(A1, [1, 3, 5, 6])
+    @tensoropt MdagM[x2 z z′ x2′] :=
+        conj(A2[z z2; Y2 x2 y2 X2]) * A2′[z′ z2; Y2 x2′ y2 X2]
+    @tensoropt MdagM[x1 x2; x1′ x2′] := MdagM[x2 z z′ x2′] *
+        conj(A1[z1 z; Y1 x1 y1 X1]) * A1′[z1 z′; Y1 x1′ y1 X1]
+    # avoid small negative eigenvalues due to numerical errors
+    D, R = eigh_trunc!(MdagM; trunc = trunctol(; atol = 1.0e-16))
+    R = sdiag_pow(D, 0.5) * R'
```
leburgel: Not really important, but since there is a docstring it might be better to state explicitly that `sqrt(Q) * R'` is returned.

src/algorithms/changebonds/local.jl
```diff
@@ -0,0 +1,127 @@
...
+    @tensoropt MdagM[x1 x2; x1′ x2′] := MdagM[x2 z z′ x2′] *
+        conj(A1[z1 z; Y1 x1 y1 X1]) * A1′[z1 z′; Y1 x1′ y1 X1]
+    # avoid small negative eigenvalues due to numerical errors
+    D, R = eigh_trunc!(MdagM; trunc = trunctol(; atol = 1.0e-16))
+    R = sdiag_pow(D, 0.5) * R'
+    return R
+end
+
+"""
+Calculate the LQ decomposition of 2-layer PEPO tensor
```
leburgel: Same comment about explaining what partial contraction of the sandwich is actually being decomposed.

src/algorithms/approximate/local_approx.jl
```diff
@@ -0,0 +1,127 @@
...
+    P1 = R2 * vh' * sinv_sqrt
+    P2 = sinv_sqrt * u' * R1
+    return P1, P2, s, ϵ
+end
+
+"""
+Compute an approximation to the product of two 1-layer InfinitePEPOs `ρ1`, `ρ2`
+with virtual bond truncated with `LocalApprox`.
+"""
+function MPSKit.approximate(ρ1::InfinitePEPO, ρ2::InfinitePEPO, alg::LocalApprox)
```
leburgel: This signature is a bit tricky I think. Here, it makes total sense to do it like this. However, for other uses of `approximate` (which are not implemented yet, but we probably should soon), things would probably look like
```julia
approximate(n0, to_approximate, alg::ApproximateAlgorithm, optional_args...)
```
where `n0` is some initial guess specifying the desired structure (e.g. PEPS with the virtual spaces you want), `to_approximate` is the thing you want to approximate (e.g. a PEPO and a PEPS, representing the application of the PEPO to the PEPS), and then your algorithm and any other things you might potentially need. Here, all of the structure of `n0` is in the `trunc` field of the `LocalApprox` algorithm, so there's no need for that one. Then `to_approximate = (ρ1, ρ2)` can just be added as two separate arguments.

I think this is fine as it is, but just to say that we might have to add quite different signatures for different algorithms. That being said, since there's really no `n0` here, it would be very artificial to match the example signature I gave here (it would become `approximate(trunc, (ρ1, ρ2), alg)`, which is a bit ridiculous).

So I don't really know how to do this better for now, just wanted to make note of this.

src/algorithms/approximate/approx_tools.jl
leburgel: Is it correct that this is not tested right now? If so it might be good to add a test with non-standard virtual space arrows. Otherwise, we might as well error out for non-standard arrows for now, no?

Yue-Zhengyuan: I'll go the simple way and just enforce standard virtual arrow directions.

src/algorithms/approximate/local_approx.jl
Yue-Zhengyuan: Also adding to this list: one may also want to approximate the inner product of two PEPSs (or PEPOs regarded as purified PEPSs), obtaining a partition function with a moderate bond dimension. This is encountered when one wants to feed it into TNR.

leburgel: Good point. If there's no good general solution we can just continue with what you suggested, but maybe we can discuss how we should approach this in more detail soon.

Thinking about it a bit more, the case implemented here and the the ones we both brought up feel like some kind of mixture of `MPSKit.approximate` (implying some variational approximation) and `MPSKit.changebonds` (implying a pure bond dimension truncation or expansion). In that sense, this use case could be written as
```julia
MPSKit.changebonds(ρ1 * ρ2, alg)
```
where we can perform the multiplication lazily as a stacking of the PEPO unit cells. That might actually be the most fitting I think. However, if we have too much trouble fitting the use cases here into the MPSKit algorithms and their usual signatures, we can always consider moving away from those altogether.

src/algorithms/approximate/local_approx.jl
Yue-Zhengyuan: If we are confident about the implementation (even for fermions about the twists), we can actually manually set all negative eigenvalues in `D.data` to 0 similar to `positive_approx`. In this way a truncation tolerance can be avoided.

src/algorithms/approximate/approx_tools.jl
Yue-Zhengyuan: I was playing it safe. There are a bunch of `deepcopy`'s in the repo elsewhere. We can open another PR later to try changing all of them to `copy`.

src/algorithms/approximate/approx_tools.jl
Yue-Zhengyuan: I'll iterate over `CartesianIndices(state.A)` then, supporting multi-layer PEPOs.

src/algorithms/approximate/approx_tools.jl
Yue-Zhengyuan: I prefer shorter names, and `x`, `y` are also used in `SUWeight` related code.

src/algorithms/approximate/local_approx.jl
Yue-Zhengyuan: What's returned is indeed `R` itself (think about it :) and similarly for `L` below). Sorry for confusing you with reused variable names.

Yue-Zhengyuan: Though I admit I can still be more rigorous, since the `R` here is not guaranteed to be upper triangular, similar to the `R` obtained from `left_orth` with a non-QR algorithm.

src/algorithms/changebonds/local.jl
Yue-Zhengyuan: I'll say "with the east virtual legs transferred to the `R` tensor"?

src/algorithms/changebonds/local.jl
Yue-Zhengyuan: Similarly, "with the west virtual legs transferred to the `L` tensor".

src/algorithms/changebonds/local.jl
leburgel: Yes, that would be better.

src/algorithms/approximate/local_approx.jl
leburgel: Do we have an appropriate test to check this, so we can then just manually discard all negative eigenvalues? If this is too much effort that's fine for now, I would just like to keep track of hardcoded tolerances as much as possible by collecting them all in the same place.

src/algorithms/approximate/local_approx.jl
leburgel: I think it's fine, even if `R` is not actually upper triangular, the intention. I just didn't notice that you were doing an `eigh` instead of a `qr`, I wasn't paying attention and just assumed because of the function name. So never mind the comment :)

src/algorithms/approximate/local_approx.jl
Yue-Zhengyuan: In TNRKit (in the context of 3D HOTRG) I have a test that uses random fermionic PEPOTensors to ensure that no negative eigenvalues are produced due to wrong twists. I can port that here.

src/algorithms/approximate/approx_tools.jl
Yue-Zhengyuan: Adopted your phrasing.

src/algorithms/approximate/local_approx.jl
Yue-Zhengyuan: I took a look at MPSKit. There is also the syntax
```
changebonds(ψ, H, alg, ...)
```
which I think is optimizing the MPO-MPS product `H * ψ`. I guess it is also OK to directly use `changebonds(ρ1, ρ2, alg)` without even performing the lazy stacking first?

Yue-Zhengyuan:
I'm curious why the `gradients` test needs so much more time (and allocates much more memory) when running with `min` (Julia v1.10; around 110 min) than with v1.12 (around 40 min). There's not such a huge gap for other test groups.

Yue-Zhengyuan:
@lkdvos Do you have time to take a look? I have changed from overloading `approximate` to `changebonds`. If this still looks weird, I can give up overloading MPSKit functions and just call it `compress`.

src/algorithms/changebonds/local.jl
```diff
@@ -0,0 +1,163 @@
...
+    ----A2-←-           ┌-←-
+      ╱ |               |
+        ↓     =  (Q)-←--R
+        | ╱             |
+    ----A1-←-           └-←-
+      ╱ ↓
+```
+Only `R` is calculated and returned.
+"""
+function qr_twolayer(A1::PEPOTensor, A2::PEPOTensor)
```
lkdvos: This might be a bit nitpicky, but can you maybe change the name of these functions, since you aren't actually performing a QR, but rather some form of square root?

Just out of curiosity, do you have any idea how this compares to a BP-style or SU-style approach? I can't help but feel like this is some rank-1 approximation of the double-layer object, but somehow in a symmetrized/hermitianized manner. (This is completely irrelevant for this PR, so feel free to ignore, I'm just curious)

src/algorithms/changebonds/local.jl
```diff
@@ -0,0 +1,163 @@
...
+    D, R = eigh_full!(MdagM)
+    project_psd!(D)
+    R = sdiag_pow(D, 0.5) * R'
+    return R
+end
+function _get_MdagM(A1::PEPOTensor, A2::PEPOTensor)
+    @assert isdual(virtualspace(A1, EAST)) && isdual(virtualspace(A2, EAST))
+    A2′ = twistdual(A2, [2, 3, 5, 6])
+    A1′ = twistdual(A1, [1, 3, 5, 6])
+    @tensoropt MdagM[x2 z z′ x2′] :=
```
lkdvos:
```suggestion
    @tensor MdagM[x2 z z′ x2′] :=
```
Given that there are only two tensors involved, there is no need to determine the best contraction order here :)

On a separate note, do you think it is possible to write this as a single `@tensor` call? This makes it easier to play around with allocation stuffafterwards, since the intermediate MdagM isn't required, and the `@tensor` call will ensure that the intermediate permutations are minimized.

src/algorithms/changebonds/local.jl
```diff
@@ -0,0 +1,163 @@
...
+    ----A1-←-           └-←-
+      ╱ ↓
+```
+Only `R` is calculated and returned.
+"""
+function qr_twolayer(A1::PEPOTensor, A2::PEPOTensor)
+    MdagM = _get_MdagM(A1, A2)
+    D, R = eigh_full!(MdagM)
+    project_psd!(D)
+    R = sdiag_pow(D, 0.5) * R'
```
lkdvos: I think here you probably want to merge the `sqrt` and `max(0, ...)` in a single pass, so maybe something like this could work?


```suggestion
    z = zero(eltype(D))
    D.data .= ifelse.(D.data .> z, sqrt.(z), z)
    lmul!(D, R')
```

src/algorithms/changebonds/local.jl
```diff
@@ -0,0 +1,163 @@
+"""
+$(TYPEDEF)
+
+Abstract super type for algorithms to change virtual bonds
+in two-dimensional tensor networks
+"""
+abstract type BondChangeAlgorithm end
```
lkdvos: Do we need this abstract supertype? I don't think this is really used anywhere now?

lkdvos (commented):
I'm still not entirely convinced by trying to fit it in `changebonds`, but I don't think it matters all that much to be honest. I think I would lean in the direction of just giving it a separate name, but if you prefer to keep it like this that definitely also works for me!

I left some other comments through the code that are mostly minor, otherwise I think this is probably fine!

src/algorithms/changebonds/local.jl
Yue-Zhengyuan: This is introduced mainly to be future proof. But I can remove it now that we stop overloading MPSKit functions.

src/algorithms/changebonds/local.jl
Yue-Zhengyuan: Sure, I can write it in a single pass. Then four tensors are involved, which I guess will need `@tensoropt`?

src/algorithms/changebonds/local.jl
Yue-Zhengyuan: > change the name of these functions

This is a bit like https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/issues/109, where I wanted to do `left/right_orth` but only need `L/R`, and don't care which algorithm is used to obtain `Q`. Maybe just call it `left_orth_twolayer`, or do you have any suggestions in the context of MAK?

> how this compares to a BP-style or SU-style approach

I have to say I don't have a clear answer... One thing this can do but BP/SU cannot is 1x1 unit cell.

> rank-1 approximation

Naively I can't see where a `truncrank(1)` is involved. What do you mean exactly?

src/algorithms/changebonds/local.jl
Yue-Zhengyuan: On second thought, I still want to keep it, but will rename it to `CompressAlgorithm`.

src/algorithms/changebonds/local.jl
lkdvos: That, or explicit brackets, or even just the `@autoopt` machinery, since that would fill in the sizes a bit more

src/algorithms/changebonds/local.jl
lkdvos: I'm not sure I see what makes it future-proof, we don't really have any functionality that can be written over generic `CompressAlgorithm`s right? If you want an abstract supertype, it might be nice to just use the `AbstractAlgorithm` from MPSKit?

src/algorithms/changebonds/local.jl
lkdvos: I think the idea I had is that this feels like the surroundings of the local bond are approximated by some kind of local tensor, i.e. there is no edge between the `qr_twolayer` and `lq_twolayer` part, but it might also just be a bit of a stretch so this probably doesn't buy us much.

How about `reduced_twolayer_pepo_left/right`? Maybe `twolayer_left_residual`, to indicate it's only the `R` factor?

It almost looks like a square root too, so I guess this is a cholesky decomposition? Honestly naming things is hard, but calling it `qr` or `lq` seems a bit misleading.
For the MatrixAlgebraKit issue, I have to admit that this slipped my attention a bit, for which I apologize. There as well, it is kind of hard to come up with a name, and there is not too much to gain from not forming the `Q` I think.

src/algorithms/changebonds/local.jl
Yue-Zhengyuan: I changed to `@autoopt` and added `d/D` to the contraction labels.

src/algorithms/changebonds/local.jl
Yue-Zhengyuan: Below is copied from Gemini. `A` below is the two-layer `A1 A2`, and `M` is the `MdagM` tensor.

----

### 1. The Structural Difference (Full vs. Triangular)

When you compute $M = A^\dagger A$, $M$ is a Hermitian positive semi-definite matrix.

* **Cholesky Decomposition** factors $M$ explicitly as $M = R_{chol}^\dagger R_{chol}$, where $R_{chol}$ is strictly an **upper triangular** matrix. (This $R_{chol}$ is identical to the $R$ you would get if you did a standard QR decomposition directly on $A$).
* **Your Eigendecomposition Method** factors $M = U^\dagger D U$ and defines $R_{eig} = \sqrt{D} U$. However, $R_{eig}$ is generally a **full matrix**, not an upper triangular one.

Mathematically, because both matrices satisfy $R^\dagger R = M$, they are related by a unitary transformation. There exists some unitary matrix $V$ such that $R_{chol} = V R_{eig}$.

### 2. Connection to Polar Decomposition

While standard QR yields the Cholesky factor, your eigendecomposition method is actually computing a variant of the **Polar Decomposition**.

In polar decomposition, $A = W P$, where $W$ is an isometry ($W^\dagger W = I$) and $P$ is a unique Hermitian positive semi-definite matrix given by $P = \sqrt{A^\dagger A} = \sqrt{M}$.

Using your eigendecomposition $M = U^\dagger D U$, the matrix square root is:


$$P = U^\dagger \sqrt{D} U$$

Since you defined your $R$ as $\sqrt{D} U$, we can see that $P = U^\dagger R$. This means your $R$ matrix is exactly the positive-definite polar factor $P$, just rotated by the unitary eigenvector matrix $U$.

src/operators/infinitepepo.jl
```diff
@@ -258,3 +258,17 @@ function _stack_tuples(A::Matrix{NTuple{N, T}}) where {N, T}
     end
     return out
 end
+
+"""
+Checks if `state` follow the standard virtual arrow convention,
+i.e. north and east domains are non-dual spaces.
+"""
+function _check_virtual_dualness(state::Union{InfinitePEPS, InfinitePEPO})
```
leburgel: It's a bit strange to have an `InfinitePEPS` method in this file. I would be fine with just adding this to src/algorithms/compress/local.jl directly. If we end up needing it elsewhere, we can always move it.

src/operators/infinitepepo.jl
```diff
@@ -258,3 +258,17 @@ function _stack_tuples(A::Matrix{NTuple{N, T}}) where {N, T}
...
+    isdual_easts = map(CartesianIndices(state.A)) do idx
```
leburgel: I don't love the `state.A` field access here. We can at least use `unitcell(state)`, but it would be even better to have `eachindex(IndexCartesian(), state)` working for use cases like this.

src/algorithms/compress/local.jl
```diff
@@ -0,0 +1,155 @@
...
+    P1 = R2 * vh' * sinv_sqrt
+    P2 = sinv_sqrt * u' * R1
+    return P1, P2, (; s, ϵ)
+end
+
+"""
+Truncate virtual bonds of the product of two 1-layer
+InfinitePEPOs `ρ1`, `ρ2` with `LocalTruncation`.
+"""
+function compress(ρ1::InfinitePEPO, ρ2::InfinitePEPO, alg::LocalTruncation)
```
leburgel: I like `compress`. I still think supplying the two PEPOs as two separate arguments is a bit weird though. I would personally be in favor of collecting the object to approximate into a single argument, either by supplying a tuple `ρ1ρ2::Tuple{<:InfinitePEPO,<:InfinitePEPO}`, or by explicitly stacking them into a single `InfinitePEPO` though a lazy multiplication `ρ1ρ2 = ρ1 * ρ2`.

A main reason would already be that if we ever want to implement this for more than two layers of PEPO, we wouldn't have to break the signature, just change the input checks, as opposed to the current signature.

src/algorithms/compress/local.jl
```diff
@@ -0,0 +1,155 @@
...
+    Pxs_info = map(Iterators.product(1:Nr, 1:Nc)) do (r, c)
+        # TODO: support SiteDependentTruncation
+        return virtual_projector(
+            ρ1[r, c], ρ2[r, c], ρ1[r, _next(c, Nc)], ρ2[r, _next(c, Nc)];
+            trunc = alg.trunc
+        )
+    end
+    # y-bond projectors: [r, c] on bond [r, c]--[r-1, c]
+    Pys_info = map(Iterators.product(1:Nr, 1:Nc)) do (r, c)
+        # TODO: reduce repeated rotations
```
leburgel: This seems pretty easy to do by preparing arrays like `ρ1_r90 = rotr90.(unitcell(ρ1))` before the loop?

leburgel:
I like the rename to `compress`, this takes away all the annoyance that comes with trying to fit things into MPSKit signatures. My only real comment is on using the two PEPOs in the product as separate arguments, the simplest issue with this being that it's not very future-proof against any changes in the number of layers being approximated.

If there's no need for a more-layer version in the foreseeable future I would be fine with it as is, just wanted to bring it up.

src/operators/infinitepepo.jl
Yue-Zhengyuan: I'll follow how we defined `eachindex` for BPEnv
```julia
Base.eachindex(env::BPEnv) = eachindex(IndexCartesian(), env.messages)
```
so we only allow Cartesian indices.

src/operators/infinitepepo.jl
Yue-Zhengyuan: The PEPO file already contains other networks, including PEPS and PartitionFunction. So I want to keep this function here.

src/algorithms/compress/local.jl
Yue-Zhengyuan: @Adwait-Naravane do you have use cases where multiple iPEPO layers are compressed at once, instead of compressing two layers at one time? I remember you mentioned doing this in 2602.13124.

Adwait-Naravane: No, I do not know of any case where you would compress more than two iPEPOs into one PEPO in a single truncation as that would be super expensive. What was done by Yuto in the method that we dubbed as "full update" was that he used some information obtained from the OR decomposition of a tower of pepo layers above and below the two pepos that were to be truncated in their local truncation. 
@leburgel the problem with the second approach is that there is a more efficient way to do local truncation of a PEPO by performing a QR on both the top and bottom pepos separately. This reduces the cost. The first approach sounds better.

Yue-Zhengyuan: @Adwait-Naravane Are you sure that doing QR separately on each layer still minimizes the 2-layer cost function? If so you can port your implementations in TNRKit here.

src/operators/infinitepepo.jl
leburgel: Okay, if it's already mixed then this is fine by me.

Yue-Zhengyuan:
I now think that `ρ1 * ρ2` is not good. It is mainly because `ρ1 * ρ2` will make a copy of the input in `cat`.

Some immediate generalizations are `(ψ_bra, ψ_ket)` (the bra-ket network for two iPEPS) and `(ρ, ψ)`. They don't fit with `*` for iPEPO only.

We never compress more than 2 layers at once. Only the cost function may have more layers. So I guess even tuples are not necessary. @Adwait-Naravane if we ever want to write this, how should we pass the additional layers to the function? Maybe inside the truncation algorithm struct?

I remember testing the layer-wise QR approach myself, and found that it wasn't minimizing the cost function. But anyway, if it can reduce cost, I can also provide an option to use it.

BTW, https://github.com/QuantumKitHub/TNRKit.jl/pull/172, which is doing essentially the same thing, took only one month to get merged. We spend too much time arguing about the interface.

leburgel:
> I now think that `ρ1 * ρ2` is not good. It is mainly because `ρ1 * ρ2` will make a copy of the input in `cat`.
> Some immediate generalizations are `(ψ_bra, ψ_ket)` (the bra-ket network for two iPEPS) and `(ρ, ψ)`. They don't fit with `*` for iPEPO only.

Sounds fair, I like the tuple option anyway.

> We never compress more than 2 layers at once. Only the cost function may have more layers. So I guess even tuples are not necessary. @Adwait-Naravane if we ever want to write this, how should we pass the additional layers to the function? Maybe inside the truncation algorithm struct?

Even if you only compress two layers at a time, you can still pass all the layers in 1 tuple and then just apply the 2-layer compression method recursively adding one extra layer in every time. So maybe I'm misunderstanding, but I don't think there's a need to pass the additional layers elsewhere or add them to the truncation algorithm struct. If we ever want to do multi-layer compression we just add 1 additional method with the same interface, which dispatches to the `::Tuple{<:InfinitePEPO, <:InfinitePEPO}` recursively internally. So this seems fine to me.

> BTW, [QuantumKitHub/TNRKit.jl#172](https://github.com/QuantumKitHub/TNRKit.jl/pull/172), which is doing essentially the same thing, took only one month to get merged. We spend too much time arguing about the interface.

We already have the tendency to break things quite often, to the dismay of quite a few users, so I wouldn't say we necessarily spend too much time discussing interfaces. However, it is true that we're just too slow. Having to wait days or even weeks for a review is just not nice. Being annoying about interface choices while also not making time to quickly follow up on reviews is a bad combination. I will try to improve on the speed at least and try to make some more time, that I can definitely get behind.

src/algorithms/changebonds/local.jl
leburgel: I can follow both reasonings, but since it does come down to an orthogonal decomposition in spirit in the end I'm going to force resolution on this.

src/algorithms/compress/local.jl
leburgel: I don't know what the discussion on QR-ing layers separately means exactly, but if it's an improvement or addition to the current approach it can always be handled in a follow up.

src/algorithms/compress/local.jl
```diff
@@ -0,0 +1,158 @@
...
+    P2 = sinv_sqrt * u' * R1
+    return P1, P2, (; s, ϵ)
+end
+
+"""
+    compress((ρ1, ρ2), alg::LocalTruncation)
+
+Compress two 1-layer iPEPOs into a 1-layer iPEPO by truncating the virtual bonds
+with `LocalTruncation`. In the tuple `(ρ1, ρ2)`, `ρ1` is the lower layer and
+`ρ2` is the upper layer.
```
leburgel, line 120: Another interface comment, but I just didn't notice before that the bottom layer came first and then the top. In most other places we usually order things top-bottom (I think). Is there a particular reason to use bottom-top ordering here? If not, I would prefer making `ρ1` the upper layer and `ρ2` the lower layer.

The same would go for truncating the application of a PEPO to a PEPS, we tend to put the state before the operator usually, so it would also be consistent with that.

src/algorithms/compress/local.jl
```diff
@@ -0,0 +1,158 @@
...
+"""
+function compress(ρs::Tuple{InfinitePEPO, InfinitePEPO}, alg::LocalTruncation)
```
leburgel:
```suggestion
function compress(ρs::Tuple{<:InfinitePEPO, <:InfinitePEPO}, alg::LocalTruncation)
```
I would expect the `Tuple{InfinitePEPO, InfinitePEPO}` to give trouble at some point?

Yue-Zhengyuan:
I'm currently testing the layerwise-QR approach and see whether it really minimizes the cost. Let's aim for merging in this week.

src/operators/infinitepepo.jl
```diff
@@ -212,6 +213,19 @@ end
 VI.scalartype(::Type{NT}) where {NT <: InfinitePEPO} = scalartype(eltype(NT))
 VI.zerovector(A::InfinitePEPO) = InfinitePEPO(zerovector(unitcell(A)))
 
+## Math
+
+"""
+    ρ1 * ρ2
+
+Lazily multiply two infinite PEPOs by stacking the layers of `ρ1` under those of `ρ2`.
```
leburgel, line 235: Here it does make sense to put `ρ1` under `ρ2`, because PEPOS act from top to bottom.

leburgel (commented):
Final comments, but otherwise looks good to me now.

src/algorithms/compress/local.jl
Yue-Zhengyuan: The reason is the same as `ρ1` being the bottom layer in `ρ1 * ρ2`. For `ρ * ψ`, it is also natural to put ρ under ψ (since the physical indices point downwards).

Yue-Zhengyuan: In `expectation_value`, we use the order (bra, O, ket), also from bottom to top.

leburgel: I see, I was thinking more of the leg ordering in the contractions themselves, but that has nothing to do with method interface. Also for constructing overlap networks we have signatures like `InfiniteSquareNetwork(top, bot = top)`, but on further thought I guess that has more to do with being able to set the default value for the optional argument than with actual logic.

Even the signature of `MPSKit.approximate` follows the same logic of starting with the operator that is being applied, which I somehow misremembered.

So I agree this makes sense, sorry about the confusion.

Yue-Zhengyuan:
Just checked more carefully on the layerwise-QR approach currently used in TNRKit. It indeed minimizes the local cost function, so I was wrong in this aspect. However, the computational cost is higher for layerwise-QR (if we trust the analysis done by Codex, attached below). So let's only keep the current HOTRG-like implementation.

@Adwait-Naravane @VictorVanthilt You may want to revisit the two-layer QR in `TTNR` and switch to 3D-HOTRG to compress the temperature (imaginary time) direction.

<details>

**`layerwise_qr=false` path** (current PEPSKit implementation):

This forms the two-layer Gram matrix `MdagM` / `MMdag`, whose matrix size is

```text
D^2 x D^2
```

With a good contraction order, each Gram construction costs roughly

```text
O(d^3 D^5)
```

then `eigh_full!` on the `D^2 x D^2` matrix costs

```text
O((D^2)^3) = O(D^6)
```

So one left/right orthogonalization costs

```text
O(d^3 D^5 + D^6)
```

and the full `virtual_projector` has the same scaling, plus the common final `R1 * R2` and SVD cost `O(D^6)`.

**`layerwise_qr=true` path** (TNRKit implementation):

Each single-layer QR treats a PEPO tensor as a matrix of size approximately

```text
(d D^3) x (d D)
```

so each QR costs

```text
O(d D^3 * (d D)^2) = O(d^3 D^5)
```

After two such QRs, the intermediate `M` has matrix size roughly

```text
(d^2 D^2) x (D^2)
```

and QR of that costs

```text
O(d^2 D^2 * (D^2)^2) = O(d^2 D^6)
```

So one left/right orthogonalization costs

```text
O(d^3 D^5 + d^2 D^6)
```

and the whole `virtual_projector` costs

```text
O(d^3 D^5 + d^2 D^6)
```

up to the common final `O(D^6)` SVD.

**Bottom line**

Cost-wise, `layerwise_qr=false` is cheaper asymptotically in the physical dimension:

```text
false: O(d^3 D^5 + D^6)
true:  O(d^3 D^5 + d^2 D^6)
```

For fixed small `d`, both are effectively `O(D^6)` in virtual bond dimension, but `layerwise_qr=true` carries an extra `d^2` prefactor in the final QR stage. So the layerwise path is not computationally cheaper in the dense uniform-dimension estimate; its advantage would have to come from numerical behavior, block/symmetry structure, or practical contraction/QR implementation details.

</detail>

leburgel (approved):

Adwait-Naravane:
Your analysis is correct. The HOTRG way of deriving the projectors is indeed better when both the layers have the same dimensions. But in TTNR, we contract a larger PEPO with a smaller PEPO (which corresponds to the initial PEPO), here I think the cost for the HOTRG is `O(d^3 D^5 + d^3 D^3)` while for TTNR is `O(d^3 D^5 + d^5 D^3)`. HOTRG looks better here as well.

Yue-Zhengyuan:
The code for finding HOTRG projectors in TNRKit does not assume that the two layers are identical. So the migration should be easy.

PR #320, opened 2026-01-21
Add gauge transformation for iPEPS and its CTMRGEnv

Yue-Zhengyuan:
This PR adds a utility function `gauge_transform` for iPEPS and its CTMRGEnv, which transforms the iPEPS or its environment using the gauge matrices `XXinv` (e.g., produced by BP gauge fixing) on the virtual bonds of the iPEPS.

It is tested by checking that the (1-site) reduced density matrices constructed from CTMRG are unchanged after an arbitrary gauge transformation.


leburgel:
I think this would be really great to have. Looking at this though, I'm wondering if it might be good to wrap the gauge matrices `XXinv` into a dedicated struct with properly documented indexing rules (basically just the diagrams you showed here). I think that might be a good reusable starting point for representing gauge transformations, even for more involved cases of different square networks. What do you think?

PR #339, opened 2026-03-04
N-site Simple Update

Yue-Zhengyuan:
This PR rewrites the 3-site simple update so that it can actually accept gate MPOs acting on an arbitrary open path along the nearest neighbor bonds. Closed paths (such as one acting on the four sites on an elementary plaquette) are not supported yet, since that requires PBC-MPS treatments.

I introduced a wrapper struct `TrotterGates`, which stores the Trotter 1-site gates, 2-site gates, and MPOs as a vector of pairs `sites => gate`, where `sites` records the path on which `gate`.

I also make some improvements on how the gates and NNN 3-site MPOs are constructed from the Hamiltonian, so Hamiltonians with 1-site terms can also be used for simple update (previously I restricted them to only contain 2-site terms). In addition, second-order Trotter decomposition is now supported.

~Another minor improvement is on the SU gauging by applying trivial gates. Now the trivial gate is never explicitly constructed and applied.~ (Removed to reduce diff; SU gauging does not cost much anyway.)


Yue-Zhengyuan:
I think this PR is now finally stable enough for a review.

src/operators/models.jl
```diff
@@ -79,15 +79,18 @@ function MPSKitModels.hubbard_model(
...
+            MPSKitModels.e⁻e⁺(T, particle_symmetry, spin_symmetry)
+    )
     interaction_term = MPSKitModels.nꜛnꜜ(T, particle_symmetry, spin_symmetry)
     site_term = U * interaction_term - mu * N
-    h = (-t) * hopping + (1 / 4) * (site_term ⊗ unit + unit ⊗ site_term)
-    return nearest_neighbour_hamiltonian(fill(pspace, size(lattice)), h)
+    return LocalOperator(
+        spaces,
+        (neighbor => hopping for neighbor in nearest_neighbours(lattice))...,
+        ((idx,) => site_term for idx in vertices(lattice))...,
```
lkdvos: I think this is strictly less efficient for us, since now we have to evaluate more contractions, so I'm not sure I really agree with this change

lkdvos (commented):
Overall this looks like an improvement to me. I'm not entirely sure how to proceed in the easiest way, since I would like to tackle the LocalOperator and the Trotterization in a separate PR from this to keep things easier to review.
How do you feel about merging this, and then me trying a follow-up PR?

src/operators/models.jl
Yue-Zhengyuan: I've reverted the changes. In the same spirit, the transverse field Ising model (magnetic field terms), the Bose-Hubbard model (chemical potential & on-site interaction) and the p-wave SC (chemical potential terms) can also be improved by grouping the 1-site terms with the NN terms. But this is not urgent anyway.

lkdvos (approved):
I think I'm okay with merging this. Before tagging I will add a follow-up with some more changes for the operator types, but that is definitely easier to do in a new PR

PR #343, opened 2026-03-19
Rename bondenv_fu to bondenv_ctm

Yue-Zhengyuan:
Just want to emphasize that the full update bond environment comes from CTMRG.


lkdvos (approved):
Ok for me, might as well before the release

PR #344, opened 2026-03-19
Fix stackoverflow in `renormalize_southwest_corner`

leburgel:
Bumped into this by accident: using `renormalize_southwest_corner` directly on an enlarged corner and its projectors gave a stackoverflow, must have been a typo.

### DIFF
--- a/src/algorithms/contractions/ctmrg/renormalize_corner.jl
+++ b/src/algorithms/contractions/ctmrg/renormalize_corner.jl
@@ -345,7 +345,7 @@ end
 function renormalize_southwest_corner(
         quadrant::AbstractTensorMap{T, S, N, N}, P_left, P_right
     ) where {T, S, N}
-    return renormalize_southwest_corner(quadrant, P_left, P_right)
+    return renormalize_corner(quadrant, P_left, P_right)
 end
 function renormalize_southwest_corner(
         E_south, C_southwest, E_west, P_left, P_right, A::PEPSSandwich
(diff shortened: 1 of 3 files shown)


Yue-Zhengyuan:
Looking back on this, it seems that all four `renormalize_xxxx_corner(quadrant, P_left, P_right)` are simply wrappers to the same `renormalize_corner(quadrant, P_left, P_right)`. How about we directly write, for example,
```julia
function renormalize_northwest_corner((row, col), enlarged_env, P_left, P_right)
    return renormalize_corner(
        enlarged_env[NORTHWEST, row, col],
        P_left[NORTH, row, col],
        P_right[WEST, _next(row, end), col],
    )
end
```
to reduce one layer of wrapping?

leburgel:
I don't think this works, because the `enlarged_env` can also contain a sparse representation of the enlarged corners, as `EnlargedCorner` instances. So the extra layer is required to dispatch to the `EnlargedCorner` implementations like
```julia
function renormalize_southwest_corner(ec::EnlargedCorner, P_left, P_right)
    return renormalize_southwest_corner(ec.E_1, ec.C, ec.E_2, P_left, P_right, ec.A)
end
```

Yue-Zhengyuan:
Is it convenient to add some test coverage here? In #326, I observed that tests only cover half of the CTMRG code.

leburgel:
Yes, we should add some coverage here before merging, seems like a good opportunity.

leburgel:
I added a small test to unblock this for now. We can add to this test stub afterwards until all contractions are covered by just comparing dense and sparse versions, even if there are some that aren't actually used in any of the algorithms yet.

Yue-Zhengyuan (approved):
Good to go for me. Though we need to reduce duplication of tests some time in the future.

PR #346, opened 2026-03-19
Simple update refactoring

Yue-Zhengyuan:
Here are some refactorings I want to backport from my NTU branch.

- `_qr_bond` now can pass additional `kwargs` for finer control of `left_orth` inside. This is useful when the gate is applied before doing `left_orth`, which needs an error-based truncation for rank-revealing.
- As I will often rotate things to put the updated bond in the "canonical" x-direction (A - B), I introduced `_bond_rotation` to do this.
- Minor adjustment of `_su_iter!` signature.

### DIFF
--- a/src/algorithms/time_evolution/simpleupdate.jl
+++ b/src/algorithms/time_evolution/simpleupdate.jl
@@ -70,7 +70,11 @@ function _bond_rotation(x, bonddir::Int, rev::Bool; inv::Bool = false)
     return if bonddir == 1 # x-bond
         rev ? rot180(x) : x
     elseif bonddir == 2 # y-bond
-        rev ? (inv ? rotr90(x) : rotl90(x)) : (inv ? rotl90(x) : rotr90(x))
+        if rev
+            inv ? rotr90(x) : rotl90(x)
+        else
+            inv ? rotl90(x) : rotr90(x)
+        end
     else
         error("`bonddir` must be 1 (for x-bonds) or 2 (for y-bonds).")
     end
--- a/src/algorithms/time_evolution/simpleupdate.jl
+++ b/src/algorithms/time_evolution/simpleupdate.jl
@@ -66,50 +66,44 @@ function TimeEvolver(
     return TimeEvolver(alg, dt, nstep, gate, state)
 end
 
-"""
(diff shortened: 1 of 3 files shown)


src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -66,50 +66,44 @@ function TimeEvolver(
...
 end
 
-"""
-Optimized simple update of nearest neighbor bonds utilizing
-reduced bond tensors without decomposing the gate into a 2-site MPO.
+function _bond_rotation(x, bonddir::Int, rev::Bool; inv::Bool = false)
+    return if bonddir == 1 # x-bond
+        rev ? rot180(x) : x
+    elseif bonddir == 2 # y-bond
+        rev ? (inv ? rotr90(x) : rotl90(x)) : (inv ? rotl90(x) : rotr90(x))
```
lkdvos: This is a bit hard to read the logic behind this, am I correct that this is:

| rev\inv | false | true |
| ------- | ----- | ---- |
| false | R | L |
| true | L | R |

Or thus:
```suggestion
        rev ⊻ inv ? rotl90(x) : rotr90(x)
```

lkdvos (approved):
Except for the comment about the boolean logic this looks good to me!

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: I actually thought about the same thing. But what holds me back is that while `rev` and `inv` are clear individually, `rev ⊻ inv` does not appear to have a intuitive meaning.

- `rev` is whether the bond is reversed w.r.t. the canonical x/y direction (w - e or s - n).
- `inv` is whether I'm rotating the bond to or from the original position.

So I'd rather write longer.

lkdvos: I'm okay with that too, but then please just write the `if` statements, the triply nested `?` on a single line is what tripped me up, I don't mind the if statments that much in this case, or even a combination of the two :)

Yue-Zhengyuan: Fixed. Should look better now.

PR #347, opened 2026-03-31
Refactor operator types

lkdvos:
This is a WIP implementation of updated structs for our operators, with the goal of being less painful for the compiler, while being a bit more natural for representing sums as dictionaries (commuting terms -> order does not matter), while products are vectors (non-commuting factors -> order retained).

Unfortunately I will not find the time to finish this, so if anyone feels like continuing the work, go ahead!
I do think most of what I have in mind is here, the general ideas are in place and this is mostly about cleaning up and dealing with Zygote.

### DIFF
--- a/docs/src/man/models.md
+++ b/docs/src/man/models.md
@@ -28,7 +28,7 @@ function MPSKitModels.transverse_field_ising(
     return LocalOperator(
         spaces,
         (neighbor => ZZ for neighbor in nearest_neighbours(lattice))...,
-        ((idx,) => X for idx in vertices(lattice))...,
+        ([idx,] => X for idx in vertices(lattice))...,
     )
 end
 ```
--- a/src/algorithms/contractions/bp_contractions.jl
+++ b/src/algorithms/contractions/bp_contractions.jl
@@ -60,13 +60,51 @@ function MPSKit.expectation_value(peps::InfinitePEPS, O::LocalOperator, env::BPE
 end
 
 function contract_local_operator(
-        inds::NTuple{1, CartesianIndex{2}},
+        inds::Vector{CartesianIndex{2}},
+        O::AbstractTensorMap,
+        ket::InfinitePEPS,
+        bra::InfinitePEPS,
+        env::BPEnv,
+    )
+    length(inds) == 1 && return contract_local_operator1x1(only(inds), O, ket, bra, env)
+
+    if length(inds) == 2
+        ind_relative = inds[2] - inds[1]
+        if ind_relative == CartesianIndex(1, 0)
+            return contract_local_operator2x1(inds[1], O, ket, bra, env)
+        elseif ind_relative == CartesianIndex(0, 1)
+            return contract_local_operator1x2(inds[1], O, ket, bra, env)
+        end
+    end
+    error("No implementation for contractions for BP environments with $inds")
+end
+function contract_local_norm(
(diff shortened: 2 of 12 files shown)


leburgel:
I'll try to have a decent look soon, but some comments/notes already:
- ~~It seems the MatrixAlgebraKit.jl v0.6.5 release broke our QR CTMRG optimization when using a `Float64` scalar type. This is now failing in the Heisenberg example, also on master. No clue what changed and this definitely needs to be sorted out, but technically unrelated.~~
    - UPDATE: this turned out not to be a fundamental issue, it just seems that some combination of changes here and in MAK v0.6.5 broke things for the specific seed we were using. Changing the seed seems to resolve the issue.
-  There used to be a check on the norm of the terms being added to a `LocalOperator`, where any term whose norm was too small was just skipped. This is no longer present in the `add_term!` implementation, causing a failure of `@assert length(imag(H).terms) == 0` for a Hamiltonian with purely real terms. The cutoff was chosen a bit arbitrarily, but I think it can't hurt to just to skip terms whose norm is exactly zero, i.e, using an `atol = zero(real(scalartype(term)))`. I can add that.
    - UPDATE: I added an `atol` check for the norm of a term in `add_term!`, and set the default to only skip terms with an exact zero norm.
- It looks like a `SpaceMismatch` snuck into the time evolution workflow. No idea where this is coming from exactly, but I'll try to investigate a bit more.

Yue-Zhengyuan:
I'll try to fix the time evolution part.

Yue-Zhengyuan:
The space mismatch is because the site order for gates on vertical bonds becomes `(north, south)` (due to how Julia sorts CartesianIndex's), which is opposite to the "canonical order" (along the virtual arrow directions) in iPEPS/PEPO. The produced bond weight should then be transposed, which was missing before.

I added more careful checks for MPO gates in LocalCircuit:
- Two consecutive sites along the path `inds` should be nearest neighbors. For this reason, we should *not* sort `inds` for MPO terms.
- `inds` cannot contain repeated coordinates (for LocalOperator as well).

I also removed the check that `H` and the Trotterized circuit should have the same number of terms. For example, each NNN term is converted to two MPO gates.
```
1        1---2
|            |
2---3        3
```

@leburgel I guess you can now proceed to see if there are other things to be cleaned up.

pbrehmer:
This is really nice, thanks! I update the docs and model definitions to define the `LocalOperator` terms using `Vector`s instead of `NTuple`s since that is now more in line with the field types (although the tuple definitions still work because `add_term!` just needs in iterable type I guess).

Yue-Zhengyuan:
It may be better to move `src/algorithms/time_evolution/trotter_gate.jl` to `src/operators/localcircuit.jl`. But this can be done in a follow-up PR if you wish to make the diff smaller.

leburgel (approved):
The QR CTMRG example failure can probably be fixed by changing the seed again, it seems the change I made broke the `ComplexF64` scalartype test at the same time it fixed the `Float64` one. I'll try to stabilize these properly later.

Gradient timeouts might be due to increased compilation times that came with the type changes here. Also not really related to the changes themselves, so we can sort it out later.

In favor of merging this as is, and doing smaller fixes and cleanup as follow-ups.

Yue-Zhengyuan:
@leburgel @pbrehmer Need your help to merge due to test errors

PR #348, opened 2026-04-06
Improve Vidal gauge conversion, CTM bondenv for PEPO

### LINKED ISSUE #352: Wrong finite-T SU result for Ising model with `force_mpo = true`
When running `test/timeevol/tf_ising_finiteT.jl` with `force_mpo = true` (i.e. converting even the nearest neighbor Trotter gates to 2-site MPOs) in the SimpleUpdate algorithm, the resulting magnetization is wrong. Strangely, this does not affect `test/timeevol/j1j2_finiteT.jl`. I'll investigate this problem and should fix it before releasing.

Yue-Zhengyuan:
This PR does the following refactorizations (sorry for packing many things together):

- In `src/algorithms/contractions/bondenv/als_solve.jl`: the axis order of tensors involved in the ALS optimization (in particular, the `S` tensors and the rank-3 reduced bond tensor) is changed to more closely follow the MPS convention.

- In `src/algorithms/contractions/bondenv/benv_ctm.jl`: CTMRG bond environment can now be constructed for an iPEPO, getting ready for finite-T FU.

- In `src/algorithms/time_evolution/apply_mpo.jl`: The conversion of an open boundary MPS to Vidal gauge is improved by removing the constraint of the number of legs of each tensor. Closes #352.

### DIFF
--- a/src/algorithms/contractions/bondenv/benv_tools.jl
+++ b/src/algorithms/contractions/bondenv/benv_tools.jl
@@ -1,3 +1,13 @@
 const BondEnv{T, S} = AbstractTensorMap{T, S, 2, 2} where {T <: Number, S <: ElementarySpace}
+const BondEnv3site{T, S} = AbstractTensorMap{T, S, 4, 4} where {T <: Number, S <: ElementarySpace}
+const Hair{T, S} = AbstractTensor{T, S, 2} where {T <: Number, S <: ElementarySpace}
+# Orthogonal tensors obtained PEPSTensor/PEPOTensor
+# with one physical leg factored out by `_qr_bond`
 const PEPSOrth{T, S} = AbstractTensor{T, S, 4} where {T <: Number, S <: ElementarySpace}
 const PEPOOrth{T, S} = AbstractTensor{T, S, 5} where {T <: Number, S <: ElementarySpace}
+
+"Convert tensor `t` connected by the bond to be truncated to a `PEPSTensor`."
+_prepare_site_tensor(t::PEPSTensor) = t
+_prepare_site_tensor(t::PEPOTensor) = first(fuse_physicalspaces(t))
+_prepare_site_tensor(t::PEPSOrth) = permute(insertleftunit(t, 1), ((1,), (2, 3, 4, 5)))
+_prepare_site_tensor(t::PEPOOrth) = permute(t, ((1,), (2, 3, 4, 5)))
--- a/src/algorithms/time_evolution/apply_mpo.jl
+++ b/src/algorithms/time_evolution/apply_mpo.jl
@@ -112,7 +112,7 @@ Then the fidelity is just
 ```
 =#
 """
-Perform QR decomposition through a PEPS tensor
+Perform QR decomposition through a `GenericMPSTensor`
 ```
              ╱            ╱
     -←-R0-←-M-←-  =>  ---Q-←-R1-←-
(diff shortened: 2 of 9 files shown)


src/algorithms/contractions/bondenv/benv_tools.jl
```diff
@@ -1,3 +1,13 @@
...
+const BondEnv3site{T, S} = AbstractTensorMap{T, S, 4, 4} where {T <: Number, S <: ElementarySpace}
+const Hair{T, S} = AbstractTensor{T, S, 2} where {T <: Number, S <: ElementarySpace}
+# Orthogonal tensors obtained PEPSTensor/PEPOTensor
+# with one physical leg factored out by `_qr_bond`
 const PEPSOrth{T, S} = AbstractTensor{T, S, 4} where {T <: Number, S <: ElementarySpace}
 const PEPOOrth{T, S} = AbstractTensor{T, S, 5} where {T <: Number, S <: ElementarySpace}
+
+"Convert tensor `t` connected by the bond to be truncated to a `PEPSTensor`."
+_prepare_site_tensor(t::PEPSTensor) = t
+_prepare_site_tensor(t::PEPOTensor) = first(fuse_physicalspaces(t))
```
leburgel, line 11: Looks like these two are currently not used. When do you expect these to be required?

src/algorithms/time_evolution/apply_mpo.jl
```diff
@@ -112,57 +112,63 @@ Then the fidelity is just
 ```
 =#
 """
-Perform QR decomposition through a PEPS tensor
+Perform QR decomposition through a `GenericMPSTensor`
```
leburgel, line 115: It's not really clear from the diagram how this `R0*M` is interpreted as an MPS tensor, is it?. This isn't really an issue, I was just a bit confused myself.

leburgel (approved):
Looks good, no real important comments from me.

Just one question more out of curiosity: what is the reason for generalizing to MPS tensors with an arbitrary number of physical legs in the MPO application routines? When will this be necessary and/or why was it not necessary before?

src/algorithms/time_evolution/apply_mpo.jl
Yue-Zhengyuan: It is more like pulling `R0` through an MPS tensor `M` with QR decomposition. Although in the diagram `R0 * M` is drawn as a PEPSTensor, it should be regarded as an MPS tensor with (west, phys, north, south) in the codomain, and (east,) in the domain. You may imagine the north, south legs as extra "physical legs" of the MPS tensor drawn in a non-planar way.

src/algorithms/contractions/bondenv/benv_tools.jl
Yue-Zhengyuan: These would be used if we decided to keep the physical index with `Q` when producing the reduced bond tensor with QR (currently we are keeping the physical index with the bond tensor). I made some attempts in #322, and it turns out to be less efficient. But since it is just two lines, I didn't bother to remove it...

Yue-Zhengyuan:
I am trying to implement 3-site iterative ALS optimization when applying a next-nearest neighbor gate (as in https://arxiv.org/abs/1711.07584). It is like "weighted truncation" of an MPS surrounded by an environment. In this case, the 3-site cluster consists of two rank-3 reduced bond tensors at two ends, and a full PEPSTensor in the middle.

I actually also want to get the routine of Vidal gauge conversion ready to be put into MPSKit, so I decided to first try it out here.

Yue-Zhengyuan:
Turns out that after merging this locally (actually only updating `src/algorithms/time_evolution/apply_mpo.jl`), the error in #352 is gone. So there were actually some error in the conversion to Vidal gauge...

Yue-Zhengyuan:
@leburgel Help in merging this one as well if you feel my replies are OK. I need this to rerun updated tests in #350.

PR #350, opened 2026-04-08
Cover `force_mpo` in simple update test

Yue-Zhengyuan:
This PR removes an error that will never be raised.

https://github.com/QuantumKitHub/PEPSKit.jl/blob/cb2c179006d3bbf1d741303795a944bfa406c78f/src/algorithms/time_evolution/simpleupdate.jl#L143-L148

As `length(sites) == 2`, the error will not be raised and is redundant.

It also updated the finite-T SU test with transverse field Ising model to test `force_mpo = true`, i.e. applying a nearest neighbor 2-site gate as an MPO (instead of a rank-4 TensorMap) on an iPEPO.


Yue-Zhengyuan:
I'll use this PR to fix #352.

test/timeevol/tf_ising_finiteT.jl
```diff
@@ -44,12 +44,13 @@ dt, nstep = 1.0e-3, 400
 β = dt * nstep
 
 # when g = 2, β = 0.4 and 2β = 0.8 belong to two phases (without and with nonzero σᶻ)
-@testset "Finite-T SU (bipartite = $(bipartite))" for bipartite in (true, false)
+@testset "Finite-T SU (force_mpo = $(force_mpo))" for force_mpo in (false, true)
     # use second order Trotter decomposition
     symmetrize_gates = true
+    bipartite = true
```
Yue-Zhengyuan, line 50: The general non-bipartite case is already tested elsewhere, so I think it is safe to skip it here.

leburgel (approved):

leburgel:
I'll merge as is. The example test failure should be fixed by #353, and the gradient test timeout will require more work to sort out.

PR #353, opened 2026-04-10
Stabilize C4v AD optimization test

leburgel:


Yue-Zhengyuan (approved):
OK to merge once the tests pass. I heard this is just because some random seeds are particularly awful in a way that can be theorized.

Yue-Zhengyuan:
There are a bunch of "Node.js 20 actions are deprecated" warnings in the CI action log.

leburgel:
Where do you see these exactly? All I see is deprecation warnings from the MatrixAlgebraKit.jl decompositions, since all `LAPACK_`-specific algorithm names were deprecated in v0.6.5, which we should address next.

EDIT: never mind, I see them. Not sure if we actually need to do anything about this tough, probably we'll just need an update of action versions used in https://github.com/QuantumKitHub/QuantumKitHubActions/, which will then solve the issue here?

Yue-Zhengyuan:
It's @lkdvos's work then, I guess...

lkdvos:
Commenting on the actions here: I will/have updated most of them, but this is also related to some downstream updates that are not in my hands (and will be updated automatically), so this can be safely ignored.

PR #356, opened 2026-04-12
Require manual construction of TimeEvolver for Simple Update

### LINKED ISSUE #355: Require explicit construction of TimeEvolver for `time_evolve`
I want to stop supporting the high-level interface of time evolution
```julia
time_evolve(
        psi0::Union{InfinitePEPS, InfinitePEPO}, H::LocalOperator, dt::Number, nstep::Int,
        alg::SimpleUpdate, env0::SUWeight; symmetrize_gates::Bool = false,
        tol::Float64 = 0.0, t0::Number = 0.0, check_interval::Int = 500
    )
```
thus forcing users to first construct a `TimeEvolver`, and use
```julia
time_evolve(
        it::TimeEvolver{<:SimpleUpdate}; 
        tol::Float64 = 0.0, check_interval::Int = 500
    )
```

Reasons are:

- For NTU #144, ground state search needs a few more arguments than other purposes: a CTMRG environment, a CTMRG algorithm, and a tolerance for energy change between two checks. I feel frustrated to maintain additional logic in the high-level interface to ensure that a correct combination of arguments is supplied.

- ~For ground state search, which already has an (BP or CTMRG) environment to evaluate expectation values, I want to introduce the ability to monitor quantities `observables` besides the energy during the evolution~ (Now I feel this is better not to be a built-in feature)

Yue-Zhengyuan:
This PR implements proposals in #355.

- A `TimeEvolver` struct needs to be explicitly constructed before calling `time_evolve`, instead of being automatically created by the latter internally.
- When searching for ground state, the Hamiltonian is a required argument of `time_evolve` to monitor the change of energy (although SimpleUpdate does not use it for convergence checking).
- ~During imaginary time evolution for iPEPS ground state, one can monitor the expectation value of the Hamiltonian and optionally an additional vector `observables` of `LocalOperator`s. In Simple Update, the expectation values are evaluated with the `SUWeight` by converting it to `CTMRGEnv`. It should be changed to `BPEnv` in the future once it also supports arbitrary coordinates.~

TODO:

- [ ] Clean up docstrings.
- [x] Update examples to explicitly construct a `TimeEvolver` for time evolution.

### DIFF
--- a/examples/heisenberg_su/main.jl
+++ b/examples/heisenberg_su/main.jl
@@ -79,7 +79,8 @@ nstep = 10000
 trunc_peps = truncerror(; atol = 1.0e-10) & truncrank(Dbond)
 alg = SimpleUpdate(; trunc = trunc_peps, bipartite = true)
 for (dt, tol) in zip(dts, tols)
-    global peps, wts, = time_evolve(peps, H, dt, nstep, alg, wts; tol, check_interval = 500)
+    evolver = TimeEvolver(peps, H, dt, nstep, alg, wts)
+    global peps, wts, = time_evolve(evolver, H; tol, check_interval = 500)
 end
 
 md"""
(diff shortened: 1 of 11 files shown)


Yue-Zhengyuan:
As discussed, the high-level `time_evolve` will temporarily work only with `SimpleUpdate`. I'll repurpose this branch to fix some other things.

Issue #351, opened 2026-04-08
Dimension mismatch in SVD pullback due to `rank_atol` for `FixedSVD`


pbrehmer:
I was running into a dimension mismatch error in the pullback of `FixedSVD` due to the default finite `rank_atol` for MatrixAlgebraKit's `svd_pullback` function. Note that we here compute the forward using a full SVD such that we can utilize the full SVD pullback. Here's the full error message:

<details><summary>Details</summary>
<p>

```
...the stacktrace before and after this part make the message to long for GitHub...

nested task error: DimensionMismatch: second dimension of ΔU (40) does not match rank of S (25)
    Stacktrace:
      [1] svd_pullback!(ΔA::Base.ReshapedArray{ComplexF64, 2, SubArray{ComplexF64, 1, Vector{ComplexF64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, A::Base.ReshapedArray{ComplexF64, 2, SubArray{ComplexF64, 1, Vector{ComplexF64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, USVᴴ::Tuple{Base.ReshapedArray{ComplexF64, 2, SubArray{ComplexF64, 1, Vector{ComplexF64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, LinearAlgebra.Diagonal{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}}, Base.ReshapedArray{ComplexF64, 2, SubArray{ComplexF64, 1, Vector{ComplexF64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}}, ΔUSVᴴ::Vector{Any}, ind::UnitRange{Int64}; rank_atol::Float64, degeneracy_atol::Float64, gauge_atol::Float64)
        @ MatrixAlgebraKit ~/.julia/packages/MatrixAlgebraKit/x7OBV/src/pullbacks/svd.jl:58
[...]

PR #357, opened 2026-04-13
Fix dimension mismatches in SVD pullback due to `rank_atol`

pbrehmer:
As described in #351, there was the possibility for a `DimensionMismatch` error due to the non-zero default `rank_atol` in MatrixAlgebraKit's SVD pullbacks. Here I set `rank_atol=0` explicitly for `FixedSVD` and `IterSVD` pullbacks, so that error shouldn't happen again.

I'm not quite sure how to test this, so if anybody has an idea, let me know.

### DIFF
--- a/src/utility/eigh.jl
+++ b/src/utility/eigh.jl
@@ -327,7 +327,7 @@ end
 ## Reverse-rule algorithms
 #
 
-function _get_pullback_gauge_tol(verbosity::Int)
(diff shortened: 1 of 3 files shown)


lkdvos:
Will close this since it should be fixed by https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/208

PR #358, opened 2026-04-13
Increase test coverage for CTMRG contractions

leburgel:
Adds coverage for all of the sparse CTMRG contractions that were previously untested, by directly comparing their results to the dense equivalents. Quite a few of these were broken or even missing, so I fixed things up where necessary and added the missing parts so that we can actually do everything we would need using sparse environments.

There's one part that's a bit unclear to me, in the application of a (half or full infinite) environment to a vector, which I think we plan to use for fully matrix-free sparse SVDs in the future. There I'm not sure if for the "right" action we need the contraction `tensorcontract(env, x)` or the multiplication `env * x` instead, and the same for the "left" action. ~~I chose contraction for now since that was the easiest and there is currently no use case to check this on. But if we need multiplication we'll have to add some additional twists.~~ If we want the use of `svdsolve` to correspond to an `svd_trunc` on the dense environment, then application should correspond to proper tensor composition, so `env * x`. I updated the implementation and test accordingly.

### DIFF
--- a/src/algorithms/ctmrg/sparse_environments.jl
+++ b/src/algorithms/ctmrg/sparse_environments.jl
@@ -99,7 +99,7 @@ function half_infinite_environment(ec_1::EnlargedCorner, ec_2::EnlargedCorner)
 end
 
 # Compute left and right projectors sparsely without constructing enlarged corners explicitly
-function left_and_right_projector(U, S, V, Q::EnlargedCorner, Q_next::EnlargedCorner)
+function contract_projectors(U, S, V, Q::EnlargedCorner, Q_next::EnlargedCorner)
     isqS = sdiag_pow(S, -0.5)
     P_left = left_projector(Q.E_1, Q.C, Q.E_2, V, isqS, Q.A)
     P_right = right_projector(Q_next.E_1, Q_next.C, Q_next.E_2, U, isqS, Q_next.A)
--- a/src/algorithms/ctmrg/sparse_environments.jl
+++ b/src/algorithms/ctmrg/sparse_environments.jl
@@ -100,9 +100,11 @@ end
 
 # Compute left and right projectors sparsely without constructing enlarged corners explicitly
 function contract_projectors(U, S, V, Q::EnlargedCorner, Q_next::EnlargedCorner)
+    Ar = _rotate_north_localsandwich(Q.A, _prev(Q.dir, 4))
+    Ar_next = _rotate_north_localsandwich(Q_next.A, Q_next.dir)
     isqS = sdiag_pow(S, -0.5)
-    P_left = left_projector(Q.E_1, Q.C, Q.E_2, V, isqS, Q.A)
-    P_right = right_projector(Q_next.E_1, Q_next.C, Q_next.E_2, U, isqS, Q_next.A)
+    P_left = left_projector(Q_next.E_1, Q_next.C, Q_next.E_2, V, isqS, Ar_next)
+    P_right = right_projector(Q.E_1, Q.C, Q.E_2, U, isqS, Ar)
     return P_left, P_right
 end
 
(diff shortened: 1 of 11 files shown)


leburgel:
Aside from the conceptual question above about what we want the environment application to mean exactly in the presence of fermionic symmetries, there's three more things to be addressed here I think:
- I copied over `twistnondual` from #311, since I naturally needed it to fix the computation of the right projector by multiplying in a right isometry which requires twists one some of its domain spaces. I don't think that's a problem, we just have to remove it from whichever one gets merged first.
- ~~I added `rrule`s for both `twistdual` and `twistnondual`, since these out-of-place implementations call in-place methods on a copy and are therefore not automatically differentiable. I did this because the CTMRG subroutines should really be differentiable, but since the sparse routines are currently not used for optimization there rules are not covered by tests right now. I can just remove them for now and add them in when we actually need them, I think that might be better.~~ I removed the unused `rrules`, they're easy to add again once they're needed.
- There's something weird going on with codecov and `@tensor begin ... end` blocks, and all of these show up as uncovered even though they are actually used in the tests. In particular, this leads to bad coverage being reported for src/algorithms/contractions/ctmrg/enlarge_corner.jl, where some of the contractions were fully written out to minimize permutations using `@tensor begin ... end` blocks. I don't know if there's something that can be done about this, or if it would actually be worth it.

src/algorithms/ctmrg/sparse_environments.jl
lkdvos, line 1: Looking at this a little bit, the only thing that seems slightly off to me is to have to store both the rotated as well as the unrotated PEPS tensors. I understand why you want to add the rotated version, (since we want to optimize the repeated application of these operators, right?), but could we get away with not having to store the unrotated version then? Trying to look into this and it seems they are only needed in the contract projectors since that still goes through the enlarged corner construction, can we get rid of that one?

The main reason why I'm saying this is that if we have to construct the enlarged corners here anyways, it is actually way more efficient to then compute the action through applying those, since then there are no more permutations involved in the contraction and it becomes simple matmul.

No worries if the reply is that for now this is the simplest solution and we can improve from there, I mostly just wanted to see if this was on purpose.

leburgel: For `HalfInfiniteEnvironment` and `FullInfiniteEnvironment`, I decided to add in the rotated tensors because I didn't want to write different contractions for every direction and I didn't want to do the permute at every application. The reason that the un-rotated versions are still there, is that for `EnlargedCorner` there are actually different contractions for every direction, so these constructors take un-rotated tensors. I had some routines where I could reuse implementations involving `EnlargedCorner` when dealing with e.g. `HalfInfiniteEnv` by splitting it into parts, at which point I would have had to de-rotate the tensors again. This also felt silly, so then I just decided to keep them both.

We could make a decision to either always rotate or never rotate, but I just didn't feel like implementing all of the rotated contractions for things that are currently unused.

lkdvos: I'm slightly confused by this, would it not just mean that in the `EnlargedCorner` you just have to always use the north direction since you rotated the tensors already? (I did not look at this in a lot of detail, so do correct me if I'm wrong).
As in, I would expect that if you end up rotating the peps sandwich, now the `HalfInfiniteEnv` is effectively always the north one, so it should somehow be possible to always use the north contraction schemes, since the peps sandwhich is the main thing that is not rotation invariant, structure-wise?

leburgel: Just considering the enlarged corners, I think you're right. However, combining this with any other object that needs to be indexed (like isometries of projectors) would then be a bit of a nightmare, since once you put in the rotated tensor and shift the direction the enlarged corner itself has no memory of its original orientation. Maybe this can also be handled, but I would rather not instantiate `EnlargedCorner`s with rotated PEPS tensors for which their direction doesn't match the actual direction in the network. I'm afraid this will inevitably give space mismatches that will be very hard to debug at some point.

If you prefer to hold off on merging this until all sparse environments (`EnlargedCorner`, `HalfInfiniteEnvironment` and `FullInfiniteEnvironment`) are handled in the same way, either only storing rotated tensors or only storing unrotated tensors, then I'm fine with that. But probably that won't be sorted out until we start using the fully sparse decompositions, at which point we'll probably naturally make a choice.

lkdvos (approved):
Since it works and is tested 🤷

PR #360, opened 2026-04-16
Fix FixedSpaceTruncation for simple update

Yue-Zhengyuan:
This PR mainly fixes `FixedSpaceTruncation` for Simple Update, which was broken after generalizing it for arbitrary MPO gates in #339. This introduces some breaking changes for `SiteDependentTruncation` as well:

- I restricted its use to InfinitePEPS/PEPO, so the size of the first dimension must be 2 (horizontal & vertical).
- `truncs[1, :, :]` now refers to horizontal (instead of NORTH/vertical) truncations, following the convention of `SUWeight`.
- ~Rotation is removed, since it is currently no longer used anywhere.~

For bipartite state tests, I refactored the function `_is_bipartite` that checks whether the input object matches the bipartite lattice structure.

Finally, in SU for the ground state, I now also output the energy estimated with SUWeight (equivalently BPEnv). But convergence is still determined from the change in SUWeight (instead of energy).


Yue-Zhengyuan:
Just realized that `SiteDependentTruncation` containing `truncspace` may not work correctly for sectors with different `dual` and `flip`. I'll try to fix it in this PR as well. I plan to first convert `FixedSpaceTruncation` to a `SiteDependentTruncation` with `truncspace`s when constructing the TimeEvolver. Then I can also unify the test for `SiteDependentTruncation` and `FixedSpaceTruncation`.

test/timeevol/sitedep_truncation.jl
```diff
@@ -1,70 +1,35 @@
...
-    @test get_bonddims(peps) == bonddims
-    @test get_bonddims(env) == bonddims
-    # check bipartite structure is preserved
-    for col in 1:2
-        cp1 = PEPSKit._next(col, 2)
-        @test (
-            peps.A[1, col] == peps.A[2, cp1] &&
-                env[1, 1, col] == env[1, 2, cp1] &&
-                env[2, 1, col] == env[2, 2, cp1]
-        )
```
sanderdemeyer, line 45: Is there a reason why you only check whether the states are the same in the new test you added, and not the weights? I assume having weights that do not match the state would throw an error in the time evolution algorithm itself?

test/timeevol/sitedep_truncation.jl
```diff
@@ -1,70 +1,35 @@
...
+    state, wts, = time_evolve(state0, ham, 0.1, 1, alg, wts0)
+    for (t, t0) in zip(state.A, state0.A)
+        @test space(t) == space(t0)
     end
-    env0 = SUWeight(peps0)
-    normalize!.(peps0.A, Inf)
-    # set trunc to be compatible with bipartite structure
-    bonddims = stack([[6 4; 4 6], [5 7; 7 5]]; dims = 1)
-    trunc = SiteDependentTruncation(collect(truncrank(d) for d in bonddims))
-    alg = SimpleUpdate(; trunc, bipartite = true)
```
sanderdemeyer, line 34: The new test only checks `bipartite = false`, it seems. I think both cases are useful to check.

On a related note, it might be useful to have an explicit check whether the `SiteDependentTruncation` is compatible with `bipartite = true` argument, preferably in the `SimpleUpdate` constructor itself.

sanderdemeyer (commented):
Apart from the 2 small comments above, this looks good to me.

test/timeevol/sitedep_truncation.jl
Yue-Zhengyuan: I'll add the check of weight spaces back.

test/timeevol/sitedep_truncation.jl
Yue-Zhengyuan: The check for compatibility of SiteDependentTruncation and `bipartite = true` is currently blocked by the lack of equality definition for `TruncationSpace`.

Yue-Zhengyuan: See https://github.com/QuantumKitHub/TensorKit.jl/issues/407

Yue-Zhengyuan:
@lkdvos Hope to get this merged soon so we can continue with #361.

test/timeevol/sitedep_truncation.jl
sanderdemeyer: In order not to hold this PR back, it's fine for me to leave this test out. When the above issue is fixed, we can always add the test later.

src/algorithms/time_evolution/apply_gate.jl
```diff
@@ -125,20 +127,15 @@ Apply 2-site `gate` on the reduced matrices `a`, `b`
...
         @tensor a2b2[-1 -2; -3 -4] := gate[1 2; -2 -3] * a[-1 1 3] * b[3 2 -4]
     else
         @tensor a2b2[-1 -2; -3 -4] := gate[-2 -3; 1 2] * a[-1 1 3] * b[3 2 -4]
     end
-    trunc = if trunc isa FixedSpaceTruncation
-        need_flip ? truncspace(flip(V)) : truncspace(V)
-    else
-        trunc
-    end
     a, s, b, ϵ = svd_trunc!(a2b2; trunc, alg = LAPACK_QRIteration())
```
lkdvos:
```suggestion
    a, s, b, ϵ = svd_trunc!(a2b2; trunc)
```
After the decomposition rework, it is probably best to either not specify the algorithm, or alternatively at least it should be `QRIteration`

src/algorithms/time_evolution/apply_mpo.jl
```diff
@@ -229,14 +229,8 @@ function _get_allprojs(
...
     @assert length(truncs) == N - 1
-    projs_errs = map(1:(N - 1)) do i
-        trunc = if isa(truncs[i], FixedSpaceTruncation)
-            tspace = space(Ms[i + 1], 1)
-            isdual(tspace) ? truncspace(flip(tspace)) : truncspace(tspace)
-        else
-            truncs[i]
-        end
-        return _proj_from_RL(Rs[i], Ls[i]; trunc)
+    projs_errs = map(zip(Rs, Ls, truncs)) do (R, L, trunc)
```
lkdvos:
```suggestion
    projs_errs = map(Rs, Ls, truncs) do R, L, trunc
```
`map` already supports multiple inputs :)

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -215,15 +219,41 @@ or until convergence of `SUWeight` set by a positive `tol`.
...
     @info "--- Time evolution (simple update), dt = $(it.dt) ---"
-    if check_convergence
-        @assert (it.state.psi isa InfinitePEPS) && it.alg.imaginary_time "Only imaginary time evolution of InfinitePEPS allows convergence checking."
+    env0, time0 = it.state.env, time()
+    for (psi, env, info) in it
+        iter = it.state.iter
+        diff = compare_weights(env0, env)
+        stop = (iter == it.nstep)
+        showinfo = (check_interval > 0) &&
+            ((iter % check_interval == 0) || (iter == 1) || stop)
```
lkdvos, line 232:
```suggestion
            ((iter % check_interval == 1) || stop)
```
Not that this really matters, but in case you want to keep the uniform frequency as well as the first iteration this also works :)

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -217,24 +217,14 @@ updated by the Trotter evolution MPO.
...
-            r, c = mod1(site1[1], Nrow), mod1(site1[2], Ncol)
-            return truncation_strategy(trunc, 2, r, c)
-        else
-            error("The path `sites` contains a long-range bond.")
+        (d, r, c), rev = _nn_bondrev(site1, site2, unitcell)
+        t = truncation_strategy(trunc, d, r, c)
+        if rev && isa(t, TruncationSpace)
+            t = truncspace(flip(t.space)')
         end
+        return t
```
lkdvos, line 228:
```suggestion
```
I don't think this is currently used because of the return on line 223, fine with either style though

src/algorithms/time_evolution/time_evolve.jl
```diff
@@ -94,3 +78,21 @@ function MPSKit.infinite_temperature_density_matrix(H::LocalOperator)
...
+"""
+function _get_fixedspacetrunc(state::InfiniteState)
+    if state isa InfinitePEPO
+        size(state, 3) != 1 && error("Input InfinitePEPO is expected to have only one layer.")
+    end
+    Nr, Nc = size(state)
+    return SiteDependentTruncation(
+        map(Iterators.product(1:2, 1:Nr, 1:Nc)) do (d, r, c)
+            V = domain(state[r, c], (d == 1) ? EAST : NORTH)
+            isdual(V) && (V = flip(V))
```
lkdvos, line 94: Could we use `north_virtualspace` and `east_virtualspace` here?

lkdvos (commented):
Left some minor comments, otherwise looks good to me!

src/algorithms/time_evolution/time_evolve.jl
Yue-Zhengyuan: I prefer to use `domain` here since for standard arrow direction they directly return non-dual spaces.

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: I can't see anything being returned on L223?
```julia
   return map(sites, Iterators.drop(sites, 1)) do site1, site2
        (d, r, c), rev = _nn_bondrev(site1, site2, unitcell) # L223
        t = truncation_strategy(trunc, d, r, c)
        if rev && isa(t, TruncationSpace)
            t = truncspace(flip(t.space)')
        end
        return t
    end
```

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: Say I set `check_interval = 10`. My intention is to output at steps 1, 10, 20, ... instead of 1, 11, 21, ... so I want to keep this unchanged.

src/algorithms/time_evolution/apply_mpo.jl
Yue-Zhengyuan: I'll also eliminated other occurrences of `map(zip(...))`.

src/algorithms/time_evolution/time_evolve.jl
lkdvos: Since you have the check here anyways, does that matter? It's really a lot easier to read what's going on with the full names

src/algorithms/time_evolution/simpleupdate3site.jl
lkdvos: Sorry, I misjudged the indenting on the GitHub view, this is fine :)

src/algorithms/time_evolution/time_evolve.jl
Yue-Zhengyuan: Returning a dual space in the standard case makes construction of `truncspace(V)` below a little bit more involved.

lkdvos: What I mean is that there is already the flip code, so it doesn't really change that much. If you really prefer it this way, that's fine by me, but I would prefer the other way ;)

lkdvos (approved):

Yue-Zhengyuan:
@lkdvos Can't merge because of an expected `build` check.

PR #361, opened 2026-04-16
improve Simpleupdate performances

ogauthe:
This is a work in progress to improve SimpleUpdate performances.

Context: SimpleUpdate is quite slow, especially for 2nd neighbor interaction. I had cases where it became the bottleneck, above CTMRG. I started working on improving the performances last week. I realized many parts could be ported upstream.

I followed 4 different strategies:
- improving type stability
- removing copies where I think they could be avoided
- removing intermediate normalization
- removing permutations by simplifying perm1∘ perm2

This is a work in progress. It already shows significant speed-up for e.g. finite temperature J1-J2 with SU(2) symmetry. I know it is possible to do much better. I just saw #360 which has overlap for `_get_cluster_trunc` and I realized it may be relevant to mention my work (objective is not to block any part of #360)


src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -80,47 +80,79 @@ function _bond_rotation(x, bonddir::Int, rev::Bool; inv::Bool = false)
...
+    invperm = (ntuple(identity, Nax - 1), (Nax,))
+    return t, open_vaxs, invperm
+end
+
 """
 Simple update optimized for nearest neighbor gates
 utilizing reduced bond tensors with the physical leg.
 """
-function _su_iter!(
+function _su_iter_gate!(
```
Yue-Zhengyuan: How about still calling it `_su_iter!` and rely on multiple dispatch?

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -143,24 +175,25 @@ function su_iter(
...
+                    state2[rp1, cp1] = deepcopy(state2[r, c])
+                    state2[rp1, c] = deepcopy(state2[r, cp1])
+                    env2[1, rp1, cp1] = deepcopy(env2[1, r, c])
+                else
+                    rm1, cm1 = _prev(r, Nr), _prev(c, Nc)
+                    state2[rm1, cm1] = deepcopy(state2[r, c])
+                    state2[r, cm1] = deepcopy(state2[rm1, c])
+                    env2[2, rm1, cm1] = deepcopy(env2[2, r, c])
+                end
             end
```
Yue-Zhengyuan: For these I think we can revert to the original to reduce one level of indentation. `deepcopy` can be replaced by `copy` (as in #360).

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -143,24 +175,25 @@ function su_iter(
         elseif length(sites) == 2
             (d, r, c), = _nn_bondrev(sites..., (Nr, Nc))
             alg.bipartite && r > 1 && continue
-            ϵ′ = _su_iter!(state2, gate, env2, sites, alg)
+            ϵ′ = _su_iter_gate!(state2, gate, env2, sites[1], sites[2], alg)
```
Yue-Zhengyuan: Here I wanted `_su_iter!` to be able to handle both `AbstractTensorMap{E, S, 2, 2}` and length-2 MPO gates, though this is mainly for test purposes to ensure that MPO gates are applied correctly even for fermions. Is it possible to recover this behavior without loss of performance?

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -63,122 +63,91 @@ function _nn_bondrev(site1::CartesianIndex{2}, site2::CartesianIndex{2}, (Nrow,
...
+function invbiperm(t::Tuple, ::Val{N}) where {N}
+    p = invperm(t)
+    return p[begin:N], p[(N + 1):end]
+end
 """
 Simple update with an N-site MPO `gate` (N ≥ 2).
 """
-function _su_iter!(
-        state::InfiniteState, gate::Vector{T}, env::SUWeight,
+function _su_iter_mpo!(
```
Yue-Zhengyuan: Again, can we continue calling it `_su_iter!`, and allow it to work with length-2 MPO gates?

Yue-Zhengyuan (commented):
Thank you for improving SU! Left some initial comments.

Yue-Zhengyuan:
There is another recent change that further slowed down SU for 2nd neighbor interaction. After introducing `LocalCircuit` in #347 (and its predecessor `TrotterGates` in #339), each term in the Hamiltonian is *individually* exponentiated to a gate in the `LocalCircuit`. Nearest neighbor and next-nearest neighbor terms are no longer grouped together when exponentiating, resulting in an increase in the number of Trotter gates to be applied.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -216,44 +248,46 @@ or until convergence of `SUWeight` set by a positive `tol`.
...
-        diff = compare_weights(env0, env)
-        stop = (iter == it.nstep) || (diff < tol)
-        showinfo = (check_interval > 0) &&
-            ((iter % check_interval == 0) || (iter == 1) || stop)
-        time1 = time()
-        if showinfo
-            @info "Space of x-weight at [1, 1] = $(space(env[1, 1, 1], 1))"
-            @info @sprintf("SU iter %-7d: |Δλ| = %.3e. Time = %.3f s/it", iter, diff, time1 - time0)
-        end
+    return LoggingExtras.withlevel(; verbosity) do
```
Yue-Zhengyuan, line 219: May I ask in VS Code, how to set a breakpoint inside things wrapped in a `do` block? I have never been able to make the debugger actually stop at such breakpoints...

Yue-Zhengyuan: I can only succeed when I set *everything* to be interpreted, but that will be painfully slow...

Yue-Zhengyuan: Seems to be a known issue of Julia VS Code extension (https://github.com/julia-vscode/julia-vscode/issues/2884). Since time evolution is still in active development, I want to refrain from introducing things that make debugging more complicated.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -63,150 +63,132 @@ function _nn_bondrev(site1::CartesianIndex{2}, site2::CartesianIndex{2}, (Nrow,
     end
 end
 
+"""
+Return a size N-k tuple with values 1 to N but the missing ones. Accept k=1 and k=2.
+"""
+function _filtered_oneto(i, ::Val{N}) where {N}
```
Yue-Zhengyuan: These are generic enough to be put into `src/utility/util.jl`.

src/environments/suweight.jl
Yue-Zhengyuan, line 1: The new `absorb_weight` achieves a huge reduction in the number of allocations. 👍 I suggest we move it and the related helper functions to a new file at `src/algorithms/contractions/absorb_weight.jl`.

test/timeevol/j1j2_finiteT.jl
```diff
@@ -25,7 +25,7 @@ pepo0 = PEPSKit.infinite_temperature_density_matrix(ham)
 wts0 = SUWeight(pepo0)
 # 7 = 1 (spin-0) + 2 x 3 (spin-1)
 trunc_pepo = truncrank(7) & truncerror(; atol = 1.0e-12)
-check_interval = 100
+check_interval = 2^32
```
Yue-Zhengyuan: I personally prefer to display some information in tests. Can you revert it back?

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -63,150 +63,132 @@ function _nn_bondrev(site1::CartesianIndex{2}, site2::CartesianIndex{2}, (Nrow,
...
+    return ntuple(k -> k < i ? k : k + 1, N - 1)
+end
+function _filtered_oneto(i, j, ::Val{N}) where {N}
+    lo, hi = minmax(i, j)
+    return ntuple(k -> k < lo ? k : k < hi - 1 ? k + 1 : k + 2, N - 2)
+end
```
lkdvos: I think this is equivalent to the following functionality from TupleTools, which we already depend on:

```julia
TupleTools.deleteat(ntuple(identity, N), (i, j, ...))
```

src/algorithms/time_evolution/apply_gate.jl
```diff
@@ -48,17 +57,10 @@ When `A`, `B` are PEPOTensors,
         5   1               4  1                                4  1
 ```
 """
-function _qr_bond(A::PT, B::PT; gate_ax::Int = 1, kwargs...) where {PT <: Union{PEPSTensor, PEPOTensor}}
+function _qr_bond(A::PT, B::PT; gate_ax::Integer = 1, kwargs...) where {PT <: Union{PEPSTensor, PEPOTensor}}
```
lkdvos: For these functions, both for type stability and human readability, I would really prefer to just use dispatch and write out the two cases manually. `_get_biperms` is an indirection that doesn't hurt because the compiler decides to inline this, but depending on compiler-internals like this tends to not be great, and honestly I think this function would overall be better suited to simply be two functions with using dispatch and hardcoding the permutations (which could then use `@tensor` calls for the permutations, which is just more readable in general.

src/algorithms/time_evolution/apply_mpo.jl
```diff
@@ -177,26 +177,23 @@ end
...
-    Rs[1], Ls[end] = R_first, L_last
+    Rs = [qr_through(nothing, first(vertices); normalize = true)]
+    Ls = [lq_through(last(vertices), nothing; normalize = true)]
+
     # get remaining R, L matrices
     for n in 2:(N - 1)
-        m = N - n + 1
-        Rs[n] = qr_through(Rs[n - 1], Ms[n]; normalize = true)
-        Ls[m - 1] = lq_through(Ms[m], Ls[m]; normalize = true)
+        push!(Rs, qr_through(last(Rs), vertices[n]; normalize = true))
```
lkdvos: I'm slightly confused that the `push!` construction would be better for type-stability/performance than the original pre-allocated vector. My intuition might also just be wrong though?

src/algorithms/time_evolution/apply_mpo.jl
```diff
@@ -219,30 +216,30 @@ function _proj_from_RL(
...
     end
-    Pas = map(Base.Fix2(getindex, 1), projs_errs)
-    wts = map(Base.Fix2(getindex, 2), projs_errs)
-    Pbs = map(Base.Fix2(getindex, 3), projs_errs)
+    Pas = first.(projs_errs)
+    wts = map(t -> t[2], projs_errs)
+    Pbs = map(t -> t[3], projs_errs)
     # local truncation error on each bond
-    ϵs = map(Base.Fix2(getindex, 4), projs_errs)
+    ϵs = last.(projs_errs)
```
lkdvos: I'm not a huge fan of this change, the previous version was easier to read for me because of how uniform it was. Is this a type stability thing since the anonymous function is better at constant propagating the index? In that case, could you also replace the `first.` and `last.` calls with `t -> t[1]` and `t -> t[4]`?

lkdvos (commented):
Some small comments, will try and actually look into this in detail next week :)

src/algorithms/time_evolution/apply_gate.jl
Yue-Zhengyuan: See latest changes in #144 where I replaced `_qr_bond` with multiple `bond_tensor_xxx` functions ([link](https://github.com/Yue-Zhengyuan/PEPSKit.jl/blob/931a575b5615ead021d5754832bcdaf03799f9bf/src/algorithms/truncation/bond_tensor.jl)).

src/algorithms/time_evolution/simpleupdate.jl
lkdvos: You can always make the anonymous function an actual function, in which you can then set breakpoints as desired. I do like that we have a `verbosity` setting to control this

Yue-Zhengyuan: Well, no. I tried the following:
```julia
using LoggingExtras

function f()
    for x in 1:10
        y = x + 3 # breakpoint here
        @infov 1 "Processed $x..."
    end
    return
end

function test_do1(verbosity::Int = 1)
    return LoggingExtras.withlevel(f; verbosity)
end

function test_do2(verbosity::Int = 1)
    return LoggingExtras.withlevel(; verbosity) do
        f()
    end
end

test_do1()
test_do2()
```
The VS Code debug setting is
```json
"julia.debuggerDefaultCompiled": [    
    "ALL_MODULES_EXCEPT_MAIN", "-LoggingExtras",
],
```
The VS Code debugger still ignores my breakpoint.

Yue-Zhengyuan: I guess it is because there is a do-block in `LoggingExtras.withlevel`.
```julia
function withlevel(f, level::Union{Int, LogLevel}=Info; verbosity::Integer=0)
    with_logger(EarlyFilteredLogger(
        args -> !(args.group isa Verbosity) || verbosity >= args.group.verbosity,
        propagate_level_override(level, current_logger()))
    ) do
        f()
    end
end
```
This internal do-block is still ignored by the VS Code debugger.

src/algorithms/time_evolution/simpleupdate3site.jl
ogauthe: oh nice, I did not know about this

src/algorithms/time_evolution/simpleupdate.jl
lkdvos: In that case the easiest is probably just to manually set `verbosity >= 2 || @info ...` everywhere, but I am not really a big fan of having to shape code such that obscure debugging tools keep working, as that seems somewhat backwards to me. Have you tried [Infiltrator.jl](https://github.com/JuliaDebug/Infiltrator.jl)? I find that in most cases, that is actually what I want, and this is a LOT faster than firing up the Julia debugger

src/algorithms/time_evolution/apply_mpo.jl
ogauthe: indeed this change was not needed for stability. I reverted most of it to reduce diff.

src/algorithms/time_evolution/apply_mpo.jl
ogauthe: Surprisingly, `getindex` is not type stable while `map` is. I now use `map` in all cases.

src/algorithms/time_evolution/simpleupdate.jl
ogauthe: In `su_iter`, we need an `if` statement to distinguish nearest neighbor from long distance at run time. Therefore my opinion is that dispatch does not bring much here, while using different function names makes it explicit and easier to find which function/method is called.

src/algorithms/time_evolution/simpleupdate.jl
ogauthe: Oh I now see the point of using dispatch. There is a traedoff here between code readibility/efficiency (gate is much faster) and this test feature. Is this something that is currently tested?

[copy from above] In su_iter, we need an if statement to distinguish nearest neighbor from long distance at run time. Therefore my idea was that dispatch does not bring much here, while using different function names makes it explicit and easier to find which function/method is called.

src/algorithms/time_evolution/simpleupdate.jl
ogauthe: Ok now I see. Let us discuss this below.

src/algorithms/time_evolution/simpleupdate3site.jl
ogauthe: removed and replaced by TupleTools

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: Maybe it's time for me to step outside the GUI comfort zone...

src/algorithms/contractions/absorb_weight.jl
Yue-Zhengyuan, line 1: Please put the docstring immediately above the `absorb_weight` function.

src/algorithms/time_evolution/simpleupdate.jl
Yue-Zhengyuan: > Is this something that is currently tested?

It is tested with Hubbard model in `test/timeevol/cluster_projectors.jl`.

> gate is much faster

The MPO way can still be optimized. When applying the gate MPO on the first and the last cluster sites, we can also use the trick of reduced bond tensors, which has not been done yet. Even for longer-range (e.g. NNN) bonds, this can still be useful. Separating out two unitary tensors at the two ends of an OBC-MPO will not affect finding the Vidal gauge. 

We can try this out in a follow-up PR if you prefer not to do too much here.

src/algorithms/time_evolution/apply_gate.jl
Yue-Zhengyuan: @ogauthe You can have a look at related parts of #144 and port them here if you wish.

Yue-Zhengyuan:
@ogauthe Do you mind if I temporarily take over to merge upstream changes and fix tests?

ogauthe:
> @ogauthe Do you mind if I temporarily take over to merge upstream changes and fix tests?

I just merged master, go ahead if there are other changes you plan for.

src/algorithms/time_evolution/simpleupdate.jl
ogauthe: Ok I will revert to dispatch. Keeping this conversation, closing the other relative to this one.

src/algorithms/time_evolution/simpleupdate3site.jl
ogauthe: closed to keep conversation above.

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -86,47 +86,81 @@ function _bond_rotation(x, bonddir::Int, rev::Bool; inv::Bool = false)
...
+    A = _bond_rotation(A0, dir, rev; inv = false)
+    B = _bond_rotation(B0, dir, rev; inv = false)
     # apply gate
-    ϵ, s = 0.0, nothing
+    ϵ = 0.0
+    local s
     gate_axs = alg.purified ? (1:1) : (1:2)
-    for gate_ax in gate_axs
-        X, a, b, Y = _qr_bond(A, B; gate_ax, positive = true)
+    for gate_ax in gate_axs  # TODO try to use type stable helper function
```
Yue-Zhengyuan: Will it be better if I always do `for gate_ax in 1:2`, and manually break after the first loop when `alg.purified === true`?

src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -86,47 +86,81 @@ function _bond_rotation(x, bonddir::Int, rev::Bool; inv::Bool = false)
...
-    B = _bond_rotation(B, bond[1], rev; inv = true)
     rev && (s = transpose(s))
-    # remove environment weights
-    siteA, siteB = map(sites) do site
-        return CartesianIndex(mod1(site[1], Nr), mod1(site[2], Nc))
+    # rotate back & remove environment weights
+    for (site, vertex, open_vaxs) in ((siteA, A, open_vaxs_A), (siteB, B, open_vaxs_B))
+        s′ = (mod1(site[1], Nr), mod1(site[2], Nc))
+        rotated = _bond_rotation(vertex, dir, rev; inv = true)
+        state[s′...] = absorb_weight(rotated, env, s′..., open_vaxs; inv = true)
```
Yue-Zhengyuan: The state tensors need normalization, especially when evolving for the ground state; otherwise, the norm will explode.

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -196,19 +153,28 @@ function _su_iter!(
...
+    s′ = CartesianIndex(mod1(first(sites)[1], Nr), mod1(first(sites)[2], Nc))
+    leftpermuted = permute(first(Ms), left_invperm)
+    state[s′] = absorb_weight(leftpermuted, env, s′, left_vaxs; inv = true)
+
+    # right
+    s′ = CartesianIndex(mod1(last(sites)[1], Nr), mod1(last(sites)[2], Nc))
+    rightpermuted = permute(last(Ms), right_invperm)
+    state[s′] = absorb_weight(rightpermuted, env, s′, right_vaxs; inv = true)
+
+    for (vertex, s, invperm, vaxs) in zip(Ms[(begin + 1):(end - 1)], sites[(begin + 1):(end - 1)], map(t -> t[3], mids), map(t -> t[2], mids))
```
Yue-Zhengyuan: @ogauthe Is it really *that* important to separate out the first and last tensors for type stability because they have 3 instead of 2 open virtual legs?

src/algorithms/time_evolution/simpleupdate3site.jl
```diff
@@ -68,126 +68,83 @@ Find the permutation to permute `out_ax`, `in_ax` legs to
...
-        end
-        wts, ϵs, = _cluster_truncate!(Ms, truncs)
-        if isa(state, InfinitePEPO)
-            Ms = [first(_unfuse_physicalspace(M, Vphy)) for (M, Vphy) in zip(Ms, Vphys)]
-        end
+    ϵ = 0.0
+    local wts
+    for gate_ax in 1:2
+        _apply_gatempo!(Ms, gates; gate_ax)
+        wts, ϵs = _cluster_truncate!(Ms, truncs)
```
Yue-Zhengyuan: I just realized that I have recently updated `_cluster_truncate!` in #348, so it can handle MPSs with any number of physical legs (and even different numbers of physical legs at each site). So no longer need to first fuse the physical legs for PEPO.

Yue-Zhengyuan:
> I know it is possible to do much better.

@ogauthe Do you mind briefly outlining what can be further improved starting from the current PR?

src/algorithms/time_evolution/simpleupdate3site.jl
Yue-Zhengyuan: Can I mitigate the problem by wrapping contents in the for-loop as a function?

Yue-Zhengyuan: I think the speedup for the J1-J2 test mainly comes from the improvement in applying nearest neighbor gates. In my test, separating out the first and the last tensors of the cluster offers little gain in performance, especially since we currently only have examples with up to 3-site MPO gates. To reduce redundancy, I'll restore the old approach, but with more things wrapped inside dedicated functions.

lkdvos: I think indeed a function barrier can mitigate some of it, but it will still give a dynamic dispatch. This might completely be fine though, that by itself doesn't cost too much as long as the type instability doesn't leak out of that region and taint the subsequent code

src/algorithms/contractions/bondenv/benv_ctm.jl
```diff
@@ -24,8 +24,8 @@ Axis order: `[DX1 DY1; DX0 DY0]`, as in
 ```
 """
 function bondenv_ctm(
-        row::Int, col::Int, X::T, Y::T, env::CTMRGEnv
-    ) where {T}
+        row::Int, col::Int, X::TX, Y::TY, env::CTMRGEnv
+    ) where {TX, TY}
```
lkdvos:
```suggestion
        row::Int, col::Int, X, Y, env::CTMRGEnv
    )
```
I don't think there is any advantage in specifying types like this?

lkdvos (approved):

src/algorithms/contractions/bondenv/benv_ctm.jl
Yue-Zhengyuan: This happens in NTU with NNN gates where you decide to keep physical leg in Y but not do so in X

lkdvos: Yes, but specifying x::TX where TX is the same as not specifying anything, and I prefer to then just not clutter the signature

PR #362, opened 2026-04-18
Move `trunc` out of bond truncation algorithm

Yue-Zhengyuan:
This PR moves `trunc::TruncationStrategy` out of bond truncation algorithms `ALSTruncation` and `FullEnvTruncation`. 

The motivation is to make supporting `FixedSpaceTruncation` and `SiteDependentTruncation` in NTU (#144) easier, so I don't need to reconstruct the bond truncation algorithm struct every time I switch the bond to be truncated (since each bond have a different `trunc`).

### DIFF
--- a/src/algorithms/truncation/bond_truncation.jl
+++ b/src/algorithms/truncation/bond_truncation.jl
@@ -13,13 +13,11 @@ $(TYPEDFIELDS)
 
 The truncation algorithm can be constructed from the following keyword arguments:
 
-* `trunc::TruncationStrategy`: SVD truncation strategy when initilizing the truncated tensors connected by the bond.
 * `maxiter::Int=50` : Maximal number of ALS iterations.
 * `tol::Float64=1e-9` : ALS converges when the relative change in bond SVD spectrum between two iterations is smaller than `tol`.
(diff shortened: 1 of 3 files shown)


lkdvos (commented):
This does look in general quite different from how we structure the rest of our function signatures, which typically have the truncation strategy in the algorithm... I'm not sure if I really understand the benefit, isn't this just a matter of using `@set alg.trunc = ...`?

Yue-Zhengyuan:
Currently the NTU struct looks like
```julia
@kwdef struct NeighbourUpdate <: TimeEvolution
    "Bond truncation algorithm after applying time evolution gate"
    opt_alg::Union{ALSTruncation, FullEnvTruncation} =
        ALSTruncation(; trunc = truncerror(; atol = 1.0e-10))
    ...
end
```
I feel it a bit weird to set `trunc = FixedSpaceTruncation()` in `opt_alg`, since the `bond_truncate` function that ultimately uses `opt_alg` does not know the original bond space before applying the gate, and FixedSpaceTruncation does not make sense in this narrow context. So I want to make `trunc` directly a field in `NeighbourUpdate`.

Though I admit that this is still not ideal, and I'm surely open to suggestions.

Issue #354, opened 2026-04-11
`info_ctmrg` fields in `leading_boundary`


ogauthe:
Hi!

When running CTMRG with `leading_boundary`, the final convergence accuracy  `η`  is an important information, especially if the loop stopped due to reaching `alg.maxiter`. It would be nice if `η` was included into `info_ctmrg`, currently it cannot be accessed from outside. I can make a PR if you like the idea.

I also realized that `info_ctmrg` contains `U_full` and `V_full` for a total memory cost $32\chi^2D^4$ on a 2x2 unit cell. This is a lot. I understand these tensors are used in AD, where the pullback is only computed on a fully converged environment. If I am correct, it is not used in the (many) converging steps before calling AD, nor when AD is not needed at all. Would it be possible to use a dedicated function in AD that would also return  `U_full` and `V_full` , but keep them out of the regular `leading_boundary`? The memory gain is clear but I am unsure how much code duplication this would generate, there is a tradeoff to be made.
EDIT: I realize this is only for `SimultaneousCTMRG`: `SequentialCTMRG` does not return these tensors.

Yue-Zhengyuan:
> It would be nice if `η` was included into `info_ctmrg`

This is included in @leburgel's PR #264 (which has been delayed for quite some time). Though you can still first make a PR if you need this right now.

leburgel:
In the meantime there were some other requests for additional convergence info, in particular returning the final convergence error as well as a dedicated convergence flag to easily check if a contraction actually converged without having to increase the verbosity. I'll split this off from #264 and file a dedicated PR which also contains the convergence flag, no reason to have to wait until that is merged for something that simple.

About the full decompositions that are being kept around, it's not as simple as dropping them from all but the final CTMRG iteration unfortunately. By default we use a fixed-point differentiation approach, in which case we indeed only use the full decompositions in a single iteration that is being differentiated at the fixed point. For this case, we can keep the full decomposition around only in the pullback, and completely remove it from the forward pass. However, we also support a naive AD mode where we really backpropagate through all of the iterations, in which case we do need the full decomposition in every iteration. But also then, we could only keep the full decomposition around in the pullback of the iterations, removing them from the forward pass.

[...]

PR #364, opened 2026-04-21
Update decomposition handling

leburgel:
Updates the handling of decompositions after the updates of MatrixAlgebraKit.jl v0.6.5. The main purpose was to remove all explicit `LAPACK_` specifications from the decomposition algorithms, now that the decomposition algorithms themselves have been decoupled from the storagetype  in MatrixAlgebraKit.jl. In the process, I also tried to streamline the way we use decompositions a bit, attempting to bring them more in line with MatrixAlgebraKit.jl while still allowing to hook into the implementation and switch between different reverse rule algorithms.

## Decomposition algorithms & selection

The main change is an update in how decomposition algorithms are selected.

- All of the MatrixAlgebraKit.jl decompositions are now selected through symbols that exactly match their algorithm name, e.g. `alg = :DivideAndConquer`.
    - This seems the most appropriate, since it means we don't need to introduce any new symbols ourselves like we do now, and we can easily refer to the MatrixAlgebraKit.jl docs.
    - The downside is that we still explicitly list all of the available algorithms in a symbol list, and we'll run into trouble if the names in MatrixAlgebraKit.jl change.
    - By default, we now just use `:DefaultAlgorithm` whenever we can, which seems like the most suitable option.

- The specification of the iterative algorithms is kept the same as it was before.

- I updated the docstrings of the decomposition algorithm structs `SVDAdjoint`, `EighAdjoint`, `QRAdjoint` to reflect all of these changes and also link directly to the relevant docstrings so users can easily find wat keyword algorithms they can provide.

- I removed the `rrule_alg` specification from the user-facing interface in `leading_boundary` and `fixedpoint`.
    - In this update the `decomposition_alg` kwarg of `leading_boundary` or `CTMRGAlgorithm` is interpreted as a **forward** algorithm. Specifying a particular rrule algorithm as well can now only be done by passing a full decomposition struct, e.g. an instantiated `SVDAdjoint`, and is considered expert usage.
    - The default rrule algorithm is now selected based on the forward algorithm, using the default that is the most logical (i.e. full rrule when we have the full decomposition, truncated one for the sparse decompositions).
    - This is in anticipation of the fact that we'll move to directly using the MatrixAlgebraKit.jl differentiation without any manual intervention, so I wanted to remove this layer already from the interface that most users interact with. Advance usage for prototyping is still possible, so this seems a good solution for now.

## Decomposition implementation

I updated the decomposition implementation to remove the intermediate `_svd_trunc!`, `_eigh_trunc!` and `_left_orth!` methods, and only use the MatrixAlgebraKit.jl methods directly.

- The call signature for `svd_trunc!`, `eigh_trunc!` and `left_orth!` using `SVDAdjoint`, `EighAdjoint` and `QRAdjoint` algorithms respectively now just matches the MatrixAlgebraKit.jl signature.
   - This means they no longer return a generic `info` struct for the truncated decompositions, but only return the truncation error. Additional info such as condition numbers is just computed manually afterwards and added directly to the CTMRG info. In particular, the full decomposition is no longer returned explicitly (see also below).
   - On the other hand, this means I had to add the `trunc` specification to the actual decomposition structs, since MatrixAlgebraKit doesn't allow mixing instantiated algorithm structs with keyword arguments. In the forward pass, the MatrixAlgebraKit implementation is then called directly using a `TruncatedAlgorithm(alg.fwd_alg, alg.trunc)`, which seemed like the cleanest way to do things.

- The reverse rules are still handled by manual interception at the moment, even though they just directly call the MAtrixAlgebraKit.jl implementations.
    - The main reason is that this is currently the only way to use the keyword arguments of the pullbacks to control things like the verbosity.
    - Once QuantumKitHub/MatrixAlgebraKit.jl#207 has been addressed, we can probably stop handling things this way and remove the `SVDAdjoint`, `EighAdjoint` and `QRAdjoint` structs altogether. Then we only have to handle reverse rules for the decompositions defined directly in PEPSKit.jl, i.e. the sparse iterative ones (until they are properly implemented elsewhere of course).

## Gauge fixing and fixed-point differentiation

I updated the implementation of the fixed iteration scheme to not use fixed precomputed decompositions, but rather fix relative and absolute phases after the core iteration using signs precomputed in the gauge fixing of the environment itself.

- I managed to remove `FixedSVD`, `FixedEigh` and `FixedQR` completely, which in turn means we never have to return, store or gauge-fix full untruncated decompositions anymore.
    - In particular, this would address part of #354 since the full decompositions are only ever stored when using differentiation, completely removing them from the regular forward pass.

- To make this work properly, I had to ensure that running the same decomposition on the same input twice actually gave the same result, otherwise we end up gauge fixing with the wrong signs.
    - For the full decompositions, this worked fine right out of the box.
    - For the sparse iterative decompositions, I had to:
        - Add a `MatrixAlgebraKit.gaugefix!(...)` call when computing the decomposition in each block.
        - Switch to using deterministic start vectors for the Krylov procedure, using `ones(...)` rather than `randn` to generate the start vector.

- I updated the settings in the tests and examples to no longer use `iterscheme = :diffgauge`, and also reduced its usage in the gradient tests directly. Things are quite a bit faster now.
   - Since `iterscheme = :diffgauge` is so much slower in practice, maybe we should just consider removing it altogether. In practice, it's sometimes just unusable, while all of the initial stability issues with `iterscheme = :fixed` seem to be resolved now.

## Further questions

Things to address before we can merge (and hopefully tag a release):

- The most important thing now is to fix the interface to something we can stick with for a while. At this point this basically comes down to settling on a set of symbols that can be used to configure algorithms, and then hopefully not change them.
    - Right now, there's a mix of capitalized and lower-case symbols, e.g. `:DivideAndConquer` for specifying an SVD algorithm versus `:gmres` for specifying a gradient linear solver. I would suggest to:
    - Just use the capitalization of the struct names whenever we reference an algorithm that is not implemented in PEPSKit.jl itself. MatrixAlgebraKit.jl is the main motivation to do this, since `alg = :DivideAndConquer` works directly in the MatrixAlgebraKit.jl interface itself, so it makes sense to just use the same. However, this mean changing some current values (e.g. `:gmres -> :GMRES`,  `:lbfgs -> :LBFGS`. Very breaking, but this seems like a good time to break.

Any other comments and questions are very welcome.


pbrehmer:
Wow, this is really nice, thanks Lander! I can give some opinions right off the bat:
- I love that the `FixedSVD`, `FixedEigh` and `FixedQR` structs are gone, this was a hassle
- I would be in favor of removing the `:diffgauge` scheme. It was there mostly for experimental purposes and only really emerged in the very beginning phases of PEPSKit when PEPS people hadn't really figured out fixed-point differentiation that well yet.
- Regarding capitalization: I like your proposal but wouldn't it be even more consistent if we also capitalize our own algorithms? For example `:halfinfinite -> :HalfInfinite` or `:simultaneous -> :Simultaneous` would make sense to me. I don't really mind that this is very breaking since the release will anyway be really breaking.

I will take a proper look tomorrow and add some more comments, but this definitely is a massive improvement.

docs/src/examples/j1j2_su/index.md
Yue-Zhengyuan, line 1: Something is wrong here. The file becomes too big.

leburgel: There was some instability in the variational optimization that triggered the warnings in the SVD pullback to go off and clutter up the logs. I moved the random seed by one, seems to have resolved the issue.

Yue-Zhengyuan:
> * I would be in favor of removing the `:diffgauge` scheme

@pbrehmer I don't know the details yet, but I remember that `:diffgauge` is required if one wants to restrict the state to be real?

pbrehmer:
> > * I would be in favor of removing the `:diffgauge` scheme
> 
> @pbrehmer I don't know the details yet, but I remember that `:diffgauge` is required if one wants to restrict the state to be real?

This was related to the fact that the gauge fixing routine in `:fixed` mode (which would previously `FixedSVD` etc.) would generally produce complex gauge signs such that real tensor would turn into complex ones within the first step of optimization. I haven't looked at the updated gauge fixing procedure but I would assume that it still produces complex gauges?

I would still be in favor of removing `:diffgauge` but maybe just add a gauge fixing method which will leave the tensors real if the current procedure is not compatible with real tensors.

leburgel:
> I would still be in favor of removing `:diffgauge` but maybe just add a gauge fixing method which will leave the tensors real if the current procedure is not compatible with real tensors.

If we can make sure the sign matrices are real at the end of the gauge fixing procedure for real environments, we should be able to use `:fixed` with real tensors. So that sounds like the way to go. I'm anyways not going to remove it here, just asking for opinions and then I'll address it in a follow-up.

src/algorithms/ctmrg/projectors.jl
```diff
@@ -178,16 +158,22 @@ end
...
     halfinf = half_infinite_environment(enlarged_corners...)
-    svd_alg = decomposition_algorithm(alg, coordinate)
-    U, S, V, info = svd_trunc!(halfinf / norm(halfinf), svd_alg; trunc = alg.trunc)
+    svd_alg = decomposition_algorithm(alg)
+    U, S, V, truncation_error = svd_trunc!(halfinf / norm(halfinf), svd_alg)
+
+    # get some decomposition info
+    truncation_error = truncation_error / norm(S) # normalize truncation error
+    condition_number = ignore_derivatives() do
+        return cond(S)
```
lkdvos: I might be mistaken here, but there is a subtle change associated with this PR that now we are computing the condition number of the truncated SVD, not of the full SVD. Since `cond` boils down to computing `max/min` of the values, this does actually affect the output significantly.
The main thing is that we don't always have access to this anyways for the iterative or randomized decompositions, so I am not advocating that we need to fix this, but it might be a bit misleading to really call this `condition_number`

src/algorithms/ctmrg/gaugefix.jl
```diff
@@ -85,9 +69,11 @@ function gauge_fix(envfinal::CTMRGEnv{C, T}, envprev::CTMRGEnv{C, T}, ::Scrambli
     end
 
     env_fixed_relative = fix_relative_phases(envfinal, signs)
-    env_fixed_full = fix_global_phases(env_fixed_relative, envprev)
 
-    return env_fixed_full, signs
+    # find additional global phases
+    env_fixed_full, corner_phases, edge_phases = fix_global_phases(env_fixed_relative, envprev)
+
+    return env_fixed_full, signs, corner_phases, edge_phases
```
lkdvos: I'm still slightly unhappy with our functions and there purpose here, in the sense that it now seems like you can't compute the phases/signs without also gauge fixing the environment. Is that required? I was staring at this briefly and I guess the phases are determined after the signs are used for fixing, so there is a non-trivial interaction there, but it would definitely be clearer in my head if the callstack looked more like:

```julia
function gauge_fix!(...)
    signs, corner_phases, edge_phases = compute_gauge_fix_gauge(...)
    return fix_phases(...)
end
```

which also avoids us having to actually perform the gaugefix at the end of the CTMRG run while we discard the gaugefixed environment anyways (I think?)

---

Just to repeat: if this does not really work, feel free to just close this comment, I'm just trying to organize things in my brain a bit as well :)

src/utility/eigh.jl
```diff
@@ -272,7 +243,7 @@ function _compute_eighdata!(
         howmany = trunc isa NoTruncation ? minimum(size(b)) : blockdim(trunc.space, c)
 
         if howmany / minimum(size(b)) > alg.fallback_threshold  # Use dense decomposition for small blocks
-            D, V = eigh_full!(b, LAPACK_QRIteration())
+            D, V = eigh_full!(b, QRIteration())
```
pbrehmer: This should probably be `Defaults.eigh_fwd_alg`?

src/utility/eigh.jl
```diff
@@ -285,7 +256,7 @@ function _compute_eighdata!(
             D, lvecs, info = eigsolve(b, x₀, howmany, :LM, eig_alg)
             if info.converged < howmany  # Fall back to dense SVD if not properly converged
                 @warn "Iterative eigendecomposition did not converge for block $c, falling back to eigh_full"
-                D, V = eigh_full!(b, LAPACK_QRIteration())
+                D, V = eigh_full!(b, QRIteration())
```
pbrehmer: Is there a reason why we want to hardcode `QRIteration` instead of `Defaults.eigh_fwd_alg`?

src/utility/svd.jl
```diff
@@ -248,40 +220,36 @@ Construct an `IterSVD` algorithm struct based on the following keyword arguments
...
 # TODO: redefine _empty_svdtensors, _create_svdtensors
-function _svd_trunc!(f, alg::IterSVD, trunc::TruncationStrategy)
+function MatrixAlgebraKit.svd_trunc!(f, alg::TruncatedAlgorithm{<:IterSVD})
+    fwd_alg = alg.alg
+    trunc = alg.trunc
     U, S, V = if isempty(blocksectors(f))
         # early return
         truncation_error = zero(real(scalartype(f)))
-        MatrixAlgebraKit.initialize_output(svd_compact!, f, LAPACK_QRIteration()) # specified algorithm doesn't matter here
+        MatrixAlgebraKit.initialize_output(svd_compact!, f, QRIteration()) # specified algorithm doesn't matter here
```
pbrehmer: Same comment, why not `Defaults.svd_fwd_alg`?

src/Defaults.jl
```diff
@@ -22,30 +22,40 @@ Module containing default algorithm parameter values and arguments.
...
-* `rrule_degeneracy_atol=$(Defaults.rrule_degeneracy_atol)` : Broadening amplitude which smoothens the divergent term in the retained contributions of an SVD or eigh pullback, in case of (pseudo) degenerate singular values
+* `rrule_degeneracy_atol=$(Defaults.rrule_degeneracy_atol)` : Broadening amplitude which
+  smoothens the divergent term in the retained contributions of an SVD or eigh pullback, in
+  case of (pseudo) degenerate singular values
 * `svd_fwd_alg=:$(Defaults.svd_fwd_alg)` : SVD algorithm that is used in the forward pass.
-    - `:sdd` : MatrixAlgebraKit's `LAPACK_DivideAndConquer`
-    - `:svd` : MatrixAlgebraKit's `LAPACK_QRIteration`
-    - `:iterative` : Iterative SVD only computing the specifed number of singular values and vectors, see [`IterSVD`](@ref PEPSKit.IterSVD)
+    - `:DefaultAlgorithm` : MatrixAlgebraKit's default SVD algorithm for a given matrix type.
+    - `:DivideAndConquer` : MatrixAlgebraKit's [`DivideAndConquer`](@extref MatrixAlgebraKit.DivideAndConquer)
```
pbrehmer, line 30: I would suggest that we replace `:DivideAndConquer` by `:SafeDivideAndConquer` because there have occasionally been LAPACK errors related to the `sdd` algorithm in PEPSKit I think.

test/utility/eigh_wrapper.jl
```diff
@@ -23,9 +24,9 @@ r = 0.5 * (r + r') # make r Hermitian
 R = randn(space(r))
 R = 0.5 * (R + R')
 
-full_alg = EighAdjoint(; fwd_alg = (; alg = :qriteration), rrule_alg = (; alg = :full))
-trunc_alg = EighAdjoint(; fwd_alg = (; alg = :qriteration), rrule_alg = (; alg = :trunc))
-iter_alg = EighAdjoint(; fwd_alg = (; alg = :lanczos), rrule_alg = (; alg = :trunc))
+full_alg = EighAdjoint(; fwd_alg = (; alg = :QRIteration), rrule_alg = (; alg = :full))
```
pbrehmer, line 29: This should also probably be `Defaults.eigh_fwd_alg` (not that it matters much)

src/algorithms/ctmrg/gaugefix.jl
pbrehmer: The thing that does not allow this separation of computing the gauge signs and fixing the phases is the computation of the global phases in `fix_global_phases`. There we take the dot product of the previous environment and the environment with fixed relative phases. To my knowledge, this is the easiest way of fixing these global signs.

While I do agree that your suggested callstack would make things clearer and it would be nice to not discard the gauge-fixed environment, I don't thing it would make a big difference in practice and performance.

src/utility/eigh.jl
```diff
@@ -58,35 +57,63 @@ $(TYPEDFIELDS)
...
+!!! note
+    Manually specifying a `rrule_alg` is considered expert-mode usage, and should only be done when full control over the implementation is desired.
+    For all regular use cases, the default reverse rule algorithms, automatically chosen based on the forward algorithm, should be sufficient.
 """
-struct EighAdjoint{F, R}
+struct EighAdjoint{F, R, T}
     fwd_alg::F
     rrule_alg::R
-end  # Keep truncation algorithm separate to be able to specify CTMRG dependent information
+    trunc::T
```
pbrehmer: Since we now keep the truncation algorithm inside `EighAdjoint` and `SVDAdjoint` (which I like), why not get rid of the `trunc` field in the `ProjectorAlgorithm`s? In the constructor/algorithm selection of the CTMRG algorithms, the `trunc` argument could directly be passed onto the decomposition algorithms. Then it might make sense to create a `DecompositionAlgorithm` abstract type and do the algorithm selection in there.

src/algorithms/ctmrg/projectors.jl
```diff
@@ -52,33 +52,13 @@ end
...
-                nothing, nothing, nothing, nothing,
-            )
-        end
-        return SVDAdjoint(; fwd_alg = fix_svd, rrule_alg = decomposition_alg.rrule_alg)
-    else
-        return decomposition_alg
-    end
+function decomposition_algorithm(alg::ProjectorAlgorithm)
+    decomposition_alg = alg.decomposition_alg
+    decomposition_alg = @set decomposition_alg.trunc = alg.trunc
```
pbrehmer: This feels a bit redundant and there I would prefer if the truncation algorithm would directly be stored inside the decomposition algorithm (see the other comment).

pbrehmer (commented):
Again I'll have to say that I really like this! This is a big step forward in cleaning up the codebase and making things more consistent. I only have some minor comments. One thing I'm still unsure about is how we should best handle the capitalization of the algorithms since I do feel a bit torn about mixing capitalized MAK-style algorithms with lowercase ones.

src/algorithms/ctmrg/projectors.jl
leburgel: Since this is the only thing we can do for sparse truncated decompositions anyway, I guess it makes more sense to change it like this. Although it definitely wasn't intentional, I was just trying to move all info besides the truncation error out of the `svd_trunc!` output.

I'm not sure what the condition number was used for in practice. @pbrehmer, does this change affect anything important that I didn't consider?

src/algorithms/ctmrg/gaugefix.jl
leburgel: I'm fine with splitting the procedure into first computing the signs and phases and then multiplying everything in.

src/utility/eigh.jl
leburgel: I would be fine with this for now, in the long run I don't think it's safe to assume that the default eigh algorithm is a full decomposition that works with `eigh_full!`. I'd prefer to just remove the algorithm specification altogether, and just use the MatrixAlgebraKit default directly.

src/utility/eigh.jl
leburgel: It was hardcoded as `QRIteration` before, I have no idea why. I just removed the `LAPACK_` part without changing anything else. I would suggest to remove the algorithm altogether then, and just use the MatrixAlgebraKit default.

src/utility/svd.jl
leburgel: Same suggestion as above.

src/Defaults.jl
leburgel: `:DivideAndConquer` is never used by default now. The default behavior is`:DefaultAlgorithm`, which corresponds to `:SafeDivideAndConquer` for CPU-based storage types. `:DivideAndConquer` is just listed here as one of the available MatrixAlgebraKit algorithms, but it's no longer used anywhere.

test/utility/eigh_wrapper.jl
leburgel: I just changed the capitalization without changing anything else. But thinking about it, this would end up being very confusion if we ever change the default from a full to a sparse decomposition. So I would actually rather leave it like this.

src/utility/eigh.jl
leburgel: I did consider this, but in the end decided not to. The main reason is that there are truncation schemes such as `FixedSpaceTruncation` and `SiteDependentTruncation` (that we use very often) that only really make sense in the context of CTMRG projector computation, while the decomposition algorithms such as `SVDAdjoint` and `EighAdjoint` require a "concrete" truncation strategy such as `TruncationSpace` or `TruncationByOrder` in order to make sense. For example, an `SVDAdjoint` with `trunc::FixedSpaceTruncation` makes no sense, since at the point where `svd_trunc` is called there's no way to figure out what space is being fixed.

In other words, since the `trunc` field of the decomposition algorithms has no knowledge of the CTMRG environment or sites in the unit cell (and I think this is good, I don't think it should), it made sense to me to systematically set this using a concrete truncation strategy derived from the projector truncation strategy (which does know about the environment and its spaces).

If there's a better way to handle this I'd be happy to reconsider, but I want to avoid instantiating an `SVDAdjoint` with `trunc::FixedSpaceTruncation` since that doesn't actually make sense to me.

src/algorithms/ctmrg/projectors.jl
leburgel: See the reply to the comment above. I agree this looks a bit weird though. Instead, we can set the concrete truncation algorithm direction in the `ðecomposition_alg` field instead of updating the `trunc` field of the projector algorithm (which is what is done one layer above here), and then just return the `decomposition_alg` directly.

src/utility/eigh.jl
lkdvos: I agree with this, we really don't want the default, we want any "dense" fallback, so I would agree with @leburgel to just remove the algorithm altogether.

src/algorithms/ctmrg/gaugefix.jl
lkdvos: Can we take the dot product of the previous environment and the gaugefixed environment without actually fixing the gauge? Schematically, something like `dot(env, phase, env)`

src/algorithms/ctmrg/projectors.jl
lkdvos: I think it was only there for debug information, but I don't really like calling it `condition_number` if it isn't that, although in practice it might not matter

leburgel: If we can't make it make sense in general, we could also consider just removing it. Is there any use for `cond` of the truncated decomposition, even for debugging?

src/algorithms/ctmrg/gaugefix.jl
leburgel: The simplest way to implement this `dot` is to just multiply in the signs  on one side first and then taking a regular `dot`, so in that sense I don't think there's a way of doing this that just comes down to gauge fixing but not returning the gauge fixed environment. Which I'm actually fine with.

leburgel: I split off the sign and phase computation. This still involves fixing relative phases in an intermediate step, but I don't think it's worth the effort of trying to come up with a way around this. Let me know if it looks better this way.

src/utility/eigh.jl
leburgel: I removed the algorithm, so we now use the dense default.

src/utility/svd.jl
leburgel: I didn't notice before, but the algorithm doesn't even matter here since the only thing being done here is initializing the output (see the comment in the code).

I replaced it with `DefaultAlgorithm()` to be a bit more generic, but this has no practical effect.

leburgel:
It seems the updated `svd_trunc` pullback in MatrixAlgebraKit v0.6.6 is giving some trouble: https://github.com/QuantumKitHub/PEPSKit.jl/actions/runs/24983503557/job/73151035377?pr=364

As far as I can tell this was a perfectly working test before, so the new pullback somehow gives `NaN`s in some weird edge cases. Maybe it would be good to still have a slower fallback in case the new implementation fails?

pbrehmer:
> It seems the updated `svd_trunc` pullback in MatrixAlgebraKit v0.6.6 is giving some trouble: https://github.com/QuantumKitHub/PEPSKit.jl/actions/runs/24983503557/job/73151035377?pr=364
> 
> As far as I can tell this was a perfectly working test before, so the new pullback somehow gives `NaN`s in some weird edge cases. Maybe it would be good to still have a slower fallback in case the new implementation fails?

This is the case where we artificially create a degenerate spectrum so it does make sense that the Sylvester solver is struggling but it even seems to struggle in the presence of Lorentzian broadening (the second error), so that is odd to me. I'm not sure why the previous Sylvester solver would produce a converging result and the current one doesn't.

src/algorithms/ctmrg/projectors.jl
pbrehmer: Indeed, this is debug info/analytics. I was using this in some cases where I wanted to see the ratio between the largest and smallest singular value but given that the truncation error is anyway normalized and will give you an idea of the rough conditioning, I would be fine with just removing it.

leburgel: Okay, then I'll just remove it for now, seems like the simplest solution.

src/algorithms/ctmrg/projectors.jl
pbrehmer: Maybe I'd prefer setting the `trunc` field already one layer above but that's of course only a minor nitpicky comment.

leburgel: I updated the approach to set the concrete truncation strategy directly into `alg.decomposition_alg.trunc`, which removes the step here. I think I like this better, the projector algorithm never has its `trunc` field updated, but instead updates the truncation in its decomposition algorithm.

src/utility/eigh.jl
pbrehmer: I see, that does make sense. Then I'm okay with keeping it like this. But perhaps we could add some comment or description in the docstring that makes it clearer that the `trunc` contained in the `ProjectorAlgorithm` may include contraction algorithm specific truncation modes which will not translate to a truncated decomposition directly.

One alternative could be that we remove the `trunc` field in `SVDAdjoint` / `EighAdjoint` and instead directly instantiate a `TruncatedAlgorithm` as the `fwd_alg` field (which anyway is done later in the call to MatrixAlgebraKit. I'm not sure if that improves things but maybe this is a bit more consistent with MAK.

leburgel:
It seems like the new `svd_trunc` pullback uses the `degeneracy_atol` as a convergence measure in its Sylvester solver loop:
https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/5f134300121729272a86b8250aff1a66dea226c5/src/pullbacks/svd.jl#L251

This probably means that the way we disable the broadening and cutoff in the SVD wrapper test
https://github.com/QuantumKitHub/PEPSKit.jl/blob/5a103ddb58121fa1cd07e1327fcb64508328f65d/test/utility/svd_wrapper.jl#L118
now has some unintended consequences in the actual pullback implementation.

lkdvos:
Hahahah yes, we should pick a reasonable value there, which probably fixes the issues. I'm assuming that was there since before we had issues if this was chosen too small, but I think this might be better now?

src/utility/eigh.jl
leburgel: > But perhaps we could add some comment or description in the docstring that makes it clearer that the `trunc` contained in the `ProjectorAlgorithm` may include contraction algorithm specific truncation modes which will not translate to a truncated decomposition directly.

This is a good point, but since this was already the case before (the possible values of the `trunc` field of a projector algorithm have not changed) I would leave this for a follow-up. The explanation of which truncation schemes there are and how to select them is a bit scattered anyways right now (especially the extra `η` keyword), so it would be good to sort this all out.

> One alternative could be that we remove the `trunc` field in `SVDAdjoint` / `EighAdjoint` and instead directly instantiate a `TruncatedAlgorithm` as the `fwd_alg` field (which anyway is done later in the call to MatrixAlgebraKit. I'm not sure if that improves things but maybe this is a bit more consistent with MAK.

I could try this out and see how we feel about it.

leburgel:
It turns out that the `degeneracy_atol` was actually not the problem, the issue was that the broadening test for symmetric tensors used a truncation space that cut through degeneracies, in which case the truncated pullback will just not work at all. Thanks to @Jutho for pointing this out. Switching on the verbosity, this shows that the loss function has a very large gauge-dependence even with the old implementation. So in this case all of the gradients were basically nonsense, but we just never noticed.

I updated the choice of degeneracies to be nicely commensurate with the truncation space, which gives sensible results for me locally.

leburgel:
I opened some issues on discussion points that aren't directly related to decomposition handling. I think this is ready for another round of review.

Jutho:
Just a small note that the actual gauge dependence warning, when using svd_trunc_pullback, is about the components of the adjoint variable in the subspaces of the **kept** degenerate eigenvalues, which are subsequently put to zero by the rule, i.e. it is about the part that is being ignored/thrown away.

The components of the cotangents that were mixing the kept versus discarded singular values of a degenerate pair cannot be assessed by `svd_trunc_pullback` because it doesn't know about the discarded singular values and thus about this degeneracy. That part is therefore also not thrown away, and thus causes things blowing up. 

That is different for `svd_pullback` which requires all singular values and will also warn about these gauge components of the adjoints and subsequently throw them away, thereby avoiding divergences.

(Just a clarification as I myself miscommunicated this to Lander earlier today).

src/algorithms/ctmrg/projectors.jl
```diff
@@ -77,8 +73,22 @@ Update the truncation strategy of a given projector algorithm, keeping all other
...
+    _set_decomposition_truncation(alg::ProjectorAlgorithm, trunc::TruncationStrategy)
+
+Set the truncation strategy of the decomposition algorithm within a given projector
+algorithm, keeping all other settings the same.
+"""
+function _set_decomposition_truncation(alg::ProjectorAlgorithm, trunc::TruncationStrategy)
+    decomp_alg = decomposition_algorithm(alg)
+    truncated_fwd_alg = TruncatedAlgorithm(decomp_alg.fwd_alg, trunc)
+    decomp_alg = @set decomp_alg.fwd_alg = truncated_fwd_alg
+    alg = @set alg.decomposition_alg = decomp_alg
```
pbrehmer, line 90:
```suggestion
alg = @set alg.decomposition_alg.fwd_alg = truncated_fwd_alg
```

pbrehmer (approved):
This is looking good to me now if all the tests run through! I only have a minor comment (you can call `@set` on nested structs) but that's just cosmetics really.

src/algorithms/ctmrg/projectors.jl
leburgel: I actually tried this before (when there was still a `trunc` field), but this caused Zygote to throw an error (see [here](https://github.com/QuantumKitHub/PEPSKit.jl/actions/runs/24988487206/job/73167755292)) in the reverse pass. Not sure why, but avoiding `@set`ing a nested field solved the issue (see [here](https://github.com/QuantumKitHub/PEPSKit.jl/pull/364/changes/494b858a3c309a3bfce8be0538d91820926034e9)).

pbrehmer: Man... Zygote is such a buzzkill. Then let's just leave it as is.

lkdvos:
Will merge this as-is and deal with the actions later, to keep things moving.

PR #365, opened 2026-04-22
Update `info` in `leading_boundary` output for CTMRG

leburgel:
Addresses #354. Changes the contents of the `info` named tuple returned by the CTMRG `leading_boundary` to now include:
- `info.converged::Bool`: Boolean flag indicating whether or not the `leading_boundary` run converged to the requested `tol` within the specified `maxiter`.
- `info.convergence_error::Real`: Convergence error after the final iteration of the `leading_boundary` run.

In addition, I would suggest we remove the intermediate objects used in the contraction algorithm (`U`/`S`/`V` for asymmetric CTMRG, `V`/`D` for eigh-C4vCTMRG, `Q`/`R` for QR-C4vCTMRG from the output `info`. This is all extremely specific to the algorithm being used, and as far as I can tell this output is never actually used. I left in the `contraction_metrics` field, since these metrics do seem useful.

The only place where this output is used is in the `info` returned by a single `ctmrg_iteration` (specifically for gauge fixing). That makes much more sense to me, since these decompositions are really intermediate objects in a single iteration. So it makes sense to keep them there. Let me know what you think about removing the decompositions from the `info` output of `leading_boundary` @pbrehmer.

### DIFF
--- a/src/algorithms/ctmrg/ctmrg.jl
+++ b/src/algorithms/ctmrg/ctmrg.jl
@@ -88,10 +88,11 @@ supplied via the keyword arguments or directly as an [`CTMRGAlgorithm`](@ref) st
 
 ## Return values
 
-The `leading_boundary` routine returns the final environment as well as an information `NamedTuple`
-that generally contains a `contraction_metrics` `NamedTuple` storing different contents depending
-on the chosen `alg`. Depending on the contraction method, the information tuple may also contain
-the final tensor decomposition (used in the projectors) including its truncation indices.
+* `env` : The final environment.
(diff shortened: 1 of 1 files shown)


Yue-Zhengyuan (approved):

leburgel:
What do you think about removing the decompositions from the info output of `leading_boundary` @pbrehmer?

pbrehmer:
Sorry, I somehow missed this review request. I think removing the decompositions from the `info` output makes a lot of sense. Especially in light of #364, there is no real reason to keep returning the decompositions. I also like the updated docstring, so this is good to go for me.

pbrehmer (approved):

PR #366, opened 2026-04-23
Improve efficiency of bond truncation

Yue-Zhengyuan:
This PR improves the efficiency of `bond_truncate` (for both `ALSTruncation` and `FullEnvTruncation`) by reusing intermediate partial contraction results.

The test `test/bondenv/bond_truncate.jl` is rewritten to simulate the more realistic scenario of applying a 2-site gate on an iPEPS with physical dimension `d` and virtual dimension `D`. It can reveal that FET is much slower when `d` is large (due to physical leg not transferred to the reduced bond tensors to be optimized).

### DIFF
--- a/src/algorithms/contractions/bondenv/als_solve.jl
+++ b/src/algorithms/contractions/bondenv/als_solve.jl
@@ -3,6 +3,25 @@ In the following, the names `Ra`, `Sa` etc comes from
 the fast full update article Physical Review B 92, 035142 (2015)
 =#
 """
+Contract the virtual legs between
+```
+    -- DX --a-- D --b-- DY --
+            ↓       ↓
+            da      db
+```
+"""
+function _combine_ket(a::MPSTensor, b::AbstractTensorMap{T, S, 1, 2}) where {T, S}
+    return @tensor ket[DX DY; da db] := a[DX da; D] * b[D; db DY]
+end
+function _combine_ket(a::MPSTensor, b::MPSTensor)
+    return @tensor ket[DX DY; da db] := a[DX da; D] * b[D db; DY]
+end
+
+function _combine_ket_for_svd(a::MPSTensor, b::MPSTensor)
+    return @tensor ket[DX da; db DY] := a[DX da; D] * b[D db; DY]
+end
+
+"""
 Construct the norm with bra bond tensors removed
 ```
     ┌benv-------┐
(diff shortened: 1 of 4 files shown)


Yue-Zhengyuan:
@sanderdemeyer I recommend you (squash-)merge this PR to the NTU branch (#144) when trying it.

src/algorithms/contractions/bondenv/als_solve.jl
```diff
@@ -2,139 +2,151 @@
...
-    return @autoopt @tensor benv[DX1 DY1; DX0 DY0] *
-        conj(a1b1[DX1 DY1; da db]) * a2b2[DX0 DY0; da db]
+    return @tensor Sa[DX1 da; D1] :=
+        benv_ket[DX1 DY1; da db] * conj(xs[2][D1 db; DY1])
+end
+function _als_tensor_S(
+        benv_ket::AbstractTensorMap{T, S, 2, 2},
+        xs::Vector{<:MPSTensor}, ::Val{2}
+    ) where {T <: Number, S <: ElementarySpace}
+    return @tensor contractcheck = true Sb[D1 db; DY1] :=
```
leburgel: Why is the `contractcheck` enabled here? Is this from debugging?

src/algorithms/truncation/bond_truncation.jl
```diff
@@ -18,8 +18,8 @@ The truncation algorithm can be constructed from the following keyword arguments
 * `tol::Float64=1e-9` : ALS converges when the relative change in bond SVD spectrum between two iterations is smaller than `tol`.
 * `check_interval::Int=0` : Set number of iterations to print information. Output is suppressed when `check_interval <= 0`. 
 """
-@kwdef struct ALSTruncation
-    trunc::TruncationStrategy
+@kwdef struct ALSTruncation{T <: TruncationStrategy}
```
leburgel, line 21: What was the motivation for promoting the truncation strategy to a type parameter?

src/algorithms/truncation/fullenv_truncation.jl
```diff
@@ -23,8 +23,8 @@ The truncation algorithm can be constructed from the following keyword arguments
 
 * [Glen Evenbly, Phys. Rev. B 98, 085155 (2018)](@cite evenbly_gauge_2018). 
 """
-@kwdef struct FullEnvTruncation
-    trunc::TruncationStrategy
+@kwdef struct FullEnvTruncation{T <: TruncationStrategy}
```
leburgel, line 26: Same question here.

leburgel (approved):
Just a few questions, but otherwise looks good to me.

Test failures and timeouts are unrelated, and will be solved by #364.

src/algorithms/contractions/bondenv/als_solve.jl
Yue-Zhengyuan: Yes, this slipped through the debugging process. I'll remove it later.

src/algorithms/truncation/bond_truncation.jl
Yue-Zhengyuan: I feel that this may be better for type stability (similar to what #361 does to `SimpleUpdate`).

Yue-Zhengyuan:
I dropped the `::Val{i}` dispatch on `_als_tensor_R`, `_als_tensor_S` so that I can reuse the underlying contraction without having to pass the entire `xs` vector all the time.

PR #370, opened 2026-04-30
Remove `:diffgauge` differentiation mode

### LINKED ISSUE #367: Remove `:diffgauge` mode for fixed-point differentiation
Running fixed-point differentiation using `iterscheme = :diffgauge` is very slow, while it no longer has any real benefits over `iterscheme = :fixed`. Therefore, we should probably just remove it (came up in #364).

pbrehmer:
Remove all uses of the `:diffgauge` differentiation mode which differentiates the gauge-fixing routine and is generally slower and more difficult to work with. Closes #367.

As a follow-up, we should at some point thing about making the current gauge-fixing scheme compatible with real tensors or maybe switch the gauge-fixing scheme to just fix the decompositions. (That seems to work in other codes.)


lkdvos (approved):
Looks good, thanks!

It's indeed even more obvious now that selecting `iterscheme = ::Symbol` is a bit strange, since there is only one option.
We could probably get rid of that thing entirely, but for sure in a follow-up PR. 
Thanks for this quick and clean PR, makes it very easy to review :)

PR #371, opened 2026-05-01
CI update

lkdvos:
This updates the repository to use the new organization-shared github actions for setting up the test suite.
Additionally, it ensures we are using the latest `[workspace]` approach for dealing with dependencies.


test/timeevol/cluster_projectors.jl
Yue-Zhengyuan, line 1: Since we are moving the test functions anyway, I want to generalize it to handle MPS with arbitrary physical legs at each site.

Yue-Zhengyuan: Not necessarily in this PR, though...

lkdvos: Yeah, this one should really not modify anything except the structure to make sure it doesn't break anything :)

leburgel (approved):

PR #374, opened 2026-05-04
ensure MPSKit overloads take all keywords

lkdvos:
Ran into some problems with the latest MPSKit version (current main branch) and the boundary MPS code which was caused by the addition of a `prepare::Bool` keyword argument, which then bypassed our specializations.
This is a backwards and forwards compatible fix that resolves this.

Tested this locally against the latest MPSKit version, but wanted to resolve this before releasing another MPSKit version, hence the fix here.

Thanks to @leburgel for reporting!


leburgel (approved):
Thanks for sorting this out!

PR #377, opened 2026-05-07
Periodic indexing

lkdvos:
Not entirely sure why I didn't come up with this sooner, but this is a *light* version of the whole `PeriodicArray` machinery that just hides all of the `mod` stuff behind getters and setters.

I do still want to eventually allow more general array types with tilings etc, but this is non-breaking and at least slightly improves some of the more messy patches of code


Yue-Zhengyuan:
I spotted places that have not been updated to use periodic indexing (mostly related to SU and BP). I'll push some changes on this.

Yue-Zhengyuan (approved):
Good to go for me!

PR #384, opened 2026-05-17
Add a CTMRG benchmark suite

lkdvos:
This PR adds the scaffolding for benchmark suites in PEPSKit, in a similar approach as is already in MPSKit and TensorKit.
I'm definitely open to organizatorial changes, it's not entirely clear to me what the best way forwards is, but at least now the results can be reproduced.


pbrehmer (approved):
This looks good to me! I don't really have anything to criticize for now (we might adjust details as we add modules in the future), so this would be ready to go for me.

Issue #380, opened 2026-05-11
SU(3) PEPS and C4v symmetry


pbrehmer:
This might be a question rather than an issue, but let's say I have an `InfinitePEPS` with $SU(3)$ spaces on the physical and virtual legs and I want to symmetrize it to make it $C_{4v}$ invariant:

```julia
using TensorKit, SUNRepresentations, PEPSKit

SU3Space = Vect[SU3Irrep]
physical_space = SU3Space("3" => 1)
virtual_space = SU3Space("3" => 1, "3⁺" => 1, "1" => 1)

peps = InfinitePEPS(physical_space, virtual_space)
symmetrize!(peps, RotateReflect())
```

It wil throw an error because spaces cannot be fit during symmetrization:

```
LoadError: SpaceMismatch: domain and codomain are not isomorphic: Rep[SU₃]((1, 0, 0) => 1) ← Rep[SU₃]((1, 0, 0) => 1)'
Stacktrace:
  [1] isomorphism!(t::TensorMap{Float64, GradedSpace{SU3Irrep, TensorKit.SortedVectorDict{SU3Irrep, Int64}}, 1, 1, Vector{Float64}})
    @ TensorKit ~/.julia/packages/TensorKit/nhfGA/src/tensors/linalg.jl:103
  [2] unitary!
    @ ~/.julia/packages/TensorKit/nhfGA/src/tensors/linalg.jl:131 [inlined]
  [3] unitary
    @ ~/.julia/packages/TensorKit/nhfGA/src/tensors/linalg.jl:171 [inlined]
  [4] unitary(V::TensorMapSpace{GradedSpace{SU3Irrep, TensorKit.SortedVectorDict{SU3Irrep, Int64}}, 1, 1})
    @ TensorKit ~/.julia/packages/TensorKit/nhfGA/src/tensors/linalg.jl:164
  [5] _fit_spaces(y::TensorKit.AdjointTensorMap{ComplexF64, GradedSpace{SU3Irrep, TensorKit.SortedVectorDict{SU3Irrep, Int64}}, 1, 4, TensorMap{ComplexF64, GradedSpace{SU3Irrep, TensorKit.SortedVectorDict{SU3Irrep, Int64}}, 4, 1, Vector{ComplexF64}}}, x::TensorMap{ComplexF64, GradedSpace{SU3Irrep, TensorKit.SortedVectorDict{SU3Irrep, Int64}}, 1, 4, Vector{ComplexF64}})
[...]

PR #385, opened 2026-05-20
Make C4v rotation and reflection symmetry checks into warnings

pbrehmer:
Here we swap out `ArgumentError`s for `@warn`s in the input checks for `C4vCTMRG` in case the PEPS is not rotation or Hermitian reflection symmetric. This may be relevant in cases such as #380, where the previous checks were too restrictive.


lkdvos (approved):
Looks good to me, although I guess the checks will still error for that particular case right?

pbrehmer:
> Looks good to me, although I guess the checks will still error for that particular case right?

In the case of #380, there will still be errors because there I explicitly call the `symmetrize!` function. I guess there the trick would be to implement a custom symmetrization scheme that won't lead to an error (I guess by flipping the physical space or something).

pbrehmer:
Well nevermind, indeed it is still erroring. I forgot that `_isapprox_localsandwich` of course adds the original and symmetrized PEPS tensor and this will make it error due to differing spaces. I guess then I should rather catch that error first before raising a warning.

PR #386, opened 2026-05-21
Rework fixed-point differentiation

leburgel:
Attempt at reorganizing the fixed-point differentiation approach. The main motivation was to remove the `iterscheme` keyword that was previously used to select a specific flavor of fixed-point differentiation, which has now become trivial since we only support one value.

As a result it now made more sense to switch the logic around: instead of using a solver algorithm struct as a representation of a gradient algorithm where the type parameter specifies how the gradient is actually computed, I now represent the gradient algorithm as a specific type and store the solver algorithm inside that. This involved coming up with a new name, which ended up being `FixedPointGradient`, and wiring this into the algorithm selection. It looks a bit silly now since there's only one option, but hopefully this is a good setup for adding different styles of implicit differentiation.

Very open to suggestions and changes, let me know what you think.


src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -18,183 +21,151 @@ function GradMode(;
...
 
-GRADIENT_MODE_SYMBOLS[:GeomSum] = GeomSum
+_default_solver_alg(::Type{<:FixedPointGradient}) = Defaults.gradient_fixedpoint_solver_alg
+_select_solver_alg_symbol(::Type{<:FixedPointGradient}, solver_alg) =
+    FIXEDPOINT_SOLVER_SYMBOLS[solver_alg]
+function _pad_solver_kwargs(::Type{<:Arnoldi}, solver_kwargs)
+    solver_kwargs = (;
+        eager = Defaults.gradient_fixedpoint_solver_eager,
+        solver_kwargs...,
+    )
```
lkdvos, line 107: I see that this was already there before, but now that my eye catches this: does this mean we get an error when `solver_kwargs` contains `eager`?

```suggestion
    solver_kwargs = merge((; eager = Defaults.gradient_fixedpoint_solver_eager), solver_kwargs)
```

src/Defaults.jl
```diff
@@ -72,20 +72,19 @@ Module containing default algorithm parameter values and arguments.
     - `:C4vEighProjector` : Projection via truncated Eigh of an enlarged corner.
     - `:C4vQRProjector` : Projection via QR decomposition of a column-enlarged corner.
 
-## Fixed-point gradient
+## Contraction
```
lkdvos: I'm not sure I follow why this is called "Contraction"

lkdvos (commented):
Left some very small comments, but this looks pretty good to me actually

src/Defaults.jl
leburgel: Definitely an accident, I have no recollection of changing this.

src/algorithms/optimization/fixed_point_differentiation.jl
leburgel: I think if `solver_kwargs` contains `eager` the default is overruled by the specified value, which is what we usually want. At least this is the pattern we use to do that I think. Probably this just has the same result as using `merge`?

src/algorithms/select_algorithm.jl
```diff
@@ -34,10 +34,6 @@ function select_algorithm(
 
     # C4vCTMRG-specific defaults
     if boundary_alg isa C4vCTMRG
-        # use :LinSolver GradMode since :EigSolver tends to have hiccups
-        if gradient_alg isa NamedTuple
-            haskey(gradient_alg, :alg) || (gradient_alg = merge((; alg = :LinSolver), gradient_alg))
-        end
```
leburgel, line 40: I removed this since I didn't really notice any issues with using an `Arnoldi` solver by default. Do you remember where this was giving trouble, and how I would check this @pbrehmer? If it's still an issue I'll put it back.

examples/3d_ising_partition_function/main.jl
```diff
@@ -153,7 +153,6 @@ of this cost function.
 boundary_alg = SimultaneousCTMRG(; maxiter = 150, tol = 1.0e-8, verbosity = 1)
 rrule_alg = EigSolver(;
     solver_alg = KrylovKit.Arnoldi(; maxiter = 30, tol = 1.0e-6, eager = true),
-    iterscheme = :fixed,
 )
```
pbrehmer: That can probably be fit into one line.

```suggestion
rrule_alg = EigSolver(; solver_alg = KrylovKit.Arnoldi(; maxiter = 30, tol = 1.0e-6, eager = true))
```

examples/bose_hubbard/main.jl
```diff
@@ -88,7 +88,7 @@ algorithms and their tolerances:
 """
 
 boundary_alg = (; tol = 1.0e-8, alg = :SimultaneousCTMRG, trunc = (; alg = :FixedSpaceTruncation))
-gradient_alg = (; tol = 1.0e-6, maxiter = 10, alg = :LinSolver)
+gradient_alg = (; tol = 1.0e-6, maxiter = 10, solver_alg = (alg = :GMRES))
```
pbrehmer: With all the example changes we have to make sure to rerun all of them such that the rendered version reflect the changes.

src/algorithms/optimization/fixed_point_differentiation.jl
lkdvos: Oh you are right, my bad!

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -275,33 +244,47 @@ function _rrule(
...
 
-@doc """
-    fpgrad(∂F∂x, ∂f∂x, ∂f∂A, y0, alg)
+@doc raw"""
+    fixedpoint_gradient(x̆, ∂ₓf, ∂ₚf, y₀, alg)
 
-Compute the gradient of the CTMRG fixed point by solving the following equation:
+Evaluates the VJP action  ``x̆ ∂ₚx`` for an intermediate variable ``x \equiv x(p)``
+characterized which satisfies the fixed-point equation ``x = f(x, p)``, given the
+VJP actions ``∂ₓf`` and ``∂ₚf`` of the iterating function ``f``.
```
pbrehmer: I think the "characterized" should be removed here? But maybe I don't understand the sentence. Thanks for expanding the docs though, this is really helpful here!

src/algorithms/optimization/fixed_point_differentiation.jl
```diff
@@ -1,12 +1,15 @@
-abstract type GradMode{F} end
+abstract type GradMode{A} end
```
pbrehmer: Although this is not really necessary, what do we think about renaming this to `GradientMode{A}`? I think the abbreviation is not really useful here and `GradientMode` sounds better in my opinion. (And we're already renaming and breaking a lot anyway.)

src/algorithms/select_algorithm.jl
pbrehmer: I think there were some complicated optimizations where in the first few optimization steps `Arnoldi` would not converge and then fall back to `GMRES`. But I can't quite remember what it was. I'm fine with using `Arnoldi` by default if we keep in mind that `GMRES` might be more stable in some cases.

test/examples/j1j2_model.jl
```diff
@@ -19,7 +19,8 @@ env₀, = leading_boundary(CTMRGEnv(peps₀, ComplexSpace(χenv)), peps₀)
 # find fixedpoint
 peps, env, E, = fixedpoint(
     H, peps₀, env₀;
-    tol = 1.0e-3, gradient_alg = (; iterscheme = :fixed), symmetrization = RotateReflect(),
+    tol = 1.0e-3,
+    symmetrization = RotateReflect(),
 )
```
pbrehmer: This is probably a one-liner.

```suggestion
peps, env, E, = fixedpoint(H, peps₀, env₀; tol = 1.0e-3, symmetrization = RotateReflect())
```

pbrehmer (commented):
This is really great, thanks! I find that it definitely cleans up and improves the entire (fixed-point) gradient framework. I only left a few minor comments.

src/algorithms/optimization/fixed_point_differentiation.jl
leburgel: I'm happy to rename it. What do you think about `GradientAlgorithm` as an alternative option? I was never too sure about `Mode` to be honest.

examples/3d_ising_partition_function/main.jl
leburgel: Also has a rogue `EigSolver` still, I'll fix that too.

src/algorithms/optimization/fixed_point_differentiation.jl
leburgel: Yes, should indeed be removed.

src/algorithms/optimization/fixed_point_differentiation.jl
pbrehmer: I like `GradientAlgorithm` even better since it is more consistent with the rest of our algorithm struct names.

src/algorithms/select_algorithm.jl
leburgel: I don't see any systematic instabilities, so I would leave the default as is for now. Sometimes the first iterations has to use the GMRES fallback, but this rarely happens more than a handful of times so I don't think this is a systematic issue.

src/algorithms/optimization/fixed_point_differentiation.jl
leburgel: Then I would go for that.

lkdvos (approved):

pbrehmer (approved):

PR #387, opened 2026-05-24
Do not sort `sites` for all LocalCircuit gates

Yue-Zhengyuan:
I feel that there is no reason to sort the sites to be acted on by a gate in `LocalCircuit`. ~Also, with periodic indexing (#377), there is also no need to move the `AbstractTensorMap` gates inside the unit cell, to be more consistent with how MPO gates are handled.~

### DIFF
--- a/src/operators/localcircuit.jl
+++ b/src/operators/localcircuit.jl
@@ -46,6 +46,7 @@ function add_factor!(operator::LocalCircuit, inds::Vector{CartesianIndex{2}}, te
         physicalspace(operator, ind_translated) == domain(term)[i] == codomain(term)[i] ||
             throw(SpaceMismatch("Incompatible physical spaces at $(ind)."))
     end
+    _shift_into_unitcell!(inds, size(operator))
     push!(operator.gates, inds => term)
     return operator
 end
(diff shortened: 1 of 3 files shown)


lkdvos:
I think I understand the benefit of not enforcing that they are sorted to make them consistent with MPO, but I feel like the shift is actually quite a nice feature. It's also slightly more consistent with the LocalOperator, where we do actually want to shift to prevent having duplicate terms that are not merged. How would you feel about also applying the shift to the MPO, keeping the shift for the `AbstractTensorMap`, and getting rid of the sorting? Alternatively, how would you feel about having this all as default behavior, with a way to opt-out?

Yue-Zhengyuan:
Since LocalOperator is a sum of local terms, it is natural to merge operators acting on the same sites there. But for LocalCircuit, even for two gates acting on the same sites, one may still want to apply them individually, instead of their product at once. So shifting the sites is not very useful here, and I prefer not to provide this option at all.

Yue-Zhengyuan:
On second thought... keeping the shift can at least help recognizing gates on the same sites. I'll restore the shift.

Yue-Zhengyuan:
I'll not add the option to opt out.

lkdvos (approved):

PR #388, opened 2026-05-25
Reduce C4v breaking in time evolution (again)

Yue-Zhengyuan:
This PR improves how NNN terms in the Hamiltonian are trotterized to reduce artificial C4v symmetry breaking in time evolution. Depends on #387.

- The sites in an NNN 3-site gate MPO are now always put in counter clockwise order. This is equivalent to the old approach (#219) of rotating the PEPS/PEPO so that the gate can always act on the (say) southeast 3-site corner.
- The order of the NNN 3-site gate MPO is changed: first act on the same corner of all plaquettes in the unitcell, and then change to gates on other corners.
- In `gate_to_mpo`, the resulting MPO is put in Vidal gauge, and the norm of the gate is evenly distributed among the MPO tensors.

@GlebFedorovich You can try this out on J1-J2 (together with `symmetrize_gates = true`).

### DIFF
--- a/src/algorithms/time_evolution/trotter_gate.jl
+++ b/src/algorithms/time_evolution/trotter_gate.jl
@@ -126,7 +126,7 @@ function _trotterize_nnn2site!(gates::Vector, H::LocalOperator, dt::Number)
         (origin, CartesianIndex(1, 0), CartesianIndex(1, 1))
     ]
     Nr = size(H, 1)
-    for x in CartesianIndices(size(H)), (dir, v) in enumerate(vs)
+    for (dir, v) in enumerate(vs), x in CartesianIndices(size(H))
         x′ = if dir == NORTHEAST || dir == SOUTHWEST
             x
         else
--- a/src/algorithms/time_evolution/trotter_gate.jl
+++ b/src/algorithms/time_evolution/trotter_gate.jl
@@ -19,14 +19,14 @@ function gate_to_mpo(
     Os = map(MPSKit.decompose_localmpo(MPSKit.add_util_leg(gate), trunc)) do O
         return permute(O, ((1, 2, 3), (4,)))
     end
+    # convert to Vidal gauge
+    _cluster_truncate!(Os, fill(notrunc(), N - 1))
     # evenly distribute the (Inf) norm
     nrms = norm.(Os, Inf)
     fac = prod(nrms)^(1 / N)
     for (i, nrm) in enumerate(nrms)
         Os[i] *= fac / nrm
     end
-    # convert to Vidal gauge
-    _cluster_truncate!(Os, fill(notrunc(), N - 1))
     # remove trivial legs in first/last tensor, and restore MPO convention
     return map(enumerate(Os)) do (i, O)
         if i == 1
--- a/src/algorithms/time_evolution/trotter_gate.jl
+++ b/src/algorithms/time_evolution/trotter_gate.jl
@@ -15,14 +15,26 @@ function gate_to_mpo(
         trunc = trunctol(; atol = MPSKit.Defaults.tol)
     ) where {N}
     @assert N >= 2
-    Os = MPSKit.decompose_localmpo(MPSKit.add_util_leg(gate), trunc)
-    return map(1:N) do i
+    # use MPS convention of domain/codomain
+    Os = map(MPSKit.decompose_localmpo(MPSKit.add_util_leg(gate), trunc)) do O
+        return permute(O, ((1, 2, 3), (4,)))
+    end
+    # evenly distribute the (Inf) norm
+    nrms = norm.(Os, Inf)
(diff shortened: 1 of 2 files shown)


src/algorithms/time_evolution/trotter_gate.jl
```diff
@@ -75,7 +87,7 @@ function _trotterize_nn2site!(
         gates::Vector, H::LocalOperator, dt::Number; force_mpo::Bool = false
     )
     vs = [CartesianIndex(0, 1), CartesianIndex(1, 0)]
-    for x in CartesianIndices(size(H)), v in vs
+    for v in vs, x in CartesianIndices(size(H))
```
lkdvos: Do you think it would make sense or help to alter the ordering a bit more to use a more trotter-error-friendly approach, where we select an ordering based on what terms commute?
On a square lattice this is just even/odd + horizontal/vertical, giving us 4 classes of commuting terms.
It's not entirely obvious to me which order to use between the different classes, but in keeping with your current change it would be something like:

```julia
for v in vs # horizontal/vertical
    for x in CartesianIndices(size(H))
        iseven(x[horizontal ? 1 : 2]) || continue
        ....
    end
    for x in CartesianIndices(size(H))
        isodd(x[horizontal ? 1 : 2]) || continue
        ....
    end
end
````

File: src/algorithms/time_evolution/trotter_gate.jl [hunk 2]
```diff
@@ -15,14 +15,26 @@ function gate_to_mpo(
         trunc = trunctol(; atol = MPSKit.Defaults.tol)
     ) where {N}
     @assert N >= 2
-    Os = MPSKit.decompose_localmpo(MPSKit.add_util_leg(gate), trunc)
-    return map(1:N) do i
+    # use MPS convention of domain/codomain
+    Os = map(MPSKit.decompose_localmpo(MPSKit.add_util_leg(gate), trunc)) do O
+        return permute(O, ((1, 2, 3), (4,)))
+    end
+    # evenly distribute the (Inf) norm
```
lkdvos, line 24: I would have expected that it makes more sense to normalize **after** going to the Vidal gauge?
Especially if you are using the `Inf` norm this seems more principled, since I don't think that one is invariant under unitary gauge transformations (which is also why I would simply use the 2-norm)

src/algorithms/time_evolution/trotter_gate.jl
Yue-Zhengyuan: > select an ordering based on what terms commute

I feel this can be an even better approach. By "even/odd" I think you are referring to the A (x + y = even) and B (x + y = odd) sublattices?

lkdvos: More or less, that boils down to the same thing indeed. I said even/odd to mean horizontal terms that start on even x or odd x, but you can stagger this for different rows too, that shouldn't matter

src/algorithms/time_evolution/trotter_gate.jl
Yue-Zhengyuan: Since the tensors in the middle of the gate MPO has larger size than the two at the beginning/end, I choose Inf norm to make the elements in all tensors have more or less the same magnitude. I'll normalize after going Vidal.

src/algorithms/time_evolution/trotter_gate.jl
Yue-Zhengyuan: I added some additional logic to handle odd unit cell sizes, for which the bonds are divided into 3 groups in that direction.

lkdvos:
Could you try and rebase/merge the main so it's slightly easier to compare? Otherwise looks good to me!

src/algorithms/time_evolution/trotter_gate.jl
```diff
@@ -15,14 +15,26 @@ function gate_to_mpo(
...
+        return permute(O, ((1, 2, 3), (4,)))
+    end
+    # convert to Vidal gauge
+    _cluster_truncate!(Os, fill(notrunc(), N - 1))
+    # evenly distribute the (Inf) norm
+    nrms = norm.(Os, Inf)
+    fac = prod(nrms)^(1 / N)
+    for (i, nrm) in enumerate(nrms)
+        Os[i] *= fac / nrm
+    end
```
lkdvos:
```suggestion
    scale!.(Os, fac ./ nrms)
```
One less allocation 💪 (not that this would matter though)

lkdvos (approved):
Left a small final comment, otherwise good to go for me!

Yue-Zhengyuan:
Well, hold on from merging... I want to further test order on NNN gate to improve C4v symmetry in NTU (there the two NNN gate acting on the same plaquette along two different diagonal directions still commute, so I have a larger freedom to change the order)

Yue-Zhengyuan:
Anyway, merge first and improve later.

PR #390, opened 2026-06-01
Add some basic support for `svd_trunc_no_error`

kshyatt:
My goal here is to start allowing users to set in the algorithm whether or not they want the truncation error computed. The benefit of not computing it is increased speed and possibly some AD benefits. For now I've done the easiest part -- hooking up `svd_trunc_no_error`, adding `ChainRules.jl` pullback support, and adding tests. Next is to figure out a nice way to control when to use it in the actual CTMRG functions themselves.

### DIFF
--- a/src/utility/svd.jl
+++ b/src/utility/svd.jl
@@ -133,6 +133,24 @@ function MatrixAlgebraKit.svd_trunc!(t::AdjointTensorMap, alg::SVDAdjoint)
     return adjoint(vt), adjoint(s), adjoint(u), ϵ
 end
 
+"""
+    svd_trunc_no_error(t, alg::SVDAdjoint)
+    svd_trunc_no_error!(t, alg::SVDAdjoint)
+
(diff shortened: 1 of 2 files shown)


pbrehmer (approved):
Looks good to me, thanks! I'm curious though, is there a specific use case that we had in mind where we need this SVD without computing the truncation error? I guess you already mentioned increased performance so that's already a good reason :-)

kshyatt:
It might be nice to not have to calculate it every time for the GPUified PEPS as computing the error is actually surprisingly expensive.

pbrehmer (approved):

kshyatt:
Are the timeouts expected? Should I just bypass-merge this?

Yue-Zhengyuan:
The weird timeouts for (min, macos) have been there for a while (it even cannot go past the pre-compilation step). Should be irrelevant to the PR.

kshyatt:
Thanks, @Yue-Zhengyuan . I'm always leery of force-merging stuff into a repo I'm relatively new to so thank you for the confirmation!

PR #397, opened 2026-06-29
Add keyword arguments to `fixedpoint` for `hasconverged` and `shouldstop` stopping criteria

pbrehmer:
This PR adds the possibility for customized converging and stopping criteria through OptimKit's `hasconverged` and `shouldstop` keyword arguments. This can be very usefu e.g.l if one would like to do early stopping of an optimization based on converging energies or if the optimization gets stuck in linesearching and thereby produces vanishing energy differences.

### DIFF
--- a/src/algorithms/optimization/peps_optimization.jl
+++ b/src/algorithms/optimization/peps_optimization.jl
@@ -166,19 +166,10 @@ information `NamedTuple` which contains the following entries:
 * `gradnorms_unitcell` : History of gradient norms for each respective unit cell entry.
 * `times` : History of optimization step execution times.
 """
-function fixedpoint(
-        operator, peps₀::InfinitePEPS, env₀;
-        (finalize!) = OptimKit._finalize!,
-        hasconverged = nothing, shouldstop = nothing, kwargs...,
-    )
-    # select OptimizationAlgorithm from kwargs
-    alg = select_algorithm(fixedpoint, env₀; kwargs...)
-
-    # default stopping criteria initialized on alg parameters that have to be select first
-    isnothing(hasconverged) && (hasconverged = OptimKit.DefaultHasConverged(alg.optimizer_alg.gradtol))
-    isnothing(shouldstop) && (shouldstop = OptimKit.DefaultShouldStop(alg.optimizer_alg.maxiter))
-
-    return fixedpoint(operator, peps₀, env₀, alg; finalize!, hasconverged, shouldstop)
+function fixedpoint(operator, peps₀::InfinitePEPS, env₀; kwargs...)
+    extra_kwarg_keys = (:finalize!, :hasconverged, :shouldstop) # these will not be passed to `select_algorithm`, only to the 2nd `fixedpoint` call
+    alg = select_algorithm(fixedpoint, env₀; filter(kw -> !(first(kw) in extra_kwarg_keys), kwargs)...)
+    return fixedpoint(operator, peps₀, env₀, alg; filter(kw -> first(kw) in extra_kwarg_keys, kwargs)...)
 end
 function fixedpoint(
         operator, peps₀::InfinitePEPS, env₀, alg::PEPSOptimize;
--- a/src/algorithms/optimization/peps_optimization.jl
+++ b/src/algorithms/optimization/peps_optimization.jl
@@ -109,7 +109,9 @@ The optimization parameters can be supplied via the keyword arguments or directl
     4. Debug info including AD outputs
 * `reuse_env::Bool=$(Defaults.reuse_env)` : If `true`, the current optimization step is initialized on the previous environment, otherwise a random environment is used.
 * `symmetrization::Union{Nothing,SymmetrizationStyle}=nothing` : Accepts `nothing` or a `SymmetrizationStyle`, in which case the PEPS and PEPS gradient are symmetrized after each optimization iteration.
-* `(finalize!)=OptimKit._finalize!` : Inserts a `finalize!` function call after each optimization step by utilizing the `finalize!` kwarg of `OptimKit.optimize`. The function maps `(peps, env), f, g = finalize!((peps, env), f, g, numiter)`.
+* `hasconverged=OptimKit.DefaultHasConverged(optimizer_alg.tol)` : Function specifying the convergence criterion with signature `bool = hasconverged(state, cost, grad, gradnorm)`, see [OptimKit.optimize](https://github.com/Jutho/OptimKit.jl/blob/master/src/OptimKit.jl) for specifics.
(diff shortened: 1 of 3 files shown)


test/examples/heisenberg.jl
```diff
@@ -150,3 +150,52 @@ end
     @test e_site2 ≈ E_ref atol = 1.0e-2
     @test all(@. ξ_h > 0 && ξ_v > 0)
 end
+
+@kwdef mutable struct ΔEnergyShouldStop{T <: Real}
+    E_last::T = 0.0
+    tol::T = 1.0e-12
+    maxiter::Int = 100
```
leburgel, line 157: This works more like a `miniter`, no?

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -165,13 +167,24 @@ information `NamedTuple` which contains the following entries:
...
+        (finalize!) = OptimKit._finalize!,
+        hasconverged = nothing, shouldstop = nothing, kwargs...,
     )
+    # select OptimizationAlgorithm from kwargs
     alg = select_algorithm(fixedpoint, env₀; kwargs...)
-    return fixedpoint(operator, peps₀, env₀, alg; finalize!)
+
+    # default stopping criteria initialized on alg parameters that have to be select first
+    isnothing(hasconverged) && (hasconverged = OptimKit.DefaultHasConverged(alg.optimizer_alg.gradtol))
+    isnothing(shouldstop) && (shouldstop = OptimKit.DefaultShouldStop(alg.optimizer_alg.maxiter))
```
leburgel: This is a bit annoying. Can we get around this by not setting any default kwarg values at all here, deleting `(; finalize!, hasconverged, shouldstop)` from the `kwargs` passed to `select_algorithm`, and then do the opposite (only retain `(; finalize!, hasconverged, shouldstop)`) and when returning `fixedpoint(operator, peps₀, env₀, alg; kwargs...)`? Or would this be just as annoying?

src/algorithms/select_algorithm.jl
```diff
@@ -23,7 +23,7 @@ function select_algorithm(
         tol = Defaults.optimizer_tol, # top-level tolerance
         verbosity = 3, # top-level verbosity
         boundary_alg = (;), gradient_alg = (;), optimizer_alg = (;),
-        symmetrization = nothing, kwargs...,
+        symmetrization = nothing, hasconverged = nothing, shouldstop = nothing, kwargs...,
```
leburgel: Why do you have to add this here? I don't think these should ever reach here with the current implementation, no?

pbrehmer: My bad, this was just leftovers from a previous attempt.

test/examples/heisenberg.jl
pbrehmer: I'm not sure what you mean; the stopping condition is that either the energy difference drops below `tol` or that the iteration number is larger or equal to `maxiter`. It can still converge before that with the specified `hasconverged` criterion so there's no `miniter`-like condition here, or am I missing something?

leburgel: Ah, I misread, I thought the or was an and. Then my question is, why enforce this maxiter here, and not just in the optimization algorithm settings? Doesn't matter much, but is there a specific reason for this?

pbrehmer: If a `shouldstop` function is specified in `OptimKit.optimize` this overrides the default criterion which just checks the `maxiter`. So in this case, if we hadn't additionally enforced the `maxiter` in that struct, the optimization would only stop if it converged or if the energy difference became small enough. That's why I added a comment to the `fixedpoint` docstring where it says that specifying `hasconverged` and `shouldstop` overrides the default behavior since that is how it's handled by OptimKit.

leburgel: I see, makes sense then!

src/algorithms/optimization/peps_optimization.jl
pbrehmer: I gave this a go and I think I prefer your suggestion. It's still a bit annoying having to filter the kwargs but I think it's less annoying that way and saves lines.

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -164,14 +166,16 @@ information `NamedTuple` which contains the following entries:
...
 """
-function fixedpoint(
-        operator, peps₀::InfinitePEPS, env₀; (finalize!) = OptimKit._finalize!, kwargs...,
-    )
-    alg = select_algorithm(fixedpoint, env₀; kwargs...)
-    return fixedpoint(operator, peps₀, env₀, alg; finalize!)
+function fixedpoint(operator, peps₀::InfinitePEPS, env₀; kwargs...)
+    extra_kwarg_keys = (:finalize!, :hasconverged, :shouldstop) # these will not be passed to `select_algorithm`, only to the 2nd `fixedpoint` call
+    alg = select_algorithm(fixedpoint, env₀; filter(kw -> !(first(kw) in extra_kwarg_keys), kwargs)...)
+    return fixedpoint(operator, peps₀, env₀, alg; filter(kw -> first(kw) in extra_kwarg_keys, kwargs)...)
```
leburgel, line 172: Not beautiful, but I think it's slightly less hacky than setting the `nothing` defaults first and then replacing them.

Maybe @lkdvos can give an independent opinion?

test/examples/heisenberg.jl
```diff
@@ -150,3 +150,52 @@ end
     @test e_site2 ≈ E_ref atol = 1.0e-2
     @test all(@. ξ_h > 0 && ξ_v > 0)
 end
+
+@kwdef mutable struct ΔEnergyShouldStop{T <: Real}
```
lkdvos: For the example it might make sense to just hard-code `T` to be `Float64`, it might be slightly clearer (and bypasses me being annoying about the initialization needing to be `eps(T)^some_power` to be generic)

lkdvos (commented):
Left a small suggestion for the example, and @leburgel's comment about kwargs seems to not be resolved yet, otherwise good to go for me!

test/examples/heisenberg.jl
pbrehmer: Fine by me, changed it!

Regarding @leburgel's comment: I did implement Lander's suggestion or am I missing something? :-) I guess the question was what your opinion is on the kwarg filtering approach. Personally, I do think it's fine and it doesn't cost us many extra lines.

src/algorithms/optimization/peps_optimization.jl
lkdvos: I don't think I can come up with much better, the only alternative I could think off is to explicitly put the optimization keywords into the `alg`, so we don't have to peel them but I think this is probably fine?

lkdvos (approved):

src/algorithms/optimization/peps_optimization.jl
pbrehmer: I considered that but was ultimately against it because I felt that having `Function`s as fields could become annoying.

PR #399, opened 2026-07-01
Correlator for purified iPEPO and right-to-left contraction

Yue-Zhengyuan:
This PR improves the `correlator` function by adding support for purified iPEPO and removing some constraints on sites `js`, with some under-the-hood deduplications.

- Deduplicate iPEPS and one-layer iPEPO trace correlator logic through private correlator context/adapters.
- Add purified iPEPO correlators:
  ```julia
  correlator(bra::InfinitePEPO, O, i, js, ket::InfinitePEPO, env::CTMRGEnv)
  ```
  matching the purified-state convention used by `expectation_value`.
- Generalize horizontal correlators so `js` may contain targets on either side of `i`.
  - `js` no longer need to be sorted.
  - `js` can contain sites on both sides of `i`.
  - Output order follows `vec(js)`.
  - `j == i` is rejected with `ArgumentError`.
- Vertical correlators inherit the same behavior through dispatching to horizontal correlators via rotation.
- Add right-to-left correlator contractions and the corresponding `edge_transfer_right` paths for `js` on the left of `i`.

## Implementation Notes

The shared correlator implementation now dispatches through context types:

- `PEPSCorrelator`
- `PEPOPurifiedCorrelator`
- `PEPOTraceCorrelator`

The horizontal sweep partitions `js` into right and left sites internally:

- right targets use the existing left-to-right contraction path;
- left targets use a mirrored right-to-left contraction path;
- results are written back into the original target order.

Low-level tensor contractions remain separate where the tensor ranks genuinely differ.

### DIFF
--- a/src/algorithms/contractions/correlator/pepo_1layer.jl
+++ b/src/algorithms/contractions/correlator/pepo_1layer.jl
@@ -1,4 +1,4 @@
-function start_correlator_left(
+function start_correlator(
         i::CartesianIndex{2}, ρ::InfinitePEPO,
         O::PFTensor, env::CTMRGEnv
     )
@@ -25,7 +25,7 @@ function start_correlator_left(
     return Vn, Vo
 end
 
-function end_correlator_right_numerator(
+function end_correlator_numerator(
         j::CartesianIndex{2}, V::CTMRGEdgeTensor{T, S, 3},
         ρ::InfinitePEPO, O::PFTensor, env::CTMRGEnv
     ) where {T, S}
@@ -44,7 +44,7 @@ function end_correlator_right_numerator(
         t[d1 d2; DN DE DS DW] * removeunit(O, 4)[dstring d2; d1]
 end
 
-function end_correlator_right_denominator(
+function end_correlator_denominator(
         j::CartesianIndex{2}, V::CTMRGEdgeTensor{T, S, 2}, env::CTMRGEnv
     ) where {T, S}
     r, c = Tuple(j)
--- a/src/algorithms/contractions/correlator/pepo_1layer.jl
+++ b/src/algorithms/contractions/correlator/pepo_1layer.jl
@@ -28,7 +28,7 @@ function start_correlator_left(
 end
 
 function end_correlator_right_numerator(
-        j::CartesianIndex{2}, V::CTMRGEdgeTensor{T, S, 3},
+        j::CartesianIndex{2}, V::AbstractTensorMap{T, S, 3, 1},
         ρ::InfinitePEPO, O::PFTensor, env::CTMRGEnv
     ) where {T, S}
     (size(ρ, 3) == 1) ||
@@ -47,7 +47,7 @@ function end_correlator_right_numerator(
 end
 
 function end_correlator_right_denominator(
-        j::CartesianIndex{2}, V::CTMRGEdgeTensor{T, S, 2}, env::CTMRGEnv
+        j::CartesianIndex{2}, V::AbstractTensorMap{T, S, 2, 1}, env::CTMRGEnv
     ) where {T, S}
     r, c = Tuple(j)
     C_northeast = corner(env, NORTHEAST, r - 1, c + 1)
@@ -86,7 +86,7 @@ end
 
 
 function end_correlator_left_numerator(
-        j::CartesianIndex{2}, V::CTMRGEdgeTensor{T, S, 3},
+        j::CartesianIndex{2}, V::AbstractTensorMap{T, S, 3, 1},
         ρ::InfinitePEPO, O::PFTensor, env::CTMRGEnv
     ) where {T, S}
     (size(ρ, 3) == 1) ||
@@ -106,7 +106,7 @@ function end_correlator_left_numerator(
 end
 
 function end_correlator_left_denominator(
-        j::CartesianIndex{2}, V::CTMRGEdgeTensor{T, S, 2}, env::CTMRGEnv
+        j::CartesianIndex{2}, V::AbstractTensorMap{T, S, 2, 1}, env::CTMRGEnv
     ) where {T, S}
     r, c = Tuple(j)
     C_northwest = corner(env, NORTHWEST, r - 1, c - 1)
--- a/src/algorithms/contractions/correlator/pepo_1layer.jl
+++ b/src/algorithms/contractions/correlator/pepo_1layer.jl
@@ -1,4 +1,6 @@
-function start_correlator(
+# -------- For left-to-right correlator contraction --------
+
+function start_correlator_left(
         i::CartesianIndex{2}, ρ::InfinitePEPO,
         O::PFTensor, env::CTMRGEnv
     )
@@ -25,7 +27,7 @@ function start_correlator(
     return Vn, Vo
 end
 
-function end_correlator_numerator(
+function end_correlator_right_numerator(
         j::CartesianIndex{2}, V::CTMRGEdgeTensor{T, S, 3},
         ρ::InfinitePEPO, O::PFTensor, env::CTMRGEnv
     ) where {T, S}
@@ -44,7 +46,7 @@ function end_correlator_numerator(
         t[d1 d2; DN DE DS DW] * removeunit(O, 4)[dstring d2; d1]
 end
 
-function end_correlator_denominator(
+function end_correlator_right_denominator(
         j::CartesianIndex{2}, V::CTMRGEdgeTensor{T, S, 2}, env::CTMRGEnv
     ) where {T, S}
     r, c = Tuple(j)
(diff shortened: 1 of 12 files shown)


src/algorithms/correlators.jl
```diff
@@ -11,52 +9,28 @@ function correlator_horizontal(
...
-            Ebot = edge(env, SOUTH, i[1] + 1, i[2])
-            sandwich = (
-                ket[mod1(i[1], end), mod1(i[2], end)], bra[mod1(i[1], end), mod1(i[2], end)],
-            )
-            T = edge_transfermatrix(Etop, sandwich, Ebot)
-            Vo = Vo * T
+    c = i # current column being handled
+    Vn, Vo = start_correlator_left(c, bra, O[1], ket, env)
+    j_last = last(js)
+    return map(enumerate(js)) do (_, j)
```
Yue-Zhengyuan: When I try just writing `map(js) do j`, the test `test/utility/correlator.jl` starts to use lots of memory and becomes slow. Would like some guide from @lkdvos in understanding what's going on...

src/algorithms/contractions/transfer.jl
Yue-Zhengyuan, line 1: It turns out that none of `edge_transfer_right` is used in PEPSKit. Instead of commenting them out, I prefer to just remove them, keeping things minimal.

src/algorithms/correlators.jl
Yue-Zhengyuan: I realized that this may be due to things inside `map` modifying `c`, `Vn`, `Vo` that are defined outside. I've restored the old approach to first pre-allocate the correlator vector.

src/algorithms/contractions/transfer.jl
Yue-Zhengyuan: @leburgel Since you helped introduce `EdgeTransferMatrix` back in #274, can you have a look and see if you actually wanted to use `edge_transfer_right` somewhere else?

leburgel: I agree removing them is the cleanest thing to do, but can you add a corresponding entry in the "Removed" section of the changelog with a link to this PR? That way we can find them in the git history easily, since this will be impossible to spot just from the PR title alone.

leburgel: > @leburgel Since you helped introduce `EdgeTransferMatrix` back in #274, can you have a look and see if you actually wanted to use `edge_transfer_right` somewhere else?

I am using them somewhere in a project that depends on PEPSKit, but I can just pin that to a version that still has them. I don't have a use case within PEPSKit itself yet, so it's indeed best to just remove them for now.

Yue-Zhengyuan: Actually I wanted to implement correlators for `j` on the left of `i`. In this case, if we don't want to rotate the state by 180 degrees, we need to contract from right to left, and use `edge_transfer_right`. This way we can test and save them in the package.

src/algorithms/contractions/correlator/pepo_1layer.jl
```diff
@@ -25,8 +27,8 @@ function start_correlator(
     return Vn, Vo
 end
 
-function end_correlator_numerator(
-        j::CartesianIndex{2}, V::CTMRGEdgeTensor{T, S, 3},
+function end_correlator_right_numerator(
+        j::CartesianIndex{2}, V::AbstractTensorMap{T, S, 3, 1},
```
Yue-Zhengyuan, line 31: Since the `V` tensors (partially contracted correlator network) are not actually CTMRG edge tensors, I prefer just annotate their type using `AbstractTensorMap`.

Yue-Zhengyuan:
A recent idea: in almost all applications, the two-site operator passed to `correlator` is made from two one-site operators, so the MPO has virtual dimension 1 (although it carries a non-trivial charge in general). Then can we first calculate the product of the MPO and the ket state (or the iPEPO), fusing the MPO virtual leg into the state? In this way, we can get rid of the edge_transfer functions carrying the "excited" leg from the MPO. The computational cost should remain almost the same.

src/algorithms/contractions/correlator/pepo_purified.jl
```diff
@@ -0,0 +1,135 @@
+# -------- For left-to-right correlator contraction --------
+
+function start_correlator_left(
+        i::CartesianIndex{2},
+        below::InfinitePEPO,
+        O::MPOTensor,
```
leburgel: Doesn't matter much since they literally mean the same, but it might be more natural to use `O::PFTensor` here, the same as it is in the pepo_1layer case?

src/algorithms/contractions/transfer.jl
```diff
@@ -176,3 +199,51 @@ function edge_transfer_left(
         Ebot[χ_SE D_S; χ_SW] *
         O[D_W D_S; D_N D_E]
 end
+
+"""
+    transfer_right(v, O, Et, Eb)
```
leburgel:
```suggestion
    edge_transfer_right(v, O, Et, Eb)
```
I made a typo in the original docstring, which was probably copied here.

src/algorithms/contractions/transfer.jl
```diff
@@ -176,3 +199,51 @@ function edge_transfer_left(
...
+    
+Apply an edge transfer matrix to the right on an excited vector..
```
leburgel:
```suggestion
Apply an edge transfer matrix to the right on an excited vector.
```

src/networks/local_sandwich.jl
```diff
@@ -108,8 +108,11 @@ end
 
 ## PEPO
 
-const PEPOSandwich{N, T <: PEPSTensor, P <: PEPOTensor} = Tuple{T, T, Vararg{P, N}}
+const PEPOPurifiedSandwich{K <: PEPOTensor, B <: PEPOTensor} = Tuple{K, B}
```
leburgel: Do you ever expect to have different types for the two elements in a purified sandwich? For `PEPSSandwich` we also always assume the bra and ket tensors to have the same type, so probably here we can just use one type parameter?
```suggestion
const PEPOPurifiedSandwich{P <: PEPOTensor} = Tuple{P, P}
```

src/algorithms/correlator_adapters.jl
```diff
@@ -0,0 +1,262 @@
...
+    return _PEPOPurifiedCorrelator(rotated_bra, rotated_ket, rotl90(context.env))
+end
+
+Base.rotl90(context::_PEPOTraceCorrelator) =
+    _PEPOTraceCorrelator(rotl90(context.ρ), rotl90(context.env))
+
+# -------- For left-to-right correlator contraction --------
+
+function _start_correlator_left(
+        i::CartesianIndex{2}, context::_BraketCorrelator, O::MPOTensor
```
leburgel: Same comment as above, is the distinction in using both `MPOTensor` and `PFTensor` in the type annotations intentional?

leburgel (commented):
Looks mostly good. Aside from the remarks below, I have two comments about some (not all that important) details:

- It seems many of the `start/end_correlator_left/right` methods have the exact same initialization in terms of indexing the environment, and potentially adding a twist to the PEPO tensor (in the bra layer). Would it be easy to only do this indexing and preparation once, and only really implement the different contractions for the different signatures? Not very important, but it would be nice if it's not a lot of work. If it is, then never mind.

- I think the solution provided by the `correlator_adapters` is quite nice. The only comment I have is on the underscores in all the type and method names there. This makes perfect sense for helper methods like `_correlator_scalartype`, but looks a bit strange to me in the type names like `_PEPSCorrelator` for example. Especially cases where an underscore method eventually lowers to an method without underscore (e.g.`_edge_transfermatrix` eventually lowering to `edge_transfermatrix`) is a bit strange to me. I think I would always expect this to be the other way around, where the underscore method would be the lowest one. Would it be acceptable to just drop the leading underscores (except of course in things like `_correlator_scalartype`), or would this lead to ambiguities?

src/algorithms/correlator_adapters.jl
Yue-Zhengyuan: No. I will unify all cases with `PFTensor`.

Yue-Zhengyuan:
> many of the start/end_correlator_left/right methods have the exact same initialization

I suggest to leave this to a follow-up PR in the future. May come up with something nice when working on long-range measurements.

> underscores in all the type and method names

I have removed the beginning underscore for the correlator adapter types and their constructors, and directly use `edge_transfermatrix` as well.

leburgel (approved):
I would personally also remove the leading underscore from the `_start/end_correlater_left/right(_numerator/denominator)` methods, since otherwise we still have underscore methods lowering to non-underscore methods which looks a bit weird to me (especially if there's no dispatch issues with not using underscores on the upper level). But this is more a personal preference, I'll leave the decision up to you and already approve.

PR #402, opened 2026-07-17
[WIP] Get PEPSKit working with Enzyme

kshyatt:
Trying to see if I can make this happen. Right now, `SimultaneousCTMRG` doesn't work because we need a reverse rule for `dtmap` (working on that) but other things... do seem to work? Kind of cool! There are some other errors I see related to LLVM I will open issues for as I encounter them. But, if others want to take a look, go for it :)


kshyatt:
As mentioned in a few comments, this is completely blocked on https://github.com/EnzymeAD/Enzyme/pull/3152 which is needed for the `dtmap(!!)` rules to work

PR #407, opened 2026-07-29
[WIP] GPU support in PEPSKit

kshyatt:
cc @dartsushi see the `boundarymps/vumps.jl` for `test/cuda` to get started


dartsushi:
You are awesome! I will try it out tomorrow

src/environments/ctmrg_environments.jl
```diff
@@ -48,20 +48,20 @@ function check_environment_virtualspace(E::CTMRGEdgeTensor)
...
+        f, ::Type{TorA}, left_vspace::S, right_vspace::S = left_vspace
+    ) where {TorA, S <: ElementarySpace}
+    return f(TorA, left_vspace ← right_vspace)
 end
 
 function _edge_tensor(
-        f, ::Type{T}, left_vspace::S, pspaces::P, right_vspace::S = left_vspace
-    ) where {T, S <: ElementarySpace, P <: ProductSpace}
-    return f(T, left_vspace ⊗ pspaces, right_vspace)
+        f, ::Type{TA}, left_vspace::S, pspaces::P, right_vspace::S = left_vspace
```
Yue-Zhengyuan: Should `TA` also be called `TorA` ? Also, can you educate me on why this "storage type" is called `TorA`?

kshyatt: This is still WIP so perhaps we can wait on comments until it's marked ready for review

kshyatt: (This line might be going away with updates to the Adapt extension)

lkdvos: @Yue-Zhengyuan, this is mostly a convention since we tend to use T for scalartype and A for storagetype (this is also the name of the type parameters in the TensorMap struct definition IIRC)

Yue-Zhengyuan: I'm mainly curious about the "Tor"

kshyatt: "T or A"

kshyatt: You can either specify an element type `T` such that the backing array type becomes `Vector{T}`, or an array type `A` like `CuVector{ComplexF64}` if you want more fine control

Yue-Zhengyuan: Ah that's surprisingly simple... I was thinking about some [fancy math](https://en.wikipedia.org/wiki/Tor_functor) named "Tor"

kshyatt: Oh definitely not yet, we're keeping it simple for now

kshyatt:
OK, I think the main blocker here is https://github.com/QuantumKitHub/TensorOperations.jl/issues/298,

kshyatt:
An additional problem is that the `timeevol/j1j2_finiteT` test is very slow. I'll look into why.

kshyatt:
And it would also make sense to set a lot of this up like the GPUArrays testsuite, where we just pass a list of array types to adapt to, to avoid the massive code explosion here.

kshyatt:
OK this diff just got a whole lot less crazy. I didn't add GPU tests for `types` or `examples` but can do so if others prefer.

kshyatt:
I'm making a few more updates to get CI to play nicely on top of the testsuite changes, and this is also waiting on an MPSKit tag. But then I think we're good to go!

kshyatt:
OK, I think the main blocker here is now the MPSKit tag that's needed, as well as the very long ROCm CI queue 😛

kshyatt:
I think this is ready for folks to start having a look @Yue-Zhengyuan @leburgel @pbrehmer <3. Happy to explain anything I can!

PR #409, opened 2026-08-04
Add a Buildkite pipeline file

kshyatt:
Should help with Buildkite instantly failing while the GPU branch develops


kshyatt:
OK this seems to be working for CUDA and the other tests are unbroken, good to merge imo

lkdvos (approved):

lkdvos:
Will force-merge this and cancel CI, don't think running the full suite is necessary here.

PR #410, opened 2026-08-04
Add adaptive tolerance scaling for boundary and gradient algorithms in variational optimization

leburgel:
Adds adaptive tolerance scaling, where the tolerance of the contraction algorithm is set based on the current norm of the gradient (default scaling factor of `1.0e-3`). The tolerance of the gradient algorithm in turn is set based on the contraction tolerance (default scaling factor `1.0e1`).

On the README example, these default settings give a 1.5x speedup in walltime needed to reach a given energy value starting from the same initial guess. I'll see what the tests say for now, and do some more trials in the meantime.

At the same time, this reworks the way the per-iteration info metrics are stored and added. Previously, the `contraction_metrics` and `gradnorms_unitcell` got a new entry every time cost function was called. These function evaluations don't necessarily match the iterations of the optimization loop. In particular, entries were added for every rejected step during the line search, which just seemed needlessly confusing. I changed to workflow to record the current values in the cost function, but only add entries in the `finalize!` routine, which is actually called after every iteration (i.e. every accepted step).

This is also in anticipation of adding preconditioning, which naturally requires access to the current gradient norm.

### DIFF
--- a/src/algorithms/optimization/peps_optimization.jl
+++ b/src/algorithms/optimization/peps_optimization.jl
@@ -155,6 +155,22 @@ keyword arguments are:
 * `ls_maxfg::Int=$(Defaults.ls_maxfg)` : Maximal number of function-gradient evaluations during linesearch.
 * `lbfgs_memory::Int=$(Defaults.lbfgs_memory)` : Size of limited memory representation of BFGS Hessian matrix.
 
+### Dynamic tolerances
+
+The boundary and gradient algorithms each additionally accept the keyword arguments below,
+which wrap the corresponding algorithm in an `MPSKit.DynamicTols.DynamicTol` that
+rescales its tolerance over the course of the optimization. This allows the intermediate
+problems to be solved only as accurately as the current optimization step requires, which
+can significantly reduce the total runtime. The boundary tolerance is scaled relative to the
+current gradient norm, while the gradient tolerance is in turn scaled relative to the
+effective boundary tolerance. These settings are only available within a
+variational optimization, not for standalone [`leading_boundary`](@ref) calls.
+
+* `dynamic_tols::Bool` : Enable dynamic tolerance scaling for this algorithm. Defaults to `$(Defaults.ctmrg_dynamic_tols)` and `$(Defaults.gradient_dynamic_tols)` for the boundary and gradient algorithm respectively.
+* `tol_min::Real` : Lower clamp on the dynamically scaled tolerance.
+* `tol_max::Real` : Upper clamp on the dynamically scaled tolerance.
+* `tol_factor::Real` : Prefactor of the dynamically scaled tolerance.
+
 ## Return values
 
 The function returns the final PEPS, CTMRG environment and cost value, as well as an
--- a/src/algorithms/optimization/peps_optimization.jl
+++ b/src/algorithms/optimization/peps_optimization.jl
@@ -137,6 +137,22 @@ keyword arguments are:
     - `:FixedPointGradient` : Compute the gradient via fixed-point differentiation, see [`FixedPointGradient`](@ref)
 * `solver_alg::Union{Algorithm,NamedTuple}`: Solver algorithm for computing the implicit gradient; see [`FixedPointGradient`](@ref) for supported algorithms.
 
+### Dynamic tolerances
+
+The boundary and gradient algorithms each additionally accept the keyword arguments below,
+which wrap the corresponding algorithm in an `MPSKit.DynamicTols.DynamicTol` that
+rescales its tolerance over the course of the optimization. This allows the intermediate
+problems to be solved only as accurately as the current optimization step requires, which
+can significantly reduce the total runtime. The boundary tolerance is scaled relative to the
+current gradient norm, while the gradient tolerance is in turn scaled relative to the
+effective boundary tolerance. These settings are only available within a
+variational optimization, not for standalone [`leading_boundary`](@ref) calls.
+
+* `dynamic_tols::Bool` : Enable dynamic tolerance scaling for this algorithm. Defaults to `$(Defaults.ctmrg_dynamic_tols)` and `$(Defaults.gradient_dynamic_tols)` for the boundary and gradient algorithm respectively.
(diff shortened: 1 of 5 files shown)


src/algorithms/optimization/peps_optimization.jl
```diff
@@ -235,13 +272,14 @@ function fixedpoint(
...
 
 Check compatibility of an initial PEPS and environment with a specified PEPS optimization algorithm.
 """
-function check_input(::typeof(fixedpoint), peps₀, env₀, alg::PEPSOptimize) end
-function check_input(::typeof(fixedpoint), peps₀, env₀, alg::PEPSOptimize{<:SimultaneousCTMRG, <:FixedPointGradient})
-    if scalartype(env₀) <: Real # :fixed mode gauge fixing is incompatible with real environments
+function check_input(::typeof(fixedpoint), peps₀, env₀, alg::PEPSOptimize)
+    if parent_alg(alg.boundary_alg) isa SimultaneousCTMRG &&
+            parent_alg(alg.gradient_alg) isa FixedPointGradient &&
+            scalartype(env₀) <: Real # :fixed mode gauge fixing is incompatible with real environments
```
pbrehmer: I think the `SimultaneousCTMRG` check could technically be removed, right? `FIxedPointGradient` should work with `SequentialCTMRG` now as well.

```suggestion
    if parent_alg(alg.gradient_alg) isa FixedPointGradient &&
            scalartype(env₀) <: Real # :fixed mode gauge fixing is incompatible with real environments
```

pbrehmer (commented):
Up to one very minor thing, this looks good to me! I really like that this also clears up how the per-iteration information is logged.

pbrehmer (approved):

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -277,8 +277,7 @@ end
 Check compatibility of an initial PEPS and environment with a specified PEPS optimization algorithm.
 """
 function check_input(::typeof(fixedpoint), peps₀, env₀, alg::PEPSOptimize)
-    if parent_alg(alg.boundary_alg) isa SimultaneousCTMRG &&
-            parent_alg(alg.gradient_alg) isa FixedPointGradient &&
+    if parent_alg(alg.gradient_alg) isa FixedPointGradient &&
```
pbrehmer: Okay darn, in my last suggestion I misunderstood this check and in fact we should have done the opposite; it also needs to check for `SequentialCTMRG` since that is also compatible with `FixedPointGradient` and therefore should not be allowed to work with real environments. I'm not sure whether this is the best way to do it but it should be something like:

```suggestion
typeof(parent_alg(alg.boundary_alg)) <: Union{SimultaneousCTMRG, SequentialCTMRG} &&
    parent_alg(alg.gradient_alg) isa FixedPointGradient &&
```

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -205,20 +238,24 @@ function fixedpoint(
...
-                push!(contraction_metrics, info.contraction_metrics)
+                latest_metrics[] = info.contraction_metrics
             end
             return cost_function(ψ, env′, operator)
         end
         g = only(gs)  # `withgradient` returns tuple of gradients `gs`
-        push!(gradnorms_unitcell, norm.(g.A))
-        push!(times, (time_ns() - start_time) * 1.0e-9)
+        latest_gradnorms[] = norm.(unitcell(g))
+        latest_time[] = (time_ns() - start_time) * 1.0e-9
```
lkdvos: Is there a reason to ask for `time_ns` to then immediately convert to `* 1e-9`?

src/algorithms/optimization/peps_optimization.jl
```diff
@@ -338,3 +375,36 @@ function symmetrize_retract_and_finalize!(
...
+Return a `finalize!` function that, after calling `finalize!` (defaulting to
+`OptimKit._finalize!`):
+* updates `tol_state[]` to the `(iter, gradnorm)` of the now-accepted optimization step,
+  used to drive `MPSKit.updatetol` for any `alg.boundary_alg`/`alg.gradient_alg` wrapped
+  in a `MPSKit.DynamicTol`
+* records `latest_metrics[]`/`latest_gradnorms[]`/`latest_time[]`, i.e. the values
+  produced by the `fg` call corresponding to the accepted step, into
+  `contraction_metrics`/`gradnorms_unitcell`/`times`
+"""
+function track_state_and_finalize!(
```
lkdvos: This is more of a suggestion about code organization, which would achieve the exact same thing, but I'm not that used to having many of these `Ref` values around as a way to keep things mutable, and combining this with an anonymous function factory might not be the most readable way of getting to this.

It might make sense to create a simple mutable struct for this, and then overload function calling? Something like
```
mutable struct TrackedFinalizer{A, B, C, D}
    tol_state::A
    latest_metrics::B
    latest_gradnorms::C
    latest_time::Float64 # I'm guessing?
    const contraction_metrics::Vector{B}
    const gradnorms_unitcell::Vector{C}
    const times::Vector{Float64}
    const finalize!
end
```

I'm not saying this is necessarily better, but I might have structured it more like this so just wanted to leave this suggestion to see what you think.

Additionally, it might even be possible to get rid of the mutable ref values and simply have a finalizer that `resize!`s the vector's by `length + 1`, and then just update the last element in the vector, which is of course mutable. (although that might lead to subtleties for the last iteration, maybe?)

src/algorithms/optimization/peps_optimization.jl
leburgel: No idea, I didn't really touch this. Nanosecond resolution for a PEPS optimization iteration doesn't sound like something that will be necessary anytime soon, so I can change it to `time`.

src/algorithms/optimization/peps_optimization.jl
leburgel: I actually like `resize!` a lot better, didn't think of that.

I'm fine with using a callable mutable struct instead, let me give this a go.

lkdvos (approved):

PR #413, opened 2026-08-06
Implicit derivatives for CTMRG

leburgel:
Adds implicit gradients for CTMRG contractions.

Reference: https://arxiv.org/abs/2607.15030

### DIFF
--- a/src/algorithms/contractions/ctmrg/characteristic_equations.jl
+++ b/src/algorithms/contractions/ctmrg/characteristic_equations.jl
@@ -294,20 +294,19 @@ function ChainRulesCore.rrule(
         ΔC = unthunk(ΔC_)
         F = similar(t)
         for (c, b) in blocks(C)
-            copyto!(block(F, c), _fourthroot_pullback(b))
+            copyto!(block(F, c), _fourthroot_pullback(block(ΔC, c), b))
         end
-        return NoTangent(), P_t(_elementwise_mult(ΔC, F))
+        return NoTangent(), P_t(F)
     end
     return C, fourthroot_pullback
 end
-function _fourthroot_pullback(C::AbstractMatrix)
-    Fdata = similar(C)
-    for i in axes(Fdata, 1), j in axes(Fdata, 2)
-        # Taking the diagonal only is okay, when dA is diagonal anyway: Fdata[i, i] = 1 / (4 * conj(C[i, i]^3))
-        # However, for Q-deformed CTMRG, we need the full version:
-        Fdata[i, j] = 1 / conj(C[i, i]^3 + C[i, i] * C[j, j]^2 + C[i, i]^2 * C[j, j] + C[j, j]^3)
-    end
-    return Fdata
+function _fourthroot_pullback(ΔC::AbstractMatrix, C::AbstractMatrix)
+    # Taking the diagonal only is okay, when dA is diagonal anyway: Fdata[i, i] = 1 / (4 * conj(C[i, i]^3))
+    # However, for Q-deformed CTMRG, we need the full version:
+    Cd = diagview(C) # column
+    Cdt = transpose(Cd) # row
+    F = @. ΔC / conj(Cd^3 + Cd * Cdt^2 + Cd^2 * Cdt + Cdt^3)
+    return F
 end
 
 # Util
--- a/src/algorithms/contractions/ctmrg/characteristic_equations.jl
+++ b/src/algorithms/contractions/ctmrg/characteristic_equations.jl
@@ -9,6 +9,12 @@ const EdgeTensor{S, N} = AbstractTensorMap{T, S, N, 1} where {T}
 const LeftProjector{S, N} = AbstractTensorMap{T, S, N, 1} where {T}
 const RightProjector{S, N} = AbstractTensorMap{T, S, 1, N} where {T}
 
+const CornerTensors{TC} = Array{TC, 3} where {TC <: CornerTensor}
+const EdgeTensors{TE} = Array{TE, 3} where {TE <: EdgeTensor}
+const LeftProjectors{TP} = Array{TP, 3} where {TP <: LeftProjector}
+const RightProjectors{TP} = Array{TP, 3} where {TP <: RightProjector}
+
+
 ## C4v symmetric case
 
 # partial contractions for different networks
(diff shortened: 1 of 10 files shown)


src/algorithms/contractions/ctmrg/characteristic_equations.jl
```diff
@@ -0,0 +1,684 @@
...
+        ::CTMRGAlgorithm{<:HalfInfiniteProjector},
+        iSfp::CornerTensors,
+        Ufp::LeftProjectors,
+        Vfp::RightProjectors,
+        ULfp::LeftProjectors,
+        VRfp::RightProjectors,
+    )
+
+TODO: write this up.
+"""
```
pbrehmer, line 530: Perhaps like this? I decided to omit the unicode diagrams here because it would look rather cluttered I think (and because I was a bit lazy).

```suggestion
"""
    generate_halfinfinite_characteristic_equation(
        ::CTMRGAlgorithm{<:HalfInfiniteProjector},
        iSfp::CornerTensors,
        Ufp::LeftProjectors,
        Vfp::RightProjectors,
        ULfp::LeftProjectors,
        VRfp::RightProjectors,
    )

Takes the fixed-point values of the inverse singular values `iSfp`, the left and right isometries `Ufp`
and `Vfp`, and their null spaces `ULfp` and `VRfp` corresponding to a converged CTMRG contraction,
and generates a function ``F(s, C, E, u, S, v)`` which characterizes the convergence of the CTMRG algorithm in terms of the characteristic equation ``F(s, C, E, u, S, v) = 0``. Here, ``s`` corresponds to a
state variable (e.g. an `InfinitePEPS` that is being optimized), and ``(C, E, u, S, v)`` represents a CTMRG
contraction environment on a generic unit cell meaning that all tensors have a directional and
unit cell index. ``C`` and ``E`` directly represent the corner and edge tensors, while ``u`` and ``v``
parametrize differentiable projectors ``U = U_{fp} + U_{L,fp} u`` and ``V = V_{fp} + V_{L,fp} V``, and ``S`` denotes the singular values of the decomposed environment.

``F`` returns a tuple of five tensor arrays, corresponding to equations for ``C``, ``E``, ``u``, ``S`` and
``v``, respectively, as shown in Eqs. (76)-(80) in [arXiv:2607.15030](@cite burgelman_implicit_2026).
"""
```

src/algorithms/contractions/ctmrg/characteristic_equations.jl
```diff
@@ -0,0 +1,684 @@
...
+    @autoopt @tensor EwCEnVd[χ_WSW D_S; χ_E] :=
+        Ew[χ_WSW D_W; χ_WNW] *
+        C[χ_WNW; χ_NNW] *
+        EnVd[χ_NNW D_W D_S; χ_E]
+    return EwCEnVd
+end
+
+# north edge renormalization
+function contract_E´(Ud::RightProjector{S, 3}, EnVd::LeftProjector{S, 5}) where {S}
+    @tensor E´[-1 -2 -3; -4] := Ud[-1; 1 2 3] * EnVd[1 2 3 -2 -3; -4]
```
lkdvos, line 46: This looks slightly fishy for fermions, but I might also be wrong. For the CTMRG contractions, when we use `U'` and `Vd'`, don't we use `*` instead of contractions?

src/algorithms/contractions/ctmrg/characteristic_equations.jl
```diff
@@ -0,0 +1,684 @@
...
+        # evaluate partial contractions to reuse
+        # north edge and its left projector
+        EnVd = contract_EnVd(E, U, O)
+        EwCEnVd = contract_EwCEnVd(Ep, C, EnVd)
+
+        # F1: corner
+        C´ = Ud * EwCEnVd
+        C´ = _project_hermitian(C´) # project output
+        λ_C = dot(C, C´)
+        F1 = C´ / λ_C - C
```
lkdvos, line 129:
```suggestion
        F1 = add!!(C´, C, -1, inv(λ_C))
```
I'm not entirely sure, but we might have rules set up for this such that it remains AD-able

lkdvos: same suggestion on the other parts in this function

src/algorithms/contractions/ctmrg/characteristic_equations.jl
```diff
@@ -0,0 +1,684 @@
...
+        for (c, b) in blocks(C)
+            copyto!(block(F, c), _squareroot_pullback(b))
+        end
+        return NoTangent(), P_t(_elementwise_mult(ΔC, F))
+    end
+    return C, squareroot_pullback
+end
+function _squareroot_pullback(C::AbstractMatrix)
+    Fdata = similar(C)
+    for i in axes(Fdata, 1), j in axes(Fdata, 2)
```
lkdvos:
```suggestion
    for j in axes(Fdata, 2), i in axes(Fdata, 1)
```

src/algorithms/contractions/ctmrg/characteristic_equations.jl
```diff
@@ -0,0 +1,684 @@
...
+        renormalize_northwest_corner((r, c), EC, PL, PR)
+    elseif dir == NORTHEAST
+        renormalize_northeast_corner((r, c), EC, PL, PR)
+    elseif dir == SOUTHEAST
+        renormalize_southeast_corner((r, c), EC, PL, PR)
+    elseif dir == SOUTHWEST
+        renormalize_southwest_corner((r, c), EC, PL, PR)
+    end
+    C′ /= dot(env.corners[coordinate...], C′)
+    return C′ - env.corners[coordinate...]
```
lkdvos: here also the `add!!` suggestion

src/algorithms/contractions/ctmrg/characteristic_equations.jl
```diff
@@ -0,0 +1,684 @@
...
+        return NoTangent(), P_t(_elementwise_mult(ΔC, F))
+    end
+    return C, fourthroot_pullback
+end
+function _fourthroot_pullback(C::AbstractMatrix)
+    Fdata = similar(C)
+    for i in axes(Fdata, 1), j in axes(Fdata, 2)
+        # Taking the diagonal only is okay, when dA is diagonal anyway: Fdata[i, i] = 1 / (4 * conj(C[i, i]^3))
+        # However, for Q-deformed CTMRG, we need the full version:
+        Fdata[i, j] = 1 / conj(C[i, i]^3 + C[i, i] * C[j, j]^2 + C[i, i]^2 * C[j, j] + C[j, j]^3)
```
lkdvos: I think here it might be worth it to be slightly more careful about the memory, since we're now allocating a block in `_fourthroot_pullback` to immediately copy into the destination of `F`, which is then copied again in `elementwise_mult`, and I guess this is not really that heavy of toll it still kind of adds up.

It also feels like there is probably a broadcasting way of writing this, which makes it GPU compatible as well as removing the intermediates, something like this might work? (untested)

```julia
Cd = diagview(C)
Ct = transpose(Cd)
@. F = ΔC / conj(Cd^3 + Cd * Ct^2 + Cd^2 * Ct + Ct^3)
```

src/algorithms/optimization/implicit_differentiation.jl
```diff
@@ -360,3 +354,331 @@ function fixedpoint_gradient(∂E∂x, ∂f∂x, ∂f∂A, x₀, alg::KrylovKit.
...
+
+    # prepare iterating function corresponding to a single gauge-fixed CTMRG iteration
+    alg_fixed = _set_fixed_truncation(alg) # fix spaces during differentiation
+    alg_gauge = _scrambling_env_gauge(alg) # select appropriate gauge-fixing algorithm
+    env_conv, info = ctmrg_iteration(InfiniteSquareNetwork(state), env, alg_fixed)
+    signs, corner_phases, edge_phases = compute_gauge_fix_gauge(env_conv, env, alg_gauge)
+    env_fixed = fix_phases(env_conv, signs, corner_phases, edge_phases)
+
+    # NOTE: explicitly keeping corner non-diagonal and (possibly) complex, for use in the backwards pass
+    C, E = only(env_fixed.corners[NORTHWEST, :, :]), only(env_fixed.edges[NORTH, :, :])
```
lkdvos: Is there a reason for doing the slicing with `:` and then immediately calling `only`? would `env.corners[NORTHWEST, 1, 1]` not give the same result?

src/algorithms/optimization/implicit_differentiation.jl
```diff
@@ -360,3 +354,331 @@ function fixedpoint_gradient(∂E∂x, ∂f∂x, ∂f∂A, x₀, alg::KrylovKit.
...
+    vjp_state(x) = F_vjp(x)[2] # state pullback
+
+    function leading_boundary_implicit_pullback((_Δenv, _Δinfo))
+        Δenv, Δinfo = unthunk(_Δenv), unthunk(_Δinfo)
+
+        # accumulate adjoints from all corners & edges through _c4v_env constructor
+        _, ΔC, ΔE = c4v_env_vjp(Δenv)
+
+        # apply Hermitian projectors to edge and corner cotangents
+        ΔE = _project_hermitian(ΔE)
```
lkdvos, line 519: I think the hermitian projectors in the versions of TensorKit we are using now are correctly differentiable

src/algorithms/optimization/implicit_differentiation.jl
```diff
@@ -360,3 +354,331 @@ function fixedpoint_gradient(∂E∂x, ∂f∂x, ∂f∂A, x₀, alg::KrylovKit.
...
+        ΔC = _project_hermitian(ΔC)
+
+        # extract cotengent of the isometry variable and verify that it is trivial
```
lkdvos:
```suggestion
        # extract cotangent of the isometry variable and verify that it is trivial
```

src/algorithms/optimization/implicit_differentiation.jl
```diff
@@ -360,3 +354,331 @@ function fixedpoint_gradient(∂E∂x, ∂f∂x, ∂f∂A, x₀, alg::KrylovKit.
...
+
+    # gauge-fix SVD isometries
+    env_conv, info = ctmrg_iteration(InfiniteSquareNetwork(state), env, alg)
+    signs, = compute_gauge_fix_gauge(env_conv, env, ScramblingEnvGauge())
+    S = normalize.(info.S)
+    U, V = fix_relative_phases(info.U, info.V, signs)
+
+    # pretend the singular values matrices are just arbitrary complex tensors
+    s = TensorMap.(S)
+    if real(scalartype(state)) != scalartype(state)
```
lkdvos:
```suggestion
    if !(scalartype(state) <: Real)
```

personal preference, doesn't matter too much

src/algorithms/optimization/implicit_differentiation.jl
```diff
@@ -360,3 +354,331 @@ function fixedpoint_gradient(∂E∂x, ∂f∂x, ∂f∂A, x₀, alg::KrylovKit.
...
+function _gauge_fix_c4v_projector(::C4vCTMRG{<:C4vEighProjector}, signs, info)
+    return info.V * signs[1]'
+end
+function _gauge_fix_c4v_projector(::C4vCTMRG{<:C4vQRProjector}, signs, info)
+    return info.Q * signs[1]'
+end
+
+function _check_algorithm_combination(::SequentialCTMRG, ::ImplicitGradient)
+    msg = "The `:ImplicitGradient` algorithm is not yet compatible with the `SequentialCTMRG` algorithm; select \
+          SimultaneousCTMRG instead to use :ImplicitGradient"
```
pbrehmer, line 418: As you suggested, I think we could just try to return the SVD tensors in the `ctmrg_iteration` of `SequentialCTMRG`, right? I'm still not super sure if this will actually lead to an element-wise converged environment but I think it should.

src/algorithms/optimization/implicit_differentiation.jl
```diff
@@ -360,3 +354,331 @@ function fixedpoint_gradient(∂E∂x, ∂f∂x, ∂f∂A, x₀, alg::KrylovKit.
...
+* `maxiter::Int=$(Defaults.gradient_maxiter)`
+* `verbosity::Int=$(Defaults.gradient_verbosity)`
+* `solver_alg::Union{Algorithm,NamedTuple}=(; alg::Symbol=:$(Defaults.gradient_fixedpoint_solver_alg))`: solver algorithm for the `ImplicitGradient` gradient algorithm.
+    - `:GMRES` : GMRES iterative linear solver, see [`KrylovKit.GMRES`](@extref) for details
+    - `:BiCGStab` : BiCGStab iterative linear solver, see [`KrylovKit.BiCGStab`](@extref) for details
+"""
+struct ImplicitGradient{A} <: GradientAlgorithm{A}
+    solver_alg::A
+end
+ImplicitGradient(; kwargs...) = GradientAlgorithm(; alg = :ImplicitGradient, kwargs...)
```
pbrehmer, line 398: Can we also add the `:ImplicitGradient` option to the [`fixedpoint` docstring](https://github.com/QuantumKitHub/PEPSKit.jl/blob/7c41e831b93109072ec78584728f818229bc5250/src/algorithms/optimization/peps_optimization.jl#L134-L135)? Something like:
```
- `:ImplicitGradient` : Compute the gradient via implicit differentiation of a characteristic set of equations, see [`ImplicitGradient`](@ref)
```

pbrehmer (commented):
Looks really good to me! I would personally like to give `SequentialCTMRG` with implicit differentiation a try but we can of course also postpone this to a later PR.

src/algorithms/contractions/ctmrg/characteristic_equations.jl
leburgel: Yes, this definitely broken for fermions. But I think currently everything about the C4v case is broken for fermions. Currently we're not allowing anyone to run a `leading_boundary` for any `Fermionic()` braiding style by throwing an explicit argument error. So it's not possible to get here in the first place when optimizing a fermionic network.

I can put another check on the braiding style at the start of the `generate_symmetric_characteristic_equation` function if that's better.

src/algorithms/optimization/implicit_differentiation.jl
leburgel: I can use the TensorKit projector for the corners, but not for the edges. For the latter it's a very shady definition of "Hermitian" which assumes additional trivial spaceflips and acts on a tensor map that is very much not square, but there's no real way around that I think.

src/algorithms/optimization/implicit_differentiation.jl
leburgel: I just liked the fact that it would error if you accidentally put in an environment with a non-trivial unit cell, but I'm not sure if it's even possible that this wouldn't be caught before ending up here.

src/algorithms/contractions/ctmrg/characteristic_equations.jl
leburgel: Well, it seems this tries to differentiate straight through `LinearAlgebra.BLAS.axpby!` and then crashes. I can see what I think are VectorInterface rules in TensorKitMooncakeExt, but it seems these just aren't there for ChainRules.

src/algorithms/optimization/implicit_differentiation.jl
pbrehmer: Okay I gave this a try and didn't manage to get it to work with `ImplicitGradient`. The characteristic equations are apparently not fulfilled which I somewhat suspected. On the other hand, `SequentialCTMRG` does seem to be compatible with `FixedPointGradient`. I guess now the question is if we still want `SequentialCTMRG` to return the SVD tensors and perhaps try to circumvent the problems by running one iteration of `SimultaneousCTMRG`.

src/algorithms/contractions/ctmrg/characteristic_equations.jl
leburgel: This seems to work. I put it in in a way that leaves one `copyto!` per block. I don't think we can do any better than that?

src/algorithms/optimization/implicit_differentiation.jl
leburgel: I tried this quickly, and just running one iteration of `SimultaneousCTMRG` doesn't really solve the problem. It seems the `SequentialCTMRG` result is just not a fixed point of `SimultaneousCTMRG`. If you start the simultaneous algorithm from the sequential fixed point, the whole thing seems to converge from scratch.  For implicit differentiation, I think we would need a custom set of characteristic equations for the sequential fixed point, which really takes into account what the algorithm actually does. But I don't exactly feel like spending time to figure that out at the moment.

Given that we currently have no use for the SVD tensors in the sequential procedure, I think we can just remove them again here and go back to not returning here. They're in the commit history here, and if we have a go at proper sequential characteristic equations later we can always find them back.

pbrehmer: Somewhat annoying but then we'll figure it out in the future. In that case I'd also be in favor of again removing the SVD tensor return from `SequentialCTMRG`. Thanks for trying!

leburgel: Removed them again, we can come back to this later.

leburgel:
Windows CI is completely refusing to work for some reason, but other than that I think this has pretty much everything I had in mind for now.

src/algorithms/optimization/implicit_differentiation.jl
leburgel: C4v CTMRG has an input check on the size of the network, I forgot about that but just confirmed now to be safe. So this will always just be 4x1x1 arrays at this point. Removing the `only` again.

src/algorithms/contractions/ctmrg/characteristic_equations.jl
leburgel: I checked to be sure, and just attempting to run a C4v leading boundary with a fermionic network throws an argument error right now. So this is safe for now, but I'll add a note indicating that this should be revisited if fermions + C4v would be supported at some point.

pbrehmer (approved):
Looks good to merge to me now! The Windows CI problems look like they are not really related to this PR in particular and rather some CI setup issue, if I'm not mistaken.

src/algorithms/optimization/implicit_differentiation.jl
```diff
@@ -360,3 +356,328 @@ function fixedpoint_gradient(∂E∂x, ∂f∂x, ∂f∂A, x₀, alg::KrylovKit.
...
+    signs, = compute_gauge_fix_gauge(env_conv, env, ScramblingEnvGauge())
+    S = normalize.(info.S)
+    U, V = fix_relative_phases(info.U, info.V, signs)
+
+    # pretend the singular values matrices are just arbitrary complex tensors
+    s = TensorMap.(S)
+    if !(scalartype(state) <: Real)
+        # complex network, complex things, required to make the derivatives work...
+        s = complex.(s)
+    end
```
lkdvos: Does this have to be differentiable? Otherwise something like this might be slightly more efficient by fusing both the complexification and the non-diagonalization:


```suggestion
    s = map(S) do s
        dst = similar(s, scalartype(state), space(s)) # TensorMap
        return copy!(dst, s)
    end
```

lkdvos (approved):
Left some minor comment and probably there are still some similar things, but I'd be okay with leaving that for follow-up with some actual profiling as well, no need to keep this stalled.

src/algorithms/optimization/implicit_differentiation.jl
leburgel: This is just a preprocessing step, doesn't need to be differentiable. So good point!

PR #415, opened 2026-08-12
[WIP] Boundary-MPS approximate contraction of an iPEPO window with CTMRG environment

Yue-Zhengyuan:
This PR uses the finite-MPO/finite-MPS zip-up compression introduced in https://github.com/QuantumKitHub/MPSKit.jl/pull/470.

## Summary

This PR adds approximate expectation-value contractions for a single-layer `InfinitePEPO` using finite boundary MPSs and a CTMRG environment:

- `MPOObservable` represents a physical open-boundary MPO acting along a path in the 2D network.
- `expectation_value_approx` contracts an `MPOObservable` in its enclosing rectangular window.
- `correlator_approx` measures a dense two-site operator between one fixed first site and many second sites, reusing partial contractions with adaptive window depth and fixed width.

Each PEPO row becomes a finite MPO, which is applied to a boundary MPS with zip-up compression and optional one-site DMRG refinement.
The implementation includes virtual-space flips and twists, with U(1) and FermionParity regression coverage.

### Path-based MPO observables

`MPOObservable` stores:

- `sites`: lattice sites on which physical MPO tensors act;
- `mpo`: the open-boundary MPO tensors, with `mpo[k]` acting on `sites[k]`;
- `path`: a non-self-intersecting nearest-neighbor path containing the operator sites in MPO order and any intermediate sites carrying the virtual string.

Constructors accept explicit MPO tensors, an explicitly routed path, or a dense `AbstractTensorMap`.
The dense constructor orders the sites and operator legs, decomposes the operator, and connects consecutive sites by horizontal-first paths.
Automatic routing is limited: it rejects paths that revisit a site or pass through another operator site out of order.
Explicit paths allow other valid routes, including paths that cross between two rows more than once at different columns.

The functions `mpo_path_first`, `mpo_path_middle`, `mpo_path_last`, and `mpo_path_string` insert the operator or route its string, trace the PEPO physical legs, and fuse the MPO string with the appropriate PEPO virtual legs.
This avoids materializing a separate `BraidingTensor` at each routing site.
Fusing these legs is not necessarily the most efficient contraction strategy, particularly at routing sites.

### Finite-window expectation values

```julia
expectation_value_approx(rho, observable, env; trunc, maxiter = 1, direction = :auto)
```

For north-to-south contraction, CTMRG corners and north edges form the initial `FiniteMPS`.
Each row includes the west/east CTMRG edges and the traced or operator-modified PEPO tensors.
After zip-up compression and optional DMRG refinement, the final MPS is contracted against the south CTMRG boundary.
The expectation value is normalized by a separate observable-free contraction of the same window.

Column contraction reuses this implementation by rotating the state, environment, and observable.
Automatic direction selection uses rows for wide or square windows and columns for tall windows; callers can override it with `:rows` or `:columns`.
The south boundary is stored in the adjointed representation expected by the MPS contractions.

`WindowApprox` is an internal wrapper for the zip-up and optional DMRG algorithms.
For both public measurement functions, the default `trunc` limits the rank to the largest CTMRG boundary dimension, and `maxiter = 0` disables DMRG refinement.

### Two-site correlators with adaptive depth

```julia
correlator_approx(rho, op, i, js, env; trunc, maxiter = 1)
```

The interface follows `correlator`: `i` is a fixed `CartesianIndex{2}`, and `js` is a vector of Cartesian indices, a `CartesianIndices` collection, or a single Cartesian index.
A collection returns a vector in `vec(js)` order; a single second site returns a scalar.
Targets must be nonempty, unique, and distinct from `i`.
Operator leg 1 always acts at `i`, and leg 2 at each `j`.

The contraction direction is automatic, with no `direction` keyword:

- Row contraction is valid if all targets satisfy `j[1] ≥ i[1]`.
- Column contraction is valid if all targets satisfy `j[2] ≤ i[2]`.
- If both are valid, the rectangle enclosing all sites determines the choice: rows for wide or square rectangles, columns for tall rectangles.
- If neither is valid, the call raises an `ArgumentError`.

For row contraction, every window has the same column range, covering `i` and all targets, but ends at its own target row.
The operator string travels south along the source column and then horizontally along the target row.
Column contraction uses the rotated equivalent: a fixed row range and a window ending at each target column.

A single forward loop maintains two north states: one without observables for normalization, and one carrying the source insertion and open MPO string.
At each target row, the algorithm constructs the CTMRG south boundary directly below it, closes the targets using shared horizontal environments, and normalizes them with the observable-free contraction using the same row closure.
The target row is closed without further truncation in both numerator and denominator.
This avoids mixing a target-row numerator with a differently truncated full-sweep denominator, which can otherwise produce identity expectations far from one.

The implementation reuses the operator decomposition, forward propagation, and horizontal contractions within each target row.
It does not store previous row MPOs or boundary histories and requires no reverse sweep or adjoint row MPOs.
Adding deeper targets preserves earlier results as long as the fixed width and automatically selected direction remain unchanged.

## Validation

Tests cover independent adaptive-window references, target ordering, column rotation on rectangular unit cells, propagation through rows without targets, and preservation of earlier values when the depth is extended.
Truncated identity checks cover U(1) and FermionParity, both contraction directions, and optional DMRG refinement.
A separate physical-state regression compares against `expectation_value`.
MPO routing and finite-window tests also cover turns, reversed ordering, and a U-shaped path.

The two correlator test files passed locally with 27 assertions, run serially with `--jobs=1`.

## Remaining work

- Generalize the contraction backend to iPEPS and purified iPEPO networks.
- Improve automatic path routing and explore alternatives to fusing MPO strings into PEPO virtual legs.


Yue-Zhengyuan:
@lkdvos @leburgel Does the MPSKit support for fermions (via planar contraction) still work if the MPS/MPO arrows do not follow the standard directions?

leburgel:
> @lkdvos @leburgel Does the MPSKit support for fermions (via planar contraction) still work if the MPS/MPO arrows do not follow the standard directions?

Yes, I think it should just work regardless of which spaces exactly are dual.

Yue-Zhengyuan:
Fermion support is now here, but the way to achieve it is not ideal. I first put the virtual arrow directions in the iPEPO in the standard direction. But the on the right boundary, the "physical space" is dual, as required by the CTMRGEnv constructor (see C₂ below):

```
    [1; 2]      [1 2; 3]        [1; 2]
    C₁-←-2      1-←-E₁-←-3      1-←-C₂
    ↓               ↓               ↑
    1               2               2
```

Thus I still need to use many planar operations (some of which I still don't fully understand) when constructing the left and right ends of each row, and the south boundary.

(Edit: now I also flip the east envspace to standardize all arrows. Things then make more sense.)

Otherwise, if I dispatch away from planar operations, it seems that I need to overload lots of things in MPSKit (not only in `ZipUp` but also `DMRG` for further refinement), which may not be worthwhile. (But anyway we still need to do it when generalizing to approximate contraction of multi-layer networks?)

leburgel:
I was going to take some time to go over this, but it seems this PR is rather large and I completely misjudged the time I allocated. I'll try to go over the details as soon as I can, but maybe I can start with a few general remarks already before I submit a full review:

---

As far as I can tell, the newly introduced `MPOObservable` could in spirit serve as a term in a `LocalOperator`, where instead of associating a dense `TensorMap` to a group of sites we associate a finite MPO (encoded here as a vector of MPO tensors). The only difference is the `path` field that is additionally included in the `MPOObservable` struct. This leads me to the question: what is the benefit of being able to manually specify a path, over just determining this automatically? Also, what exactly is the difficulty in this automatic path finding (as indicated in the note above)? If it's not more expensive to always use a default path, then we can directly use MPOs as `LocalOperator` terms, whose expectation value we could then evaluate exactly or approximately. This would be very nice to have, hence my question.

---

For the finite window contraction, would it be possible to directly pass an algorithm for this? I can imagine wanting to just use `DMRG2` (or even just `DMRG`) to do this, rather than starting from `ZipUp` and then optionally refining. Probably the way it is now is fine for practical purposes, at least the ones you had in mind, but it feels like it would be useful to have direct control over the algorithm.

---

Regarding fusing the MPO path, instead of fusing everything into the virtual spaces, we could (at some point) try to keep the virtual MPO legs around explicitly as auxiliary legs for MPS tensors when they are encountered. The problem is then to contract two "matching" auxiliary legs corresponding to the same virtual MPO bond when they encounter each other in two tensors being contracted. This is a lot more tedious to do in our framework, since we don't have "labeled" tensor indices that recognize each other automatically. Here we can just pass around explicit labels for all of the auxilary legs that come from a virtual MPO bond, and keep track of them. Not sure how feasible this is, and definitely not sure if this is something we want to do here, but just wanted to bring it up already.

---

Regarding fermion support, I'm not sure if I'm entirely following the discussion. I think the MPSKit routines should just be able to handle physical spaces with a "non-standard" duality. Using planar operations where possible should ensure this just works, in the way that is consistent with MPSKit conventions. Therefore, I don't really see the need to consider dispatching away from planar operations. The only reason for this as far as I'm concerned would be to ensure differentiability, which is not a goal at all for now I think. So even flipping this east env space shouldn't be necessary, I think.

The only reason to not use planar operations should be if we simply can't, e.g. for multi-layer networks. There, we have to overload the core contractions used in MPSKit anyway, but we additionally have to put additional twists to ensure the result is what is expected by MPSKit conventions. For the MPSKit contractions already overloaded in PEPSKit, this should have been done in a way that they can also handle arbitrary arrow directions.

---

For multiplying in the bottom boundary, I would really like to avoid (doubly) conjugating edge tensors, this always leads to issues sooner or later in my experience. The question then becomes if we need to twist anything on the resulting top MPS (after all of the MPO applications) to ensure we can just contract in south edges in the usual way and actually get what we want. I'm not entire sure about that.

src/algorithms/contractions/mpo_path/pepo_1layer.jl
```diff
@@ -0,0 +1,238 @@
...
+    direction === :south && return SOUTH
+    direction === :west && return WEST
+    throw(ArgumentError("invalid MPO path direction: $direction"))
+end
+
+"""
+Return the tensor-expression label for the PEPO virtual leg in a cardinal direction.
+"""
+function _mpo_path_virtual_label(direction::Symbol)
+    return (:N, :E, :S, :W)[_mpo_path_direction(direction)]
```
leburgel: I would probably just use the same kind of short circuited switches as above, instead of instantiating and indexing this tuple on every call.

src/algorithms/contractions/mpo_path/pepo_1layer.jl
```diff
@@ -0,0 +1,238 @@
...
+    A_e = tensorexpr(:A′, (:din, :dout), (:N, :E, :S, :W))
+    F_e = tensorexpr(:F, fused_label, (virtual_label, :r))
+    rhs = Expr(:call, :*, op_e, A_e, F_e)
+    contraction = macroexpand(
+        @__MODULE__, :(return @tensoropt $result_e := $rhs)
+    )
+
+    return quote
+        _check_pepo_first_physicalspace(A, op)
+        A′ = twistdual(A, 2)
```
leburgel: I find it a bit fishy to let the generated expression depend directly on the choice of variable name. I know these generated functions for contractions are inherently fishy, but I think it would at least be more transparent/safer to split off the regular function part in the beginning of the quote here, and then call the generated function on the variables, really isolating the generated part without any "normal" function part like assigning variables.

leburgel: Same comment for `mpo_path_last`,  `mpo_path_middle` and `mpo_path_string`, I would more cleanly separate the preprocessing and tensor creation steps from the generated contraction applied to these tensors.

Yue-Zhengyuan:
In principle I should split this PR in two: first focusing only on `expectation_value_approx`, with a follow up with `correlator_approx`. But these two share much infrastructure, so I put them together...

- Path routing for MPO terms

I agree that needing an additional `path` to specify the MPO is not ideal. The difficulty is in writing a clever algorithm that creates a path that go through the MPO sites in order, while also being non-self-intersecting. You can consider the sites (1, 1) - (6, 1) - (2, 0) - (3, 2) - (4, 0) - (5, 2) (which is unlikely to appear in practice, though). It is not obvious to me which should be the "canonical" path joining them without self-intersection.

- Algorithm struct for window contraction

It is already there, called `WindowApprox`. The problem with skipping `ZipUp` is how you initialize a `ψ0` for `DMRG`/`DMRG2` without randomly guessing.

Currently this struct is not exported. Instead, we provide kwargs `trunc` and `maxiter` in the API to build the struct internally:
```julia
WindowApprox(Zipup(; trunc), _approx_dmrg(maxiter))
```

- Fusing MPO virtual space with the PEPO

I want to postpone this. Currently `ZipUp` (and `DMRG`?) only support(s) MPS with one physical leg per site.

- Fermion support

Changing arrow directions in MPSKit is like planar contraction with flippers without twists. But this is not the case in PEPSKit, where we should use `flip` that preserves `@tensor` calls. First flipping all arrows to the standard direction appears to automatically put the required twists back in the network; the change of arrows is more like a by-product that is unimportant to MPSKit. I suggest we keep this pre-processing for now, and only revisit it at the end.

- Conjugation of south boundary

In the cache `WindowRowCache` used by `correlator_approx`, we not only need the south CTM boundary, but also need the contraction of the south boundary with some rows above it. So avoiding conjugating south boundary twice will require more than a custom version of `dot`. We will also need to write `ψ * O`, where the MPO `O` acts on the *bra*-state `ψ`. This is a bit redundant to me.

src/algorithms/contractions/mpo_path/pepo_1layer.jl
leburgel: I think it would look a bit cleaner to just use the same kind of short circuited switches as above, instead of instantiating and indexing this tuple on every call.

src/algorithms/contractions/mpo_path/pepo_1layer.jl
```diff
@@ -0,0 +1,238 @@
...
+        F = _mpo_path_outgoing_fuser(
+            fuser(storagetype(A), domain(A, $direction_index)', space(op, 3)),
+            $direction_index,
+        )
+        $contraction
+    end
```
leburgel: This is more a personal preference, but since these kinds of generated functions are already hard to follow as is, I think it would be more transparent/safer to split off the regular function part in the beginning of the quote here, and then call the generated function on the variables, really isolating the generated part without any "normal" function part like assigning variables.

Same comment for `mpo_path_last` and `mpo_path_middle`, I think it would look a bit less confusing if only the actual contraction expression builder was captured in a generated function.

src/algorithms/contractions/window/twosite/caching.jl
```diff
@@ -0,0 +1,124 @@
...
+        ρ::InfinitePEPO, env::CTMRGEnv,
+        rowrange::UnitRange{Int}, colrange::UnitRange{Int}, alg::WindowApprox,
+    )::WindowRowCache
+    row_mpos = Dict(
+        row => _row_mpo(ρ, nothing, env, row, colrange)
+            for row in rowrange
+    )
+    nrows = length(rowrange)
+    north = _north_boundary_mps(env, first(rowrange), colrange)
+    south = _south_boundary_mps(env, last(rowrange), colrange)
```
leburgel: I would personally implement this through rotation without conjugation, which reverses the left-right ordering of the south edge tensors and collects them into a finite MPS. This respects the natural virtual orientation of the south edge tensors, and avoids having to take the "adjoint" of individual MPS tensors.

src/algorithms/contractions/window/twosite/caching.jl
```diff
@@ -0,0 +1,124 @@
...
+    for (k, row) in enumerate(rowrange)
+        north_prefixes[k + 1] = _approximate_window_step(
+            row_mpos[row], north_prefixes[k], alg
+        )
+    end
+
+    south_suffixes = Vector{typeof(south)}(undef, nrows + 1)
+    south_suffixes[end] = south
+    for (k, row) in Iterators.reverse(enumerate(rowrange))
+        W = _adjoint_mpo(row_mpos[row])
```
leburgel: Instead of computing an MPO adjoint here, I would prefer to rotate the finite MPO by 180°, and use this to compute the product with the south MPS. This will allow us to similarly avoid conjugating the south boundary MPS.

src/algorithms/contractions/window/twosite/pepo_1layer.jl
```diff
@@ -0,0 +1,251 @@
+# Source-cached dense two-site contractions for single-layer PEPO networks.
+
+"""
+Validate a dense two-site measurement, choose its sweep orientation, and dispatch to the
+row-oriented source-cached contraction.
+"""
+function _correlator_approx(
+        ρ::InfinitePEPO, op::AbstractTensorMap,
+        bonds::Vector{NTuple{2, CartesianIndex{2}}},
```
leburgel: Why would these be called `bonds`? If I understand correctly these represent sites of physical spaces, which we usually call `inds`. I'm fine with a different name, it's just that `bonds` to me more implies virtual bonds, so connections between sites rather than sites themselves.

src/algorithms/contractions/window/twosite/caching.jl
```diff
@@ -0,0 +1,124 @@
...
+- `south_suffixes`: `nrows + 1` adjointed south boundary MPSs.
+  Entry `k + 1` is below row `k` in the window.
+- `norm`: the approximate contraction of the window with no observable inserted.
+"""
+struct WindowRowCache{M <: FiniteMPO, N <: FiniteMPS, S <: FiniteMPS, T <: Number}
+    rowrange::UnitRange{Int}
+    colrange::UnitRange{Int}
+    row_mpos::Dict{Int, M}
+    north_prefixes::Vector{N}
+    south_suffixes::Vector{S}
```
leburgel: I'm not sure I like the `prefix` and `suffix` names. I would prefer `north/south_boundaries`, or `north/south_mpss`, I find this a bit more clear.

src/algorithms/contractions/window/twosite/caching.jl
```diff
@@ -0,0 +1,124 @@
...
+            for row in rowrange
+    )
+    nrows = length(rowrange)
+    north = _north_boundary_mps(env, first(rowrange), colrange)
+    south = _south_boundary_mps(env, last(rowrange), colrange)
+
+    north_prefixes = Vector{typeof(north)}(undef, nrows + 1)
+    north_prefixes[1] = north
+    for (k, row) in enumerate(rowrange)
+        north_prefixes[k + 1] = _approximate_window_step(
```
leburgel: Also the name here isn't really clear to me. This really is an `approximate` run, so both `window` and `step` are a bit confusing to me. I'd be find with just `_approximate` even if we don't want to shadow `MPSKit.approximate`, but `window` can be implied by typing the implementation for finite MPS/MPO, and `step` seems off because it might sound like you're doing one "step" of an approximate (e.g. one iteration of VOMPS).

src/algorithms/contractions/window/twosite/caching.jl
leburgel: It just requires carefully reversing the south boundary MPS if we want to reuse exactly the same `_approximate_window_step` method.

src/algorithms/contractions/window/twosite/caching.jl
```diff
@@ -0,0 +1,124 @@
...
+
+- `rowrange` and `colrange`: the coordinate ranges defining the window.
+- `row_mpos`: the row MPOs without observables, one per row and including the west and east CTMRG edges.
+- `north_prefixes`: `nrows + 1` north boundary MPSs.
+  Entry `k` is above row `k` in the window.
+- `south_suffixes`: `nrows + 1` adjointed south boundary MPSs.
+  Entry `k + 1` is below row `k` in the window.
+- `norm`: the approximate contraction of the window with no observable inserted.
+"""
+struct WindowRowCache{M <: FiniteMPO, N <: FiniteMPS, S <: FiniteMPS, T <: Number}
```
leburgel: This is all semantics, but I think "patch" might be more appropriate than "window" here, since it's finite in both directions? To me "window" sounds more like something that's finite in both directions.

src/algorithms/expval_approx.jl
```diff
@@ -0,0 +1,60 @@
...
+
+"""
+Return the smallest row and column ranges containing a collection of lattice sites.
+"""
+function _window_ranges(sites)
+    rows = getindex.(sites, 1)
+    cols = getindex.(sites, 2)
+    return UnitRange(extrema(rows)...), UnitRange(extrema(cols)...)
+end
+
```
leburgel, line 40: Just a note, we should reuse the same method everywhere, also in the other expectation value code. See also `_patch_ranges` in #423, we can reuse just one depending on which one of these lands first.

src/algorithms/contractions/window/twosite/caching.jl
```diff
@@ -0,0 +1,124 @@
...
+
+    south_suffixes = Vector{typeof(south)}(undef, nrows + 1)
+    south_suffixes[end] = south
+    for (k, row) in Iterators.reverse(enumerate(rowrange))
+        W = _adjoint_mpo(row_mpos[row])
+        south_suffixes[k] = _approximate_window_step(
+            W, south_suffixes[k + 1], alg
+        )
+    end
+    norm = dot(south, north_prefixes[end])
```
leburgel: Also this we should be able to avoid by rotating over taking adjoints.

src/algorithms/contractions/window/pepo_1layer.jl
```diff
@@ -0,0 +1,189 @@
...
+    tensors = [insertleftunit(W, 1)]
+    append!(
+        tensors,
+        (
+            _window_site_tensor(ρ, observable, row, col)
+                for col in colrange
+        ),
+    )
+    E = permute(
+        flip(edge(env, EAST, row, cmax + 1), (1, 3)),
```
leburgel, line 188: I would really have to look at these kinds of `flip`s in detail to see why this is fine and/or unavoidable. As long as it works I guess that's not the most important thing right now, but seeing these things does make me wonder every time...

src/algorithms/contractions/window/tools.jl
```diff
@@ -0,0 +1,96 @@
...
+
+"""
+Apply a finite MPO to a finite MPS with zip-up truncation and optional DMRG refinement.
+"""
+function _approximate_window_step(W::FiniteMPO, ψ::FiniteMPS, alg::WindowApprox)
+    ψ′, = approximate((W, ψ), alg.zipup)
+    isnothing(alg.dmrg) && return ψ′
+    ψ′, = approximate(ψ′, (W, ψ), alg.dmrg)
+    return ψ′
+end
```
leburgel: Circling back to this, I'd actually really prefer it if we could just use `MPSKit.approximate` for this. I know this is not really possible right now, but it shouldn't take to much effort to make it possible, and it should clean this up a bit.

The main thing relevant to this PR is that I just don't see much practical reason to run a `DMRG` refinement after a `ZipUp` pass. I know DMRG-based approximation has the benefit of being variationally optimal if the MPO is Hermitian, but in practice I would only use DMRG approximation as a slightly cheaper alternative to SVD compression (at least usually it is a bit cheaper). So from my usual point of view, I don't see the real use of the optional additional refinement with DMRG.

The implementation issue is mostly due to the differing calling signatures for zip-up compression, and the issue with picking an initial state for a DMRG approximate. Given the recent updates to DMRG to actually allow bond expansion during the run (https://github.com/QuantumKitHub/MPSKit.jl/pull/438), I think we can pretty much solve this by plumbing `approximate((W, ψ), ::DMRG) -> approximate(ψ, (W, ψ), ::DMRG)` and ensuring we have a proper expansion set up, and passing a `trunc` keyword to DMRG exactly like we would for `ZipUp`.

I'm not necessarily saying we need to do all this now or wait for it, but I just still feel we could do without the extra `WindowApprox` struct, which would feel a bit more natural to me.

src/algorithms/contractions/window/tools.jl
```diff
@@ -0,0 +1,96 @@
...
+"""
+function _bra_mps_tensor(A::MPSTensor)
+    return repartition(A', 2, 1; copy = true)
+end
+
+"""
+Construct the planar adjoint of a finite MPO while restoring MPSKit's local MPO leg partition.
+"""
+function _adjoint_mpo(W::FiniteMPO)
+    return FiniteMPO(map(A -> transpose(A', ((3, 1), (4, 2)); copy = true), parent(W)))
```
leburgel: This is the kind of operation I've always struggled with. I guess it's well defined as in it's the only locally planar operation you could come up with to characterize the adjoint of an MPO, but I still really struggle with this conceptually.

If this is possible to do consistently, I would always prefer to work with rotations as opposed to adjoints.

src/algorithms/contractions/window/tools.jl
```diff
@@ -0,0 +1,96 @@
...
+    ψ′, = approximate(ψ′, (W, ψ), alg.dmrg)
+    return ψ′
+end
+
+"""
+Convert a south-boundary MPS tensor into the stored bra representation expected by `dot` using a planar repartition.
+"""
+function _bra_mps_tensor(A::MPSTensor)
+    return repartition(A', 2, 1; copy = true)
+end
```
leburgel: The action of this exactly corresponds to `_dag(A::GenericMPSTensor` which we have defined somewhere else, for the case of one physical leg. I'm fine with leaving this implementation, but maybe we change the name to `_bra_mps_tensor` everywhere.

I would still prefer not to have to do this, if this is possible in a way that doesn't require all too much work.

src/algorithms/contractions/window/pepo_1layer.jl
leburgel: What exactly would break if we didn't flip here?

src/algorithms/correlator_approx.jl
```diff
@@ -0,0 +1,57 @@
...
+"""
+$(SIGNATURES)
+
+Approximately measure a dense two-site operator on one or more ordered pairs of sites in a single-layer PEPO.
+The first and second operator legs act on the first and second sites of each pair, respectively.
+Multiple pairs are evaluated in one shared window and reuse open-string boundary contractions.
+Ordered pairs in a batched call must be unique.
+"""
+function correlator_approx(
+        ρ::InfinitePEPO, op::AbstractTensorMap, bond::Tuple, env::CTMRGEnv;
```
leburgel: If `op` is two-site, can you type restrict it through the numin/out type parameters?

src/operators/mpo_observable.jl
```diff
@@ -0,0 +1,185 @@
...
+An open-boundary MPO embedded along a non-self-intersecting nearest-neighbor path on the
+square lattice. The tensor `mpo[k]` acts on `sites[k]`; `path` also contains intermediate
+sites that only carry the MPO string.
+
+The first and last MPO tensors use the reduced endpoint partitions `(1, 2)` and `(2, 1)`;
+all intermediate tensors use the standard `(2, 2)` MPO partition.
+"""
+struct MPOObservable{M}
+    sites::Vector{CartesianIndex{2}}
+    path::Vector{CartesianIndex{2}}
```
leburgel, line 13: The main reason why I originally asked if we could get rid of this `path` field is that this would then just become a specific term in a `LocalOperator`. Conceptually, I quite like just being able to call `expectation_value_approx` on any `LocalOperator`, where we then evaluate every term individually in the appropriate way (using some kind of `local_expectation_value_aproximate` hook).

For dense tensors, we could just convert them to an MPO when calling the approximate version. The path routing could then happen inside this local expectation value computation, concerning only the relevant path.

The non-intersecting path routing problem is indeed hard in general, if we're handed an MPO acting on specific sites. But usually, we would want to evaluate some operator which we can represent as a dense tensor map (I assume this is the most common case?), which we then decompose into an MPO and evaluate approximately. If we choose the site order in this decomposition in a way that takes into account the routing that we'll have to do, this makes the problem much easier. So I think we could get rid of the routing difficulties by simply allowing sites that are not ordered according to the lattice ordering. This anyway makes a lot of sense for MPO observables I think, where intuitively we would order the sites to end up with the virtual string of smallest length, which will generally be quite easy to route.

Does this make sense, and sound like a reasonable thing to try?

In the end I would like to not do any path routing and virtual fusing at all, and just pass around the MPO indices appropriately automatically. This is not at all necessary now, but it would be good to already get ahead of the routing issue.

src/operators/mpo_observable.jl
```diff
@@ -0,0 +1,185 @@
...
+function MPOObservable(sites, mpo)
+    sites′ = CartesianIndex{2}[_mpo_observable_site(site) for site in sites]
+    mpo′ = collect(mpo)
+    path = _route_mpo_observable(sites′)
+    return MPOObservable(sites′, path, mpo′)
+end
+
+"""
+Construct an MPO observable from a dense operator. Operator sites are ordered in the same
+way as `LocalOperator` terms, and consecutive sites are connected by horizontal-first
```
leburgel, line 38: If we let go of the operator site ordering being the same as for dense tensor local terms, I think we could simplify things quite a bit?

leburgel (commented):
Sorry it took me so long to get through this, I couldn't seem to find the time to sit down and get through the whole thing in one go.

I have a first round of comments, most of which are related to the general remarks I had a while back. An overview:

- Path routing for MPO terms: I think we can make this problem much easier by giving up on ordering the sites according to the matrix-like linear ordering we use for dense local operators. There the choice doesn't matter much so imposing a canonical ordering makes sense. But for MPOs, choosing the site ordering in a way that makes the routing easy makes much more sense to me.

- Algorithm struct for window contraction: I left a more in-detail comment below, but I think that if we use DMRG with an expansion we can merge things quite cleanly actually.

- Fusing MPO virtual space with the PEPO: definitely fair, fusing is the quickest way to actually get something up and running.

- Fermion support: as always I am still confused by this. If it's verified to work properly, there's not many objections I can raise, so that would be fine for me. We can always revisit how we do things later, but I don't have any conceptual arguments right now.

- Conjugation of the south boundary: again there's a more detailed comment below, but I think we can avoid conjugations by rotating instead. This comes with its own issues, but I think it's conceptually cleaner than taking adjoints. To implement ψ * O, we would just rotate O, compute the normal thing, and take care to reverse ψ back and forth in the appropriate way. This is the same way I would do infinite boundary MPS contractions, so I think that could be a viable alternative.

src/algorithms/contractions/window/pepo_1layer.jl
Yue-Zhengyuan: It's flipping the environment space of the east edge. You can take a look at the docstring of `_row_mpo`: the flip make the "physical legs" of the row MPO non-dual, so it acts on a row above it twist-free.

src/algorithms/contractions/window/tools.jl
Yue-Zhengyuan: I have a simple update benchmark with Heisenberg model at beta = 0.5.

For `D = 4`, `chi = 16`, and `S_exchange` on `(1,1) -> (3,3)`:

| Method | Value | Relative error | Median time |
|---|---:|---:|---:|
| Exact | `0.000950771967226` | - | - |
| Zipup only | `0.000948608755141` | `2.28e-3` | `11.32 ms` |
| Zipup + DMRG | `0.000950771965260` | `2.07e-9` | `13.82 ms` |

For even larger bond dimensions / lower temperature the error reduction is more significant: with D = 7, chi = 28 and beta = 2.0,

Method | Value | Relative error
-- | -- | --
Exact | 0.06416644199 | -
Zipup only | 0.06297905212 | 1.85e-2
Zipup + DMRG | 0.06416644470 | 4.23e-8

I think we really need to keep DMRG refinement for accuracy.

Yue-Zhengyuan: Benchmark at larger bond dimensions D = 7, chi = 28 and beta = 2.0:

Method | Value | Relative error
-- | -- | --
Exact | 0.06416644199 | -
Zipup only | 0.06297905212 | 1.85e-2
Zipup + DMRG | 0.06416644470 | 4.23e-8

src/algorithms/contractions/window/twosite/caching.jl
Yue-Zhengyuan: I think we still need to handle the physical legs if we use the rotation approach? South boundary will have dual physical spaces.

src/algorithms/contractions/window/twosite/pepo_1layer.jl
Yue-Zhengyuan: `bonds` is a list of two-site links on which we want to measure two-point correlations. So this is intentional.

src/algorithms/contractions/window/twosite/caching.jl
Yue-Zhengyuan: Then we need to write a new `dot` which does not adjoint the first argument?

src/algorithms/contractions/window/twosite/caching.jl
leburgel: I somehow thought dual physical spaces weren't really an issue for MPSKit, I though we only made hard assumptions on the virtual space duality. I might be wrong or be misremembering, I just thought I tried this out at some point a long time ago.

src/algorithms/contractions/window/twosite/caching.jl
leburgel: For this we could use an edge transfer matrix applied to the identity, and trace the results. Provided the rotations don't just give other issues I didn't think of.

src/algorithms/contractions/window/tools.jl
leburgel: I didn't know this at all, thanks for sharing this!

That definitely invalidates my point of not needing the DMRG refinement. This then makes me wonder about using only DMRG approximation (using the recently added builtin bond expansion).

One slightly nicer way to set up the ZipUp + DMRG refinement would be to use the MPSKit `UnionAlg` machinery to chain algorithms together without defining new types.

But given this accuracy argument, if a ZipUp pass doesn't cost all that much compared to a DMRG approximate iteration, the current solution might be fine for now and we can always change it later.

src/algorithms/correlator_approx.jl
Yue-Zhengyuan: This is a lesson from @lkdvos: with `op::AbstractTensorMap`, we can manually give a clear error message "correlator_approx requires a two-site operator". If we rely on the type annotation check, we will only get a general method error vague to users.

src/algorithms/contractions/window/twosite/caching.jl
Yue-Zhengyuan: Let's continue the discussion at the definition of the function.

src/algorithms/contractions/window/tools.jl
Yue-Zhengyuan: > Also the name here isn't really clear to me. This really is an approximate run, so both window and step are a bit confusing to me. I'd be find with just `_approximate` even if we don't want to shadow MPSKit.approximate, but window can be implied by typing the implementation for finite MPS/MPO, and step seems off because it might sound like you're doing one "step" of an approximate (e.g. one iteration of VOMPS).

For now I have changed the name of the function to `_approximate`. But indeed it is nice if MPSKit already provides the combined ZipUp + DMRG.

In my benchmark, I only run one iteration of DMRG, so the use of ZipUp is providing a good initialization. I'm not sure if merely using DMRG will need more iterations.

Yue-Zhengyuan: One downside of using the name `_approximate` is we are calling an un-underscored function inside an underscored function. It generally should be the other way around.

src/algorithms/contractions/window/twosite/caching.jl
Yue-Zhengyuan: `Patch` and `Window` are almost synonyms to me. I adopted `window` following YASTN. Did you mistype your comment and actually want to say you feel that `window` can be infinite in one dimension?

src/operators/mpo_observable.jl
Yue-Zhengyuan: For `expectation_value`, we may add an additional algorithm struct like `ExactContract` for the existing exact contraction, and dispatch to boundary MPS approximation using the contraction algorithm:
```julia
expectation_value(state, O, env, [alg = ExactContract()])
expectation_value(state, O, env, alg = WindowApprox(zipup, dmrg))
```
So extra names like `expectation_value_approx` will be unnecessary.

`correlator_approx` may still be kept, since it accepts a list of 2-site bonds instead a fixed first site + a list of second sites. (But I may also make the transition?)

I think even in common cases, the MPO is not always first constructed as a dense map? (e.g. some are just a `TensorProductTerm`) And it is not easy either to come up with a algorithm that reorder the sites acted on by the operator so that the string length is minimized.

src/algorithms/contractions/window/tools.jl
Yue-Zhengyuan: Previously `correlator_approx` is using a fixed window for all correlators. Now I take an adaptive depth approach. For example, in a 4x4 patch, with the 1st site at row 1, if the 2nd site is at row 2, then I'll actually be using the upper 2x4 patch to calculate the correlator (the spirit is the same as the liner `correlator`). Then we no longer need to calculate `south-boundary * row-MPO`. I have therefore removed `_adjoint_mpo`, and even the entire `WindowCache` system.

src/algorithms/contractions/window/twosite/caching.jl
Yue-Zhengyuan: `_adjoint_mpo` is now no longer needed, as explained in another conversation.

Yue-Zhengyuan:
Major changes made today (mainly on `correlator_approx`):

- Changed `correlator_approx` to accept one fixed site `i` and second sites `js`, matching `correlator`. Direction is selected automatically from the target positions; square windows prefer row-by-row contraction.
- Introduced adaptive window depth with fixed width: each contraction ends at its target row or column.
- Removed `WindowRowCache`, the reverse sweep that needs `_adjoint_mpo`, and operator swapping.

src/algorithms/contractions/window/tools.jl
Yue-Zhengyuan: South-boundary adjoint is now removed. Basically I now rotate the south boundary by 180 degrees, but without actually rotating it. So sites in the south boundary are ordered from right (east) to left (west).

Yue-Zhengyuan:
Now the only big issue is designing the MPO terms. Ideally this should also include PBC-MPOs that form a closed loop. I'll take a closer look at #425 and see if I can get some inspirations.

src/algorithms/contractions/window/twosite/caching.jl
leburgel: Yes, I meant "patch" sounds finite in both directions, where "window" to me sounds like it could be infinite in one direction.

PR #416, opened 2026-08-15
Don't compute diff unnecessarily in simple update

kshyatt:
Right now this `compare_weights` is being run every iteration, even when completely unused. We can move it down into the `showinfo` block to avoid wasting some time.

### DIFF
--- a/src/algorithms/time_evolution/simpleupdate.jl
+++ b/src/algorithms/time_evolution/simpleupdate.jl
@@ -215,12 +215,12 @@ function MPSKit.time_evolve(
         env0, time0 = it.state.env, time()
         for (psi, env, info) in it
             iter = it.state.iter
-            diff = compare_weights(env0, env)
             stop = (iter == it.nstep)
             showinfo = (check_interval > 0) &&
                 ((iter % check_interval == 0) || (iter == 1) || stop)
             time1 = time()
             if showinfo
+                diff = compare_weights(env0, env)
                 @infov 2 "Space of x-weight at [1, 1] = $(space(env[1, 1, 1], 1))"
                 @infov 2 @sprintf("SU iter %-7d: |Δλ| = %.3e. Time = %.3f s/it", iter, diff, time1 - time0)
             end
--- a/src/algorithms/time_evolution/simpleupdate.jl
+++ b/src/algorithms/time_evolution/simpleupdate.jl
@@ -220,9 +220,8 @@ function MPSKit.time_evolve(
                 ((iter % check_interval == 0) || (iter == 1) || stop)
             time1 = time()
             if showinfo
-                diff = compare_weights(env0, env)
-                @infov 2 "Space of x-weight at [1, 1] = $(space(env[1, 1, 1], 1))"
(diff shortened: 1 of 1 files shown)


src/algorithms/time_evolution/simpleupdate.jl
```diff
@@ -215,12 +215,12 @@ function MPSKit.time_evolve(
...
             iter = it.state.iter
-            diff = compare_weights(env0, env)
             stop = (iter == it.nstep)
             showinfo = (check_interval > 0) &&
                 ((iter % check_interval == 0) || (iter == 1) || stop)
             time1 = time()
             if showinfo
+                diff = compare_weights(env0, env)
                 @infov 2 "Space of x-weight at [1, 1] = $(space(env[1, 1, 1], 1))"
                 @infov 2 @sprintf("SU iter %-7d: |Δλ| = %.3e. Time = %.3f s/it", iter, diff, time1 - time0)
```
lkdvos:
```suggestion
@infov 2 "Space of x-weight at [1, 1] = $(space(env[1, 1, 1], 1))"
@infov 2 @sprintf("SU iter %-7d: |Δλ| = %.3e. Time = %.3f s/it", iter, compare_weights(env0, env), time1 - time0)
```
I think this additionally avoids computing it if verbosity is low enough.

lkdvos (commented):
Nice catch! Small extra suggestion but otherwise happy to merge

src/algorithms/time_evolution/simpleupdate.jl
lkdvos: also this isn't formatted because I don't know how to deal with github

kshyatt: I'll commit and fix it in post

kshyatt:
OK all fixed :)

Yue-Zhengyuan:
My original intention is to check weight difference after each step of SU, so we can stop immediately when it converges, even when the interval to output this information is usually larger. For cases where convergence is not checked (finite-T and real time) it is unnecessary; I just kept weight diff there in case I want to have a look. Even when convergence is checked, I guess we can afford to run a few more steps to reach one full `check_interval`.

kshyatt:
Maybe then we can change the behavior depending on a flag or kwarg?

Yue-Zhengyuan:
How about calculating weight difference after each step only for ground state evolution, and every `check_interval` for other cases?

kshyatt:
Sure, that also works for me. Feel free to push to this branch.

Yue-Zhengyuan:
Wait a few hours before I get back to my laptop

kshyatt:
Yeah of course, no one is about to die from over comparison of weights 😹

Yue-Zhengyuan:
Turns out this is already achieved 😂 The ground state evolution is dispatched to `time_evolve(it, H; ...)` that further needs a Hamiltonian `H`, and it still has the old behavior that weight diff is calculated in each iteration. Other purposes use `time_evolve(it; ...)`, which is already fixed.

Yue-Zhengyuan (approved):

kshyatt:
The GPU stuff is irrelevant here so I'll just merge, then. Thanks everyone! <3

PR #419, opened 2026-08-19
Avoid needlessly recomputing svd_vals for c4v

### LINKED ISSUE #418: For C4V CTMRG, we're computing 7 unneeded `svd_vals`
I checked with @leburgel -- for this symmetry, all 4 corners must be the same. Yet in [`calc_convergence`](https://github.com/QuantumKitHub/PEPSKit.jl/blob/fe38b6b8ac63d345235a20595a96949961263b1d/src/algorithms/ctmrg/ctmrg.jl#L213) we compute `svd_vals` of every corner no matter what. 3 of these could be eliminated, could help a bit for larger tensors/GPU.

kshyatt:
Also switch to using `eigh_vals` in the QR decomposition case, as that will be a lot less expensive than `svd_vals` for the GPU. I think it should still be valid for the check, but I can get rid of that if needed. It's sort of a kludge to fix https://github.com/QuantumKitHub/PEPSKit.jl/issues/418 without reworking all the environments.

### DIFF
--- a/src/algorithms/ctmrg/c4v.jl
+++ b/src/algorithms/ctmrg/c4v.jl
@@ -135,9 +135,24 @@ function check_input(
     return nothing
 end
 
-#
-## C4v-symmetric CTMRG iteration (called through `leading_boundary`)
-#
+# The corners in C₄ᵥ CTMRG are identical and the four edges are related by
+# rotation, so all edges produce the same singular values. `eigh_vals` is about
+# as expensive as `svd_vals` on CPU but much cheaper on GPU, and gives effectively
+# the same information. For the QR projector, off-diagonal elements are still
+# present until CTMRG converges, so we need a fallback for the non-diagonal case.
+corner_spectrum(C::AbstractTensorMap, ::C4vCTMRG) = eigh_vals(C)
+corner_spectrum(C::DiagonalTensorMap, ::C4vCTMRG) = svd_vals(C)
+
+"""
+    convergence_tensors(env::CTMRGEnv, alg::C4vCTMRG) -> (corners, edges)
+
+The corners and edges whose singular values determine convergence.
+
+For C₄ᵥ-symmetric CTMRG, only one corner and one edge are needed since
+the corners are identical, and the edges have identical singular values.
+"""
+convergence_tensors(env::CTMRGEnv, ::C4vCTMRG) =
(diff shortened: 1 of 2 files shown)


src/algorithms/ctmrg/c4v.jl
```diff
@@ -135,9 +135,24 @@ function check_input(
...
-#
-## C4v-symmetric CTMRG iteration (called through `leading_boundary`)
-#
+# The corners in C₄ᵥ CTMRG are identical and the four edges are related by
+# rotation, so all edges produce the same singular values. `eigh_vals` is about
+# as expensive as `svd_vals` on CPU but much cheaper on GPU, and gives effectively
+# the same information. For the QR projector, off-diagonal elements are still
+# present until CTMRG converges, so we need a fallback for the non-diagonal case.
+corner_spectrum(C::AbstractTensorMap, ::C4vCTMRG) = eigh_vals(C)
+corner_spectrum(C::DiagonalTensorMap, ::C4vCTMRG) = svd_vals(C)
```
pbrehmer: Is there a reason why we would want to use `svd_vals` in the `C4vCTMRG` case at all? Shouldn't it be fine to just use the first method that calls `eigh_vals` on any type of `C` (including `DiagonalTensorMap`s).

pbrehmer (commented):
I only have a minor comment but other than that this looks great, thanks!

src/algorithms/ctmrg/c4v.jl
kshyatt: It should I suppose, I just wanted to keep things as consistent as possible with the situation before. The `DiagonalTensorMap` case only arises anyway when using `EighProjector` so we are guaranteed the eigenvalues are all real anyway...

pbrehmer: I'm certainly fine with that but I just thought that this might introduce unnecessary costs with GPUs. But since this PR is already breaking anyway we might just simplify it and get rid of the `svd_vals` here, right? At least that would be my preference.

kshyatt: Will do. For the diagonal case, the GPU also short circuits because it can detect the array is diagonal, so no actual "work" is done

pbrehmer (approved):

PR #420, opened 2026-08-20
Allow toggling of svd algorithm in BP with Defaults

kshyatt:


src/algorithms/bp/gaugefix.jl
lkdvos: Would it make sense to immediately add this as an option here, and then pass this on to the functions below? This would make it actually configurable from the top-level

kshyatt: Could be indeed, though you can also simply override the `Defaults.svd_fwd_alg` in a script? Either is fine for me

lkdvos: I somehow didn't really intend the defaults to be override-able but that might just be a limitation of my imagination

kshyatt: Alright, then let me add it to the algorithm params :)

src/algorithms/bp/gaugefix.jl
```diff
@@ -2,9 +2,30 @@
...
+- `:QRIteration` : MatrixAlgebraKit's [`QRIteration`](@extref MatrixAlgebraKit.QRIteration)
+- `:Bisection` : MatrixAlgebraKit's [`Bisection`](@extref MatrixAlgebraKit.Bisection)
+- `:Jacobi` : MatrixAlgebraKit's [`Jacobi`](@extref MatrixAlgebraKit.Jacobi)
+- `:SVDViaPolar` : MatrixAlgebraKit's [`SVDViaPolar`](@extref MatrixAlgebraKit.SVDViaPolar)
+- `:SafeDivideAndConquer` : MatrixAlgebraKit's [`SafeDivideAndConquer`](@extref MatrixAlgebraKit.SafeDivideAndConquer)
 """
-@kwdef struct BPGauge
-    # TODO: add options
+@kwdef struct BPGauge{F}
+    svd_fwd_alg::F = Defaults.svd_fwd_alg
```
lkdvos, line 28:
```suggestion
    svd_alg::F = Defaults.svd_fwd_alg
```

I don't think there is a great reason to mention that it is a forward algorithm here, and not an AD one? (also update docstring)

lkdvos (approved):

PR #421, opened 2026-08-25
Ensure dynamic tolerance scaling works for every gradient algorithm

leburgel:
Missed in #413 after #410 was merged, but the fact that the custom `MPSKit._updatetol` implementation was specialized to `FixedPointGradient` meant that the new default configuration for an optimization using `fixedpoint` was broken when using an `ImplicitGradient`.

I added such a run to the example tests, to test `ImplicitGradient` via the public interface, and not just through raw gradient accuracy tests.

UPDATE: This exposed some issues with the corner type of the environment returned by `leading_boundary` when using implicit differentiation. In this case, we return the corner as a complex non-diagonal tensor map before evaluating the energy density, since the implicit differentiation approach relies on allowing complex non-diagonal adjoints of the corner. When setting `reuse_env = true`, the `update!` runs into a type mismatch between the initial environment (which will have a real diagonal corner when using a `C4vEighProjector`) and the corner to be updated (complex non-diagonal when used to evaluate the energy). Instead of converting the initial environment to have a complex non-diagonal corner, I opted to convert the corner type of the updated environment to a real diagonal if necessary.

### DIFF
--- a/src/algorithms/optimization/implicit_differentiation.jl
+++ b/src/algorithms/optimization/implicit_differentiation.jl
@@ -55,6 +55,10 @@ function GradientAlgorithm(;
     return alg_type(solver)
 end
 
+# a `GradientAlgorithm` has no top-level `tol` field (it lives on `solver_alg`), so the
+# default `MPSKit.DynamicTols._updatetol` (which sets `alg.tol`) doesn't apply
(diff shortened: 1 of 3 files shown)


pbrehmer (approved):
Looks good to me!

PR #422, opened 2026-08-27
Subspace-Iteration CTMRG

Yue-Zhengyuan:
This PR adds one more CTMRG projector algorithm `SubspaceIterationProjector`, introduced in https://arxiv.org/abs/2607.15158.

The basic idea is that, after constructing the full-infinite environment `M`, one use QR/LQ to obtain "rangefinder" tensors `X, Y` to project it to `Y M X` which lives in a much smaller "oversampling" space. This space is intentionally chosen to be still larger than the target environment space χ. It then greatly reduces the cost to do SVD, resulting in significant speed up especially on GPU. 

According to the paper, one can achieve over 100x speedup for D = 6, χ = 144 on GPU. So it appears to be quite a worthwhile addition.

The rangefinders `X`, `Y` may be reused during CTMRG iterations, thus a caching mechanism has to be introduced. 

## TODO

Since `SubspaceIterationProjector` contains parameters beyond `trunc`, it is a bit difficult to just specify it with a symbol. We need to introduce more API flexibility in defining an CTMRG / projector algorithm.

The initial implementation is restricted to fixed-space truncation, trivial symmetry sectors (it is not straightforward to efficiently determine the charges in the oversampling space) and non-differentiated forward CTMRG contractions. Missing features will be gradually added back in follow-up commits.


leburgel:
I don't really have time to go over this in detail right now, but we should consider the difference between this approach and that of #263, and consider if they are related. Switching to a randomized SVD to compute the same projectors we already support could be a bit cleaner in terms of implementation, so it might be worth it to work this out a bit more before committing here.

Issue #414, opened 2026-08-10
Interface: Pack `(bra, ket)` into one tuple for `expectation_value`


Yue-Zhengyuan:
Currently, in `expectation_value` and `reduced_densitymatrix`:

- For iPEPS measurement, we can provide either a single `ket::InfinitePEPS`, or `bra::InfinitePEPS` and `ket::InfinitePEPS` as two separate arguments.
- For purified iPEPO, we always provide `bra::InfinitePEPO` and `ket::InfinitePEPO`.
- For one-layer iPEPO, we provide only one `ρ::InfinitePEPO`.

So we have either one or two arguments to specify the states. Do we want to unify the interface to one state argument by packing `(bra, ket)` into one tuple?

PR #423, opened 2026-08-31
Refactor `expectation_value` implementations

leburgel:
Some refactoring and reorganization related to the evaluation of expectation values. The main motivation for this refactor is making it easier to add support for computing expectation values of different kinds of operators, and evaluate these using different kinds of environments. Some examples of this are given in #425 and #426, which are built on top of this refactor. 

Short summary:
- Moved around some existing code to different/new files, hopefully improving the organization a bit.
- Remove as many type annotations as possible from the methods used along the evaluation of expectation values, allowing to reuse the main entry points for different types of operators and environments.
- Insert a dedicated `local_expectation_value` method to catch local expectation values for different kinds of operators. This goes through the `reduced_densitymatrix` approach for `AbstractTensorMap` local terms, but can follow a different path for different kinds of local terms (see for example #425).
- Implement `BPEnv` expectation values via reduced density matrix contraction along the generic `AbstractTensorMap` path, by specializing the appropriate methods.
- Write the main local patch contraction as generic methods in terms of a fixed set of contraction expression generators . These generators then dispatch on the appropriate type to generate an appropriate expression. The idea would be that to add support support for a different type of state, operator or environment, we now only need to overload one of the main expression generators `boundary_contraction_expr`, `bulk_contraction_expr` or `operator_contraction_expr`.


Yue-Zhengyuan:
I'll try coming up with a source tree of expectation-value-related functions to better discuss where each function should be placed.

src/utility/util.jl
```diff
@@ -62,6 +62,66 @@ function absorb_s(U::AbstractTensorMap, S::DiagonalTensorMap, V::AbstractTensorM
...
+"""
+    absorb_left(
+        E::AbstractTensorMap{<:Any, S}, C::AbstractTensorMap{<:Any, S, 1, 1}
+    ) where {S}
+
+Absorb a matrix `C` into the left of a tensor `E` by contracting the first leg of `E`
+with the second leg of `C`.
+"""
+function absorb_left(
+        A::AbstractTensorMap{<:Any, S}, C::AbstractTensorMap{<:Any, S, 1, 1}
```
Yue-Zhengyuan, line 74: Argument name should better match the docstring.

src/utility/util.jl
```diff
@@ -62,6 +62,66 @@ function absorb_s(U::AbstractTensorMap, S::DiagonalTensorMap, V::AbstractTensorM
...
+    return E * twistdual(C, 1)
+end
+
+"""
+    absorb_left_right(
+        A::AbstractTensorMap{<:Any, S}, C::AbstractTensorMap{<:Any, S, 1, 1}
+    ) where {S}
+
+Absorb a matrix `C` into the right of a tensor `A` by contracting the last leg of `A`
+with the first leg of `C`.
```
Yue-Zhengyuan: Docstring is outdated. I suppose you copied it but forgot to update it.

src/algorithms/contractions/localoperator.jl
```diff
@@ -122,270 +250,221 @@ function _contract_state_expr(rowrange, colrange, height, cartesian_inds = nothi
...
+# --------------------------------------------------------------------
+
+"""
+$(SIGNATURES)
+
+Contract the rectangular patch spanned by `inds` with `operator` inserted on those sites,
+leaving no physical leg open, and return the resulting scalar.
+
+The sites are carried as `Val` parameters so that the patch geometry is available while the
+contraction is generated, and `state` is the tuple of layers making up the sandwich. The
```
Yue-Zhengyuan: "`state` is the tuple of layers" is not quite precise, since in principle  it can also be just a single multi-layer iPEPO.

src/algorithms/expectation_value.jl
```diff
@@ -0,0 +1,319 @@
...
+
+
+# Expectation value of a local partition function tensor
+# ------------------------------------------------------
+
+"""
+    expectation_value(pf::InfinitePartitionFunction, inds => O, env::CTMRGEnv)
+
+Compute the expectation value corresponding to inserting a local tensor(s) `O` at
+position `inds` in the partition function `pf` and contracting the chole using a given CTMRG
```
Yue-Zhengyuan: Typo: "chole" should be "whole".

src/algorithms/expectation_value.jl
```diff
@@ -0,0 +1,319 @@
...
+    end
+end
+network_value(state, env::CTMRGEnv) = network_value(InfiniteSquareNetwork(state), env)
+
+"""
+    _contract_site(ind::Tuple{Int,Int}, network::InfiniteSquareNetwork, env::CTMRGEnv)
+
+Contract around a single site `ind` of a square network using a given CTMRG environment.
+"""
+function _contract_site(ind::Tuple{Int, Int}, network, env::CTMRGEnv)
```
Yue-Zhengyuan: Maybe we should match docstring and annotate `network::InfiniteSquareNetwork`.

src/algorithms/expectation_value.jl
```diff
@@ -0,0 +1,319 @@
...
+Compute the contribution of a single term of a [`LocalOperator`](@ref) to the expectation
+value ⟨bra|O|ket⟩ / ⟨bra|ket⟩ or tr(O * state) / tr(state), where `operator` is the local
+term acting on the sites `inds`.
+
+The implementation is overloaded based on the type of operator to be evaluated
+"""
+function local_expectation_value end
+
+# AbstractTensorMap evaluation goes through reduced density matrix
+function local_expectation_value(inds, bra, operator::AbstractTensorMap, ket, env)
```
Yue-Zhengyuan, line 48: I see you are adopting my suggestion in #414 by packing `(bra, ket)` in a tuple. But only the docstring follows the new signature. The source code is still using the old syntax.

src/algorithms/expectation_value.jl
```diff
@@ -0,0 +1,319 @@
...
+@doc """
+    reduced_densitymatrix(inds, ket::InfinitePEPS, bra::InfinitePEPS = ket, env)
+    reduced_densitymatrix(inds, ket::InfinitePEPO, bra::InfinitePEPO, env)
+    reduced_densitymatrix(inds, state::InfinitePEPO, env)
+
+Construct the reduced density matrix `ρ` of `|ket⟩⟨bra|`, where both `ket` and `bra`
+correspond to either a PEPS or a PEPO representing a PEPS with ancillary legs.
+Alternatively, construct the reduced density matrix `ρ` of a mixed state specified by the 
+density matrix PEPO `state`. The reduced density matrix is contracted around the open
+indices `inds` using the environment `env`, and is normalized such that `tr(ρ) = 1`.
```
Yue-Zhengyuan: While the normalization should in theory be `tr(ρ) = 1`, in the code we are actually using `str(ρ) = 1`. We have some previous discussion on this (#283), and it is better to thoroughly clarify it this time to avoid more future confusions.

src/algorithms/expectation_value.jl
```diff
@@ -0,0 +1,319 @@
...
+)
+Base.@deprecate(
+    reduced_densitymatrix(
+        inds::NTuple{N, Tuple{Int, Int}}, args...
+    ) where {N},
+    reduced_densitymatrix(collect(CartesianIndex.(inds)), args...)
+)
+
+"""
+    contract_local_operator(inds, O, ket::InfinitePEPS, bra::InfinitePEPS = ket, env)
```
Yue-Zhengyuan, line 126: There are actually no overloads yet that omits `bra`. Also, do we want to adapt a new breaking `(bra, ket)` interface?

src/algorithms/expectation_value.jl
```diff
@@ -0,0 +1,319 @@
...
+
+Base.@deprecate(
+    contract_local_operator(
+        inds::NTuple, args...
+    ),
+    contract_local_operator(collect(inds), args...)
+)
+
+"""
+    contract_local_norm(inds, ket::InfinitePEPS, bra::InfinitePEPS = ket, env)
```
Yue-Zhengyuan, line 170: Same issue as `contract_local_operator`: lack of overload without `bra`, and whether we want to adapt a new `(bra, ket)` interface.

src/algorithms/expectation_value.jl
```diff
@@ -0,0 +1,319 @@
...
+    contract_local_norm(inds, ket::InfinitePEPO, bra::InfinitePEPO, env)
+    contract_local_norm(inds, state::InfinitePEPO, env)
+
+Contract a local norm corresponding to the overlap `ket` and `bra` states, computing a patch
+of `⟨bra|ket⟩`, where `ket` and `bra` correspond to either a PEPS or a PEPO representing a
+PEPS with ancillary legs. Alternatively, contract a local norm patch of a density matrix
+PEPO `state`, computing a patch of `tr(state)`. The norm patch is contracted around the open
+indices `inds` using the environment `env`.
```
Yue-Zhengyuan: Let's be more specific and say *virtual* indices (of the patch).

src/algorithms/expectation_value.jl
```diff
@@ -0,0 +1,319 @@
...
+
+"""
+    contract_local_norm(inds, ket::InfinitePEPS, bra::InfinitePEPS = ket, env)
+    contract_local_norm(inds, ket::InfinitePEPO, bra::InfinitePEPO, env)
+    contract_local_norm(inds, state::InfinitePEPO, env)
+
+Contract a local norm corresponding to the overlap `ket` and `bra` states, computing a patch
+of `⟨bra|ket⟩`, where `ket` and `bra` correspond to either a PEPS or a PEPO representing a
+PEPS with ancillary legs. Alternatively, contract a local norm patch of a density matrix
+PEPO `state`, computing a patch of `tr(state)`. The norm patch is contracted around the open
```
Yue-Zhengyuan: I'm not sure if `tr` is the correct unambiguous terminology. It gives the feeling that the trace is taken without fermion twists.

What if we just do not mention the two different internal implementations (direct contraction vs density matrix)?

Yue-Zhengyuan (commented):
First-round comments focusing on docstring inconsistencies.

src/algorithms/contractions/localoperator.jl
```diff
@@ -10,35 +15,141 @@ physicallabel(args...) = tensorlabel(:d, args...)
...
+
+"""
+$(SIGNATURES)
+
+Wrap an assembled product in `@autoopt @tensor`, with `lhs` on the left if given and as a
+scalar otherwise, emitting `contractcheck = true` only when `check` is set.
+"""
+function _tensor_expr(prod, lhs, check::Bool)
+    return if isnothing(lhs)
+        check ? :(@autoopt @tensor contractcheck = true $prod) : :(@autoopt @tensor $prod)
```
Yue-Zhengyuan: Is it possible to write something like `@autoopt @tensor contractcheck = $check $prod` ?

Yue-Zhengyuan:
Some files put too many things together in my opinion. I would like to suggest the following source tree:

<details>

```julia
src/
└── algorithms/
    ├── expectation_value.jl
    │   ├── local_expectation_value(inds, bra, operator::AbstractTensorMap, ket, env)
    │   │   local_expectation_value(inds, state, operator::AbstractTensorMap, env)
    │   │   Select the evaluation strategy by operator type; evaluate the current dense-operator implementation through a reduced density matrix.
    │   │
    │   └── MPSKit.expectation_value(bra::S, O::LocalOperator, ket::S, env) where {S<:InfiniteState}
    │       MPSKit.expectation_value(peps::InfinitePEPS, O::LocalOperator, env)
    │       MPSKit.expectation_value(state::InfinitePEPO, O::LocalOperator, env)
    │       MPSKit.expectation_value(pf::InfinitePartitionFunction, op::Pair{CartesianIndex{2},<:AbstractTensorMap{T,S,2,2}}, env) where {T,S}
    │       MPSKit.expectation_value(pf::InfinitePartitionFunction, op::Pair{Tuple{Int,Int}}, env)
    │       Delegate each LocalOperator term to local_expectation_value and sum the results, or evaluate a local tensor insertion in a partition function.
    │
    ├── reduced_densitymatrix.jl
    │   └── reduced_densitymatrix(inds::Vector{CartesianIndex{2}}, ket::InfinitePEPS, bra::InfinitePEPS, env)
    │       reduced_densitymatrix(inds, ket::InfinitePEPS, env)
    │       reduced_densitymatrix(inds::Vector{CartesianIndex{2}}, state::InfinitePEPO, env)
    │       reduced_densitymatrix(inds::Vector{CartesianIndex{2}}, ket::InfinitePEPO, bra::InfinitePEPO, env)
    │       reduced_densitymatrix(inds::NTuple{N,CartesianIndex{2}}, args...) where {N} [deprecated]
    │       reduced_densitymatrix(inds::NTuple{N,Tuple{Int,Int}}, args...) where {N} [deprecated]
    │       Select an optimized density-matrix contraction when available, otherwise construct a normalized density matrix with _contract_densitymatrix.
    │
    ├── contractions/
    │   ├── tensor_traces.jl
    │   │   ├── _str(::Bosonic, t::AbstractTensorMap)
    │   │   │   _str(::Fermionic, t::AbstractTensorMap{<:Any,<:Any,N,N}) where {N}
    │   │   │   Dispatch the trace calculation on bosonic or fermionic braiding style.
    │   │   │
    │   │   ├── str(t::AbstractTensorMap)
    │   │   │   Compute the ordinary trace for bosonic tensors and the supertrace for fermionic tensors.
    │   │   │
    │   │   └── trmul(H::AbstractTensorMap{<:Any,S,N,N}, ρ::AbstractTensorMap{<:Any,S,N,N}) where {S,N}
    │   │       Contract an operator with a density matrix without materializing their product.
    │   │
    │   ├── absorb.jl
    │   │   ├── absorb_left(A::AbstractTensorMap{<:Any,S}, C::AbstractTensorMap{<:Any,S,1,1}) where {S}
    │   │   │   absorb_left(P::AbstractTensorMap{<:Any,S,1,N}, C::AbstractTensorMap{<:Any,S,1,1}) where {S,N}
    │   │   │   Contract a matrix into the leftmost leg of a tensor.
    │   │   │
    │   │   ├── absorb_right(A::AbstractTensorMap{<:Any,S}, C::AbstractTensorMap{<:Any,S,1,1}) where {S}
    │   │   │   absorb_right(E::AbstractTensorMap{<:Any,S,N,1}, C::AbstractTensorMap{<:Any,S,1,1}) where {S,N}
    │   │   │   Contract a matrix into the rightmost leg of a tensor.
    │   │   │
    │   │   └── absorb_left_right(T::AbstractTensorMap{<:Any,S}, CL::AbstractTensorMap{<:Any,S,1,1}, CR::AbstractTensorMap{<:Any,S,1,1}) where {S}
    │   │       Contract matrices into both outer legs of a tensor.
    │   │
    │   ├── bp_messages.jl
    │   │   ├── absorb_north_message(A::PEPSTensor, M::PEPSMessage)
    │   │   │   Absorb a north BP message into a PEPS tensor.
    │   │   │
    │   │   ├── absorb_east_message(A::PEPSTensor, M::PEPSMessage)
    │   │   │   Absorb an east BP message into a PEPS tensor.
    │   │   │
    │   │   ├── absorb_south_message(A::PEPSTensor, M::PEPSMessage)
    │   │   │   Absorb a south BP message into a PEPS tensor.
    │   │   │
    │   │   ├── absorb_west_message(A::PEPSTensor, M::PEPSMessage)
    │   │   │   Absorb a west BP message into a PEPS tensor.
    │   │   │
    │   │   ├── contract_north_message(A::PEPSSandwich, M_west::PEPSMessage, M_north::PEPSMessage, M_east::PEPSMessage)
    │   │   │   Update the BP message leaving a site toward the north.
    │   │   │
    │   │   ├── contract_east_message(A::PEPSSandwich, M_north::PEPSMessage, M_east::PEPSMessage, M_south::PEPSMessage)
    │   │   │   Update the BP message leaving a site toward the east.
    │   │   │
    │   │   ├── contract_south_message(A::PEPSSandwich, M_east::PEPSMessage, M_south::PEPSMessage, M_west::PEPSMessage)
    │   │   │   Update the BP message leaving a site toward the south.
    │   │   │
    │   │   └── contract_west_message(A::PEPSSandwich, M_south::PEPSMessage, M_west::PEPSMessage, M_north::PEPSMessage)
    │   │       Update the BP message leaving a site toward the west.
    │   │
    │   ├── local_patch/
    │   │   ├── expr_utils.jl
    │   │   │   ├── tensorlabel(args...)
    │   │   │   │   Join label components into a tensor-contraction symbol.
    │   │   │   │
    │   │   │   ├── envlabel(args...)
    │   │   │   │   Construct an environment-leg label.
    │   │   │   │
    │   │   │   ├── virtuallabel(args...)
    │   │   │   │   Construct a state virtual-leg label.
    │   │   │   │
    │   │   │   ├── physicallabel(args...)
    │   │   │   │   Construct a physical-leg label.
    │   │   │   │
    │   │   │   ├── _open_slot(open, r, c)
    │   │   │   │   Return the operator slot assigned to a patch site.
    │   │   │   │
    │   │   │   ├── _bulk_virtuallabels(i, j, layer, gridsize)
    │   │   │   │   Return the north, east, south, and west labels of one bulk tensor.
    │   │   │   │
    │   │   │   ├── _patch_inds(inds::Type)
    │   │   │   │   _patch_inds(inds::Tuple{Vararg{Val}})
    │   │   │   │   Recover and validate patch coordinates encoded in Val parameters.
    │   │   │   │
    │   │   │   ├── _patch_ranges(sites)
    │   │   │   │   Return the row and column ranges spanned by patch sites.
    │   │   │   │
    │   │   │   ├── _patch_gridsize(sites)
    │   │   │   │   Return the rectangular dimensions spanned by patch sites.
    │   │   │   │
    │   │   │   └── _patch_shape_string(sites)
    │   │   │       Format the rectangular patch dimensions for diagnostics.
    │   │   │
    │   │   ├── network_expr.jl
    │   │   │   ├── contractcheck(state::Type)
    │   │   │   │   contractcheck(::Type{<:InfinitePEPO})
    │   │   │   │   contractcheck(::Type{<:Tuple{Vararg{InfinitePEPO}}})
    │   │   │   │   Select whether tensor-space compatibility is checked for a state representation.
    │   │   │   │
    │   │   │   ├── _tensor_expr(prod, lhs, check::Bool)
    │   │   │   │   Assemble indexed tensor expressions into an optimized `@tensor` contraction.
    │   │   │   │
    │   │   │   ├── boundary_contraction_expr(env::Type, rowrange, colrange)
    │   │   │   │   boundary_contraction_expr(::Type{<:CTMRGEnv{C,T}}, rowrange, colrange) where {C,T}
    │   │   │   │   Generate indexed CTMRG boundary tensor expressions for an `@tensor` contraction, or report that an environment implementation is unavailable.
    │   │   │   │
    │   │   │   ├── bulk_contraction_expr(state::Type, rowrange, colrange, open)
    │   │   │   │   bulk_contraction_expr(::Type{<:Tuple{InfinitePEPS,InfinitePEPS}}, rowrange, colrange, open)
    │   │   │   │   bulk_contraction_expr(::Type{<:InfinitePEPO}, rowrange, colrange, open)
    │   │   │   │   bulk_contraction_expr(::Type{<:Tuple{InfinitePEPO,InfinitePEPO}}, rowrange, colrange, open)
    │   │   │   │   bulk_contraction_expr(state::Type{<:Tuple{Vararg{InfinitePEPO}}}, rowrange, colrange, open)
    │   │   │   │   Generate indexed PEPS or PEPO tensor expressions for the patch interior, or report that a state implementation is unavailable.
    │   │   │   │
    │   │   │   └── operator_contraction_expr(operator::Type, nsites)
    │   │   │       operator_contraction_expr(::Type{<:AbstractTensorMap}, nsites)
    │   │   │       Generate an indexed dense-operator expression for an `@tensor` contraction, or report that an operator implementation is unavailable.
    │   │   │
    │   │   ├── patch_contractions.jl
    │   │   │   ├── _contract_local_operator(inds::NTuple{N,Val}, operator, state, env) where {N}
    │   │   │   │   Assemble and execute the unnormalized operator-inserted patch contraction for direct evaluation.
    │   │   │   │
    │   │   │   ├── contract_local_operator(inds::Vector{CartesianIndex{2}}, O, ket::InfinitePEPS, bra::InfinitePEPS, env)
    │   │   │   │   contract_local_operator(inds::Vector{CartesianIndex{2}}, O, state::InfinitePEPO, env)
    │   │   │   │   contract_local_operator(inds::Vector{CartesianIndex{2}}, O, ket::InfinitePEPO, bra::InfinitePEPO, env)
    │   │   │   │   contract_local_operator(inds::Vector{Tuple{Int,Int}}, O, args...)
    │   │   │   │   contract_local_operator(inds::NTuple, args...) [deprecated]
    │   │   │   │   Validate the inputs and invoke the generated direct contraction for an operator insertion.
    │   │   │   │
    │   │   │   ├── _contract_local_norm(inds::NTuple{N,Val}, state, env) where {N}
    │   │   │   │   Assemble and execute the corresponding patch-normalization contraction with closed physical legs.
    │   │   │   │
    │   │   │   └── contract_local_norm(inds::Vector{CartesianIndex{2}}, ket::InfinitePEPS, bra::InfinitePEPS, env)
    │   │   │       contract_local_norm(inds::Vector{CartesianIndex{2}}, state::InfinitePEPO, env)
    │   │   │       contract_local_norm(inds::Vector{CartesianIndex{2}}, ket::InfinitePEPO, bra::InfinitePEPO, env)
    │   │   │       contract_local_norm(inds::Vector{Tuple{Int,Int}}, args...)
    │   │   │       contract_local_norm(inds::NTuple, ket::InfinitePEPS, bra::InfinitePEPS, env) [deprecated]
    │   │   │       Validate the inputs and invoke the generated direct contraction used to normalize an operator insertion.
    │   │   │
    │   │   └── densitymatrix/
    │   │       ├── generic.jl
    │   │       │   └── _contract_densitymatrix(inds::NTuple{N,Val}, state, env) where {N}
    │   │       │       Assemble and execute a generated patch contraction with open physical legs.
    │   │       │
    │   │       ├── ctmrg.jl
    │   │       │   ├── reduced_densitymatrix1x1(ind::CartesianIndex{2}, ket::InfinitePEPS, bra::InfinitePEPS, env::CTMRGEnv)
    │   │       │   │   Contract an optimized one-site PEPS density matrix using a CTMRG environment.
    │   │       │   │
    │   │       │   ├── reduced_densitymatrix2x1(ind::CartesianIndex, ket::InfinitePEPS, bra::InfinitePEPS, env::CTMRGEnv)
    │   │       │   │   Contract an optimized vertical two-site PEPS density matrix using a CTMRG environment.
    │   │       │   │
    │   │       │   └── reduced_densitymatrix1x2(ind::CartesianIndex, ket::InfinitePEPS, bra::InfinitePEPS, env::CTMRGEnv)
    │   │       │       Contract an optimized horizontal two-site PEPS density matrix using a CTMRG environment.
    │   │       │
    │   │       └── bp.jl
    │   │           ├── _contract_densitymatrix(inds::NTuple{N,Val}, state, env::BPEnv) where {N}
    │   │           │   Reject patch geometries that cannot be represented by local BP messages.
    │   │           │
    │   │           ├── reduced_densitymatrix1x1(ind::CartesianIndex{2}, ket::InfinitePEPS, bra::InfinitePEPS, env::BPEnv)
    │   │           │   Contract a one-site PEPS density matrix using BP messages.
    │   │           │
    │   │           ├── reduced_densitymatrix2x1(coord::CartesianIndex{2}, ket::InfinitePEPS, bra::InfinitePEPS, env::BPEnv)
    │   │           │   Contract a vertical two-site PEPS density matrix using BP messages.
    │   │           │
    │   │           └── reduced_densitymatrix1x2(coord::CartesianIndex{2}, ket::InfinitePEPS, bra::InfinitePEPS, env::BPEnv)
    │   │               Contract a horizontal two-site PEPS density matrix using BP messages.
    │   │
    │   └── ctmrg/
    │       └── network_value.jl
    │           ├── _contract_site(ind::Tuple{Int,Int}, network, env::CTMRGEnv)
    │           │   _contract_site(C_northwest, C_northeast, C_southeast, C_southwest, E_north::CTMRG_PEPS_EdgeTensor, E_east::CTMRG_PEPS_EdgeTensor, E_south::CTMRG_PEPS_EdgeTensor, E_west::CTMRG_PEPS_EdgeTensor, O::PEPSSandwich)
    │           │   _contract_site(C_northwest, C_northeast, C_southeast, C_southwest, E_north::CTMRG_PF_EdgeTensor, E_east::CTMRG_PF_EdgeTensor, E_south::CTMRG_PF_EdgeTensor, E_west::CTMRG_PF_EdgeTensor, O::PFTensor)
    │           │   _contract_site(C_northwest, C_northeast, C_southeast, C_southwest, E_north::TE, E_east::TE, E_south::TE, E_west::TE, O::PEPOSandwich{H}) where {TE<:CTMRGEdgeTensor,H}
    │           │   Contract one network site with its surrounding CTMRG corners and edges.
    │           │
    │           ├── _contract_corners(ind::Tuple{Int,Int}, env::CTMRGEnv)
    │           │   _contract_corners(C_northwest::CTMRGCornerTensor, C_northeast::CTMRGCornerTensor, C_southeast::CTMRGCornerTensor, C_southwest::CTMRGCornerTensor)
    │           │   Contract four CTMRG corners for network-value normalization.
    │           │
    │           ├── _contract_vertical_edges(ind::Tuple{Int,Int}, env::CTMRGEnv)
    │           │   _contract_vertical_edges(C_northwest::CTMRGCornerTensor, C_northeast::CTMRGCornerTensor, C_southeast::CTMRGCornerTensor, C_southwest::CTMRGCornerTensor, E_east::CTMRGEdgeTensor{T,S,N}, E_west::CTMRGEdgeTensor{T,S,N}) where {T,S,N}
    │           │   Contract two vertical edges and four corners for network-value normalization.
    │           │
    │           ├── _contract_horizontal_edges(ind::Tuple{Int,Int}, env::CTMRGEnv)
    │           │   _contract_horizontal_edges(C_northwest::CTMRGCornerTensor, C_northeast::CTMRGCornerTensor, C_southeast::CTMRGCornerTensor, C_southwest::CTMRGCornerTensor, E_north::CTMRGEdgeTensor{T,S,N}, E_south::CTMRGEdgeTensor{T,S,N}) where {T,S,N}
    │           │   Contract two horizontal edges and four corners for network-value normalization.
    │           │
    │           ├── network_value(network::InfiniteSquareNetwork, env::CTMRGEnv)
    │           │   network_value(state, env::CTMRGEnv)
    │           │   Evaluate the normalized value per unit cell of an infinite square network.
    │           │
    │           └── LinearAlgebra.norm(peps::InfinitePEPS, env::CTMRGEnv)
    │               Evaluate the CTMRG norm of an infinite PEPS.
    │
    └── optimization/
        └── peps_optimization.jl
            └── cost_function(peps::InfinitePEPS, env, O::LocalOperator)
                Return the real part of an expectation value as the PEPS optimization cost.
```

</details>

src/algorithms/expectation_value.jl
leburgel: I did try this at some point but got myself tied in a knot, so I decided to revert that for now. I just forgot to revert the docstring. I planned to revisit this once #414 was settled.

Regarding your suggestion itself, I think I agree for the lower-level methods `local_expectation_value`, `reduced_densitymatrix`, `contract_local_operator` and `contract_local_norm`, there it would save quite a bit of argument assembling that's never used. I do like `expectation_value(bra, O, ket = bra)` though. It's just a bit unfortunate that `reduced_densitymatrix` is exported, making this breaking over over something that doesn't give much actual benefits.

src/algorithms/expectation_value.jl
leburgel: Yes, at this level I think packing the arguments as `(bra, ket)` makes sense, the argument juggling on these methods is quite annoying and is never actually used. My only concern is whether it's really worth breaking `reduced_densitymatrix` over this.

src/algorithms/expectation_value.jl
leburgel: I would be fine with not mentioning the different internal implementations only if we merge the calling signatures in terms of a single `state` variable. Now it's kind of impossible to explain the available calling signatures without making a distinction between the internal implementations.

src/algorithms/contractions/localoperator.jl
leburgel: I think it is possible, but I wanted to reproduce the previous expressions exactly. So instead of adding a `contractcheck = false` in some places, I chose to just omit it from the generated expression altogether.

src/algorithms/contractions/bp_contractions.jl
```diff
@@ -48,104 +48,44 @@ absorb_south_message(A::PEPSTensor, M::PEPSMessage) =
...
-        inds::Vector{CartesianIndex{2}},
-        O::AbstractTensorMap,
-        ket::InfinitePEPS,
-        bra::InfinitePEPS,
-        env::BPEnv,
+# Belief Propagation reduced density matrices
+# -------------------------------------------
+# BP messages live on the bonds of the network, so there is no analogue of the generated
+# rectangular-patch contraction used for a `CTMRGEnv`: only single sites and nearest neighbor
+# pairs can be contracted without introducing loop corrections.
```
lkdvos: In principle that is not really true, you could do the exact same thing as with the CTMRG environments without the edges being connected, although that approximation tends to be too large to be very useful which is one of the reasons why it isn't defined here. I'm fine with leaving this comment if you want though

src/algorithms/contractions/localoperator.jl
```diff
@@ -10,35 +15,141 @@ physicallabel(args...) = tensorlabel(:d, args...)
...
+"""
+$(SIGNATURES)
+
+Check whether the `@tensor` contraction for a given state type should be emitted with
+`contractcheck = true`. Enabled for PEPO sandwiches, disabled for other state types,
+in order to preserve the legacy behavior for different state types.
+"""
+contractcheck(state::Type) = false
+contractcheck(::Type{<:InfinitePEPO}) = true
+contractcheck(::Type{<:Tuple{Vararg{InfinitePEPO}}}) = true
```
lkdvos: Is this "legacy behavior" not just test code that crept into the released code? I don't think there is a great reason to keep this enabled?

src/algorithms/contractions/bp_contractions.jl
leburgel: I'll update the comment.

src/algorithms/contractions/localoperator.jl
leburgel: I think it is, but I didn't just want to throw anything out. If we no longer need it, we can just remove it. On the other hand, if it doesn't cost anything, there's not much harm in leaving it even if it's not needed. That was my reasoning for putting this in. But things just run now, so the error check will just literally never do anything unless we change the contraction code.

I'm fine with removing this altogether if that's better.

lkdvos: It does cost a little bit of runtime, although I am guessing that is negligible, but also some compile time, but mostly code to maintain, so I'd be happy to just remove that altogether.

leburgel:
> Some files put too many things together in my opinion. I would like to suggest the following source tree:
> ...

I did the reorganization mostly as suggested:

<details>
<summary>Updated file tree</summary>

```
src/
├── algorithms/
│   ├── expectation_value/
│   │   ├── expectation_value.jl
│   │   │   ├── local_expectation_value(inds, bra, operator::AbstractTensorMap, ket, env)
│   │   │   │   local_expectation_value(inds, state, operator::AbstractTensorMap, env)
│   │   │   │   Select the evaluation strategy by operator type; evaluate the current dense-operator implementation through a reduced density matrix.
│   │   │   │
│   │   │   └── MPSKit.expectation_value(bra::S, O::LocalOperator, ket::S, env) where {S<:InfiniteState}
│   │   │       MPSKit.expectation_value(peps::InfinitePEPS, O::LocalOperator, env)
│   │   │       MPSKit.expectation_value(state::InfinitePEPO, O::LocalOperator, env)
│   │   │       MPSKit.expectation_value(pf::InfinitePartitionFunction, op::Pair{CartesianIndex{2},<:AbstractTensorMap{T,S,2,2}}, env) where {T,S}
│   │   │       MPSKit.expectation_value(pf::InfinitePartitionFunction, op::Pair{Tuple{Int,Int}}, env)
│   │   │       Delegate each LocalOperator term to local_expectation_value and sum the results, or evaluate a local tensor insertion in a partition function.
│   │   │
│   │   ├── reduced_densitymatrix.jl
│   │   │   ├── reduced_densitymatrix(inds::Vector{CartesianIndex{2}}, ket::InfinitePEPS, bra::InfinitePEPS, env)
│   │   │   │   reduced_densitymatrix(inds, ket::InfinitePEPS, env)
│   │   │   │   reduced_densitymatrix(inds::Vector{CartesianIndex{2}}, state::InfinitePEPO, env)
│   │   │   │   reduced_densitymatrix(inds::Vector{CartesianIndex{2}}, ket::InfinitePEPO, bra::InfinitePEPO, env)
│   │   │   │   reduced_densitymatrix(inds::NTuple{N,CartesianIndex{2}}, args...) where {N} [deprecated]
│   │   │   │   reduced_densitymatrix(inds::NTuple{N,Tuple{Int,Int}}, args...) where {N} [deprecated]
│   │   │   │   Select an optimized density-matrix contraction when available, otherwise construct a normalized density matrix with _contract_densitymatrix.
│   │   │   │
│   │   │   ├── reduced_densitymatrix1x1(ind, ket, bra, env)
│   │   │   │   Construct the reduced density matrix of `|ket⟩⟨bra|` on the single site `ind` using an optimized contraction for the environment `env`; the CTMRG and BP implementations live in contractions/local_patch/densitymatrix/.
│   │   │   │
│   │   │   ├── reduced_densitymatrix2x1(ind, ket, bra, env)
│   │   │   │   Construct the reduced density matrix of `|ket⟩⟨bra|` on the vertical pair of sites starting at `ind` using an optimized contraction for the environment `env`; the CTMRG and BP implementations live in contractions/local_patch/densitymatrix/.
│   │   │   │
│   │   │   └── reduced_densitymatrix1x2(ind, ket, bra, env)
│   │   │       Construct the reduced density matrix of `|ket⟩⟨bra|` on the horizontal pair of sites starting at `ind` using an optimized contraction for the environment `env`; the CTMRG and BP implementations live in contractions/local_patch/densitymatrix/.
│   │   │
│   │   ├── patch_contractions.jl
│   │   │   ├── contract_local_operator(inds::Vector{CartesianIndex{2}}, O, ket::InfinitePEPS, bra::InfinitePEPS, env)
│   │   │   │   contract_local_operator(inds::Vector{CartesianIndex{2}}, O, state::InfinitePEPO, env)
│   │   │   │   contract_local_operator(inds::Vector{CartesianIndex{2}}, O, ket::InfinitePEPO, bra::InfinitePEPO, env)
│   │   │   │   contract_local_operator(inds::Vector{Tuple{Int,Int}}, O, args...)
│   │   │   │   contract_local_operator(inds::NTuple, args...) [deprecated]
│   │   │   │   Validate the inputs and invoke the generated direct contraction for an operator insertion.
│   │   │   │
│   │   │   └── contract_local_norm(inds::Vector{CartesianIndex{2}}, ket::InfinitePEPS, bra::InfinitePEPS, env)
│   │   │       contract_local_norm(inds::Vector{CartesianIndex{2}}, state::InfinitePEPO, env)
│   │   │       contract_local_norm(inds::Vector{CartesianIndex{2}}, ket::InfinitePEPO, bra::InfinitePEPO, env)
│   │   │       contract_local_norm(inds::Vector{Tuple{Int,Int}}, args...)
│   │   │       contract_local_norm(inds::NTuple, ket::InfinitePEPS, bra::InfinitePEPS, env) [deprecated]
│   │   │       Validate the inputs and invoke the generated direct contraction used to normalize an operator insertion.
│   │   │
│   │   ├── network_value.jl
│   │   │   ├── network_value(network::InfiniteSquareNetwork, env::CTMRGEnv)
│   │   │   │   network_value(state, env::CTMRGEnv)
│   │   │   │   Evaluate the normalized value per unit cell of an infinite square network.
│   │   │   │
│   │   │   ├── LinearAlgebra.norm(peps::InfinitePEPS, env::CTMRGEnv)
│   │   │   │   Evaluate the CTMRG norm of an infinite PEPS.
│   │   │   │
│   │   │   └── contract_local_tensor(inds::Tuple{Int,Int}, O::PFTensor, env::CTMRGEnv{C,<:CTMRG_PF_EdgeTensor}) where {C}
│   │   │       contract_local_tensor(ind::Tuple{Int,Int,Int}, O::PEPOTensor, network::InfiniteSquareNetwork{<:PEPOSandwich}, env::CTMRGEnv)
│   │   │       contract_local_tensor(inds::CartesianIndex, O::AbstractTensorMap, env::CTMRGEnv)
│   │   │       Contract a local tensor inserted into a partition function or into the PEPO of a network, using the surrounding CTMRG environment.
│   │   │
│   │   ├── correlators.jl
│   │   │   ├── const CoordCollection{N} = Union{AbstractVector{CartesianIndex{N}}, CartesianIndices{N}}
│   │   │   │   Accept either a vector of coordinates or a CartesianIndices range as correlator targets.
│   │   │   │
│   │   │   ├── MPSKit.correlator(bra::S, O, i::CartesianIndex{2}, js::CoordCollection{2}, ket::S, env::CTMRGEnv) where {S<:InfiniteState}
│   │   │   │   MPSKit.correlator(bra::S, O, i::CartesianIndex{2}, j::CartesianIndex{2}, ket::S, env::CTMRGEnv) where {S<:InfiniteState}
│   │   │   │   MPSKit.correlator(state::InfinitePEPS, O, i::CartesianIndex{2}, j, env::CTMRGEnv)
│   │   │   │   Build a braket correlator context and delegate to the generic correlator driver.
│   │   │   │
│   │   │   └── MPSKit.correlator(ρ::InfinitePEPO, O, i::CartesianIndex{2}, js::CoordCollection{2}, env::CTMRGEnv)
│   │   │       MPSKit.correlator(ρ::InfinitePEPO, O, i::CartesianIndex{2}, j::CartesianIndex{2}, env::CTMRGEnv)
│   │   │       Build a trace correlator context for a single-layer PEPO and delegate to the same driver.
│   │   │
│   │   └── correlator_adapters.jl
│   │       ├── struct PEPSCorrelator / PEPOPurifiedCorrelator / PEPOTraceCorrelator
│   │       │   const BraketCorrelator = Union{PEPSCorrelator, PEPOPurifiedCorrelator}
│   │       │   _braket_correlator(bra, ket, env::CTMRGEnv)
│   │       │   Bundle the bra, ket and environment into the context the driver dispatches on.
│   │       │
│   │       ├── edge_transfermatrix(row::Int, col::Int, context)
│   │       │   _correlator_scalartype(context, O::FiniteMPO)
│   │       │   _correlator_unitcell(context)
│   │       │   Base.rotl90(context)
│   │       │   Expose the per-context quantities the generic driver needs, including the rotation used for vertical correlators.
│   │       │
│   │       ├── _correlator(context, O, i::CartesianIndex{2}, js::AbstractVector{CartesianIndex{2}})
│   │       │   _correlator_horizontal(context, operator, i, js)
│   │       │   _correlator_vertical(context, operator, i, js)
│   │       │   _check_horizontal_correlator_sites(i, js)
│   │       │   Route a target set to the horizontal or vertical sweep and validate that it forms a single row or column.
│   │       │
│   │       └── _correlator_horizontal_right!(G, targets, context, O::FiniteMPO, i::CartesianIndex{2})
│   │           _correlator_horizontal_left!(G, targets, context, O::FiniteMPO, i::CartesianIndex{2})
│   │           Walk outward from the reference site to each target, reusing the partial contraction as the separation grows.
│   │
│   ├── contractions/
│   │   ├── absorb.jl
│   │   │   ├── absorb_left(A::AbstractTensorMap{<:Any,S}, C::AbstractTensorMap{<:Any,S,1,1}) where {S}
│   │   │   │   absorb_left(P::AbstractTensorMap{<:Any,S,1,N}, C::AbstractTensorMap{<:Any,S,1,1}) where {S,N}
│   │   │   │   Contract a matrix into the leftmost leg of a tensor.
│   │   │   │
│   │   │   ├── absorb_right(A::AbstractTensorMap{<:Any,S}, C::AbstractTensorMap{<:Any,S,1,1}) where {S}
│   │   │   │   absorb_right(E::AbstractTensorMap{<:Any,S,N,1}, C::AbstractTensorMap{<:Any,S,1,1}) where {S,N}
│   │   │   │   Contract a matrix into the rightmost leg of a tensor.
│   │   │   │
│   │   │   └── absorb_left_right(T::AbstractTensorMap{<:Any,S}, CL::AbstractTensorMap{<:Any,S,1,1}, CR::AbstractTensorMap{<:Any,S,1,1}) where {S}
│   │   │       Contract matrices into both outer legs of a tensor.
│   │   │
│   │   ├── bp_messages.jl
│   │   │   ├── const PEPSMessage = AbstractTensorMap{<:Any,<:Any,1,1}
│   │   │   │   Name the one-in one-out tensor carrying a BP message along a bond.
│   │   │   │
│   │   │   ├── absorb_north_message(A::PEPSTensor, M::PEPSMessage)
│   │   │   │   Absorb a north BP message into a PEPS tensor.
│   │   │   │
│   │   │   ├── absorb_east_message(A::PEPSTensor, M::PEPSMessage)
│   │   │   │   Absorb an east BP message into a PEPS tensor.
│   │   │   │
│   │   │   ├── absorb_south_message(A::PEPSTensor, M::PEPSMessage)
│   │   │   │   Absorb a south BP message into a PEPS tensor.
│   │   │   │
│   │   │   ├── absorb_west_message(A::PEPSTensor, M::PEPSMessage)
│   │   │   │   Absorb a west BP message into a PEPS tensor.
│   │   │   │
│   │   │   ├── contract_north_message(A::PEPSSandwich, M_west::PEPSMessage, M_north::PEPSMessage, M_east::PEPSMessage)
│   │   │   │   Update the BP message leaving a site toward the north.
│   │   │   │
│   │   │   ├── contract_east_message(A::PEPSSandwich, M_north::PEPSMessage, M_east::PEPSMessage, M_south::PEPSMessage)
│   │   │   │   Update the BP message leaving a site toward the east.
│   │   │   │
│   │   │   ├── contract_south_message(A::PEPSSandwich, M_east::PEPSMessage, M_south::PEPSMessage, M_west::PEPSMessage)
│   │   │   │   Update the BP message leaving a site toward the south.
│   │   │   │
│   │   │   └── contract_west_message(A::PEPSSandwich, M_south::PEPSMessage, M_west::PEPSMessage, M_north::PEPSMessage)
│   │   │       Update the BP message leaving a site toward the west.
│   │   │
│   │   ├── local_patch/
│   │   │   ├── expr_utils.jl
│   │   │   │   ├── tensorlabel(args...)
│   │   │   │   │   Join label components into a tensor-contraction symbol.
│   │   │   │   │
│   │   │   │   ├── envlabel(args...)
│   │   │   │   │   Construct an environment-leg label.
│   │   │   │   │
│   │   │   │   ├── virtuallabel(args...)
│   │   │   │   │   Construct a state virtual-leg label.
│   │   │   │   │
│   │   │   │   ├── physicallabel(args...)
│   │   │   │   │   Construct a physical-leg label.
│   │   │   │   │
│   │   │   │   ├── _open_slot(open, r, c)
│   │   │   │   │   Return the operator slot assigned to a patch site.
│   │   │   │   │
│   │   │   │   ├── _bulk_virtuallabels(i, j, layer, gridsize)
│   │   │   │   │   Return the north, east, south, and west labels of one bulk tensor.
│   │   │   │   │
│   │   │   │   ├── _patch_inds(inds::Type)
│   │   │   │   │   _patch_inds(inds::Tuple{Vararg{Val}})
│   │   │   │   │   Recover and validate patch coordinates encoded in Val parameters.
│   │   │   │   │
│   │   │   │   ├── _patch_ranges(sites)
│   │   │   │   │   Return the row and column ranges spanned by patch sites.
│   │   │   │   │
│   │   │   │   ├── _patch_gridsize(sites)
│   │   │   │   │   Return the rectangular dimensions spanned by patch sites.
│   │   │   │   │
│   │   │   │   └── _patch_shape_string(sites)
│   │   │   │       Format the rectangular patch dimensions for diagnostics.
│   │   │   │
│   │   │   ├── network_expr.jl
│   │   │   │   ├── _tensor_expr(prod, lhs = nothing)
│   │   │   │   │   Assemble indexed tensor expressions into an optimized `@tensor` contraction.
│   │   │   │   │
│   │   │   │   ├── boundary_contraction_expr(env::Type, rowrange, colrange)
│   │   │   │   │   boundary_contraction_expr(::Type{<:CTMRGEnv{C,T}}, rowrange, colrange) where {C,T}
│   │   │   │   │   Generate indexed CTMRG boundary tensor expressions for an `@tensor` contraction, or report that an environment implementation is unavailable.
│   │   │   │   │
│   │   │   │   ├── bulk_contraction_expr(state::Type, rowrange, colrange, open)
│   │   │   │   │   bulk_contraction_expr(::Type{<:Tuple{InfinitePEPS,InfinitePEPS}}, rowrange, colrange, open)
│   │   │   │   │   bulk_contraction_expr(::Type{<:InfinitePEPO}, rowrange, colrange, open)
│   │   │   │   │   bulk_contraction_expr(::Type{<:Tuple{InfinitePEPO,InfinitePEPO}}, rowrange, colrange, open)
│   │   │   │   │   bulk_contraction_expr(state::Type{<:Tuple{Vararg{InfinitePEPO}}}, rowrange, colrange, open)
│   │   │   │   │   Generate indexed PEPS or PEPO tensor expressions for the patch interior, or report that a state implementation is unavailable.
│   │   │   │   │
│   │   │   │   └── operator_contraction_expr(operator::Type, nsites)
│   │   │   │       operator_contraction_expr(::Type{<:AbstractTensorMap}, nsites)
│   │   │   │       Generate an indexed dense-operator expression for an `@tensor` contraction, or report that an operator implementation is unavailable.
│   │   │   │
│   │   │   ├── patch_contractions.jl
│   │   │   │   ├── _contract_local_operator(inds::NTuple{N,Val}, operator, state, env) where {N}
│   │   │   │   │   Assemble and execute the unnormalized operator-inserted patch contraction for direct evaluation.
│   │   │   │   │
│   │   │   │   └── _contract_local_norm(inds::NTuple{N,Val}, state, env) where {N}
│   │   │   │       Assemble and execute the corresponding patch-normalization contraction with closed physical legs.
│   │   │   │
│   │   │   └── densitymatrix/
│   │   │       ├── generic.jl
│   │   │       │   └── _contract_densitymatrix(inds::NTuple{N,Val}, state, env) where {N}
│   │   │       │       Assemble and execute a generated patch contraction with open physical legs.
│   │   │       │
│   │   │       ├── ctmrg.jl
│   │   │       │   ├── reduced_densitymatrix1x1(ind::CartesianIndex{2}, ket::InfinitePEPS, bra::InfinitePEPS, env::CTMRGEnv)
│   │   │       │   │   Contract an optimized one-site PEPS density matrix using a CTMRG environment.
│   │   │       │   │
│   │   │       │   ├── reduced_densitymatrix2x1(ind::CartesianIndex, ket::InfinitePEPS, bra::InfinitePEPS, env::CTMRGEnv)
│   │   │       │   │   Contract an optimized vertical two-site PEPS density matrix using a CTMRG environment.
│   │   │       │   │
│   │   │       │   └── reduced_densitymatrix1x2(ind::CartesianIndex, ket::InfinitePEPS, bra::InfinitePEPS, env::CTMRGEnv)
│   │   │       │       Contract an optimized horizontal two-site PEPS density matrix using a CTMRG environment.
│   │   │       │
│   │   │       └── bp.jl
│   │   │           ├── _contract_densitymatrix(inds::NTuple{N,Val}, state, env::BPEnv) where {N}
│   │   │           │   Reject patch geometries that cannot be represented by local BP messages.
│   │   │           │
│   │   │           ├── reduced_densitymatrix1x1(ind::CartesianIndex{2}, ket::InfinitePEPS, bra::InfinitePEPS, env::BPEnv)
│   │   │           │   Contract a one-site PEPS density matrix using BP messages.
│   │   │           │
│   │   │           ├── reduced_densitymatrix2x1(coord::CartesianIndex{2}, ket::InfinitePEPS, bra::InfinitePEPS, env::BPEnv)
│   │   │           │   Contract a vertical two-site PEPS density matrix using BP messages.
│   │   │           │
│   │   │           └── reduced_densitymatrix1x2(coord::CartesianIndex{2}, ket::InfinitePEPS, bra::InfinitePEPS, env::BPEnv)
│   │   │               Contract a horizontal two-site PEPS density matrix using BP messages.
│   │   │
│   │   └── ctmrg/
│   │       └── network_value.jl
│   │           ├── _contract_site(ind::Tuple{Int,Int}, network, env::CTMRGEnv)
│   │           │   _contract_site(C_northwest, C_northeast, C_southeast, C_southwest, E_north::CTMRG_PEPS_EdgeTensor, E_east::CTMRG_PEPS_EdgeTensor, E_south::CTMRG_PEPS_EdgeTensor, E_west::CTMRG_PEPS_EdgeTensor, O::PEPSSandwich)
│   │           │   _contract_site(C_northwest, C_northeast, C_southeast, C_southwest, E_north::CTMRG_PF_EdgeTensor, E_east::CTMRG_PF_EdgeTensor, E_south::CTMRG_PF_EdgeTensor, E_west::CTMRG_PF_EdgeTensor, O::PFTensor)
│   │           │   _contract_site(C_northwest, C_northeast, C_southeast, C_southwest, E_north::TE, E_east::TE, E_south::TE, E_west::TE, O::PEPOSandwich{H}) where {TE<:CTMRGEdgeTensor,H}
│   │           │   Contract one network site with its surrounding CTMRG corners and edges.
│   │           │
│   │           ├── _contract_corners(ind::Tuple{Int,Int}, env::CTMRGEnv)
│   │           │   _contract_corners(C_northwest::CTMRGCornerTensor, C_northeast::CTMRGCornerTensor, C_southeast::CTMRGCornerTensor, C_southwest::CTMRGCornerTensor)
│   │           │   Contract four CTMRG corners for network-value normalization.
│   │           │
│   │           ├── _contract_vertical_edges(ind::Tuple{Int,Int}, env::CTMRGEnv)
│   │           │   _contract_vertical_edges(C_northwest::CTMRGCornerTensor, C_northeast::CTMRGCornerTensor, C_southeast::CTMRGCornerTensor, C_southwest::CTMRGCornerTensor, E_east::CTMRGEdgeTensor{T,S,N}, E_west::CTMRGEdgeTensor{T,S,N}) where {T,S,N}
│   │           │   Contract two vertical edges and four corners for network-value normalization.
│   │           │
│   │           └── _contract_horizontal_edges(ind::Tuple{Int,Int}, env::CTMRGEnv)
│   │               _contract_horizontal_edges(C_northwest::CTMRGCornerTensor, C_northeast::CTMRGCornerTensor, C_southeast::CTMRGCornerTensor, C_southwest::CTMRGCornerTensor, E_north::CTMRGEdgeTensor{T,S,N}, E_south::CTMRGEdgeTensor{T,S,N}) where {T,S,N}
│   │               Contract two horizontal edges and four corners for network-value normalization.
│   │
│   └── optimization/
│       └── peps_optimization.jl
│           └── cost_function(peps::InfinitePEPS, env, O::LocalOperator)
│               Return the real part of an expectation value as the PEPS optimization cost.
│
└── utility/
    └── tensor_traces.jl
        ├── _str(::Bosonic, t::AbstractTensorMap)
        │   _str(::Fermionic, t::AbstractTensorMap{<:Any,<:Any,N,N}) where {N}
        │   Dispatch the trace calculation on bosonic or fermionic braiding style.
        │
        ├── str(t::AbstractTensorMap)
        │   Compute the ordinary trace for bosonic tensors and the supertrace for fermionic tensors.
        │
        └── trmul(H::AbstractTensorMap{<:Any,S,N,N}, ρ::AbstractTensorMap{<:Any,S,N,N}) where {S,N}
            Contract an operator with a density matrix without materializing their product.
```
</details>

The only notable differences are that:
- I added an extra `expectation_value` folder directly under `src/algorithms` to collect everything related to expectation values. I also move the correlator code here, since it seemed related enough to put there. I can always revert if we don't want that.
- I split some more "public" methods from their implementations under `src/algorithms/contractions` (e.g. `contract_local_norm` and `contract_local_operator`, `network_value`, ...). This felt more appropriate, since `src/contractions` was originally intended to contain the lowest level contraction methods.

Happy to reorganize further, or more thoroughly if we no longer like the use of the `src/algorithms/contractions` approach.

src/utility/tensor_traces.jl
```diff
@@ -0,0 +1,27 @@
...
+    tex = tensorexpr(:t, ntuple(identity, N), ntuple(identity, N))
+    return macroexpand(@__MODULE__, :(@tensor $tex))
+end
+
+"""
+    trmul(H, ρ)
+
+Compute `tr(H * ρ)` without forming `H * ρ`.
+"""
+@generated function trmul(
```
Yue-Zhengyuan, line 21: This is actually computing `str(H * ρ)`. @lkdvos we really need to understand why supertrace appears.

src/PEPSKit.jl
```diff
@@ -95,15 +96,21 @@ include("algorithms/contractions/ctmrg/halfinf_env.jl")
...
 include("algorithms/contractions/ctmrg/characteristic_equations.jl")
 
+include("algorithms/contractions/absorb.jl")
 include("algorithms/contractions/absorb_weight.jl")
 include("algorithms/contractions/transfer.jl")
-include("algorithms/contractions/localoperator.jl")
 include("algorithms/contractions/vumps_contractions.jl")
-include("algorithms/contractions/bp_contractions.jl")
+include("algorithms/contractions/bp_messages.jl")
+include("algorithms/contractions/local_patch/expr_utils.jl")
```
Yue-Zhengyuan, line 108: Including this (which contains `envlabel` etc.) after `network_value` will cause error if the package is directly included by source. Things are fine if the package is precompiled first, but we may better reorder the includes.

src/algorithms/contractions/local_patch/patch_contractions.jl
```diff
@@ -0,0 +1,60 @@
...
+"""
+@generated function _contract_local_operator(
+        inds::NTuple{N, Val}, operator, state, env
+    ) where {N}
+    sites = _patch_inds(inds)
+    rowrange, colrange = _patch_ranges(sites)
+
+    multiplication_ex = Expr(
+        :call, :*,
+        boundary_contraction_expr(env, rowrange, colrange)...,
```
Yue-Zhengyuan, line 25: Codex discovers that user extensions of `boundary_contraction_expr` after loading PEPSKit will not work, as demonstrated by the following example code.

<details>

```julia
using PEPSKit, TensorKit

struct WrappedEnv{E}
    inner::E
end

"""Reuse the CTMRG boundary expression for a wrapped environment."""
PEPSKit.boundary_contraction_expr(::Type{WrappedEnv{E}}, rows, cols) where {E} =
    PEPSKit.boundary_contraction_expr(E, rows, cols)

"""Forward corner access to the wrapped environment."""
PEPSKit.corner(env::WrappedEnv, args...) = PEPSKit.corner(env.inner, args...)

"""Forward edge access to the wrapped environment."""
PEPSKit.edge(env::WrappedEnv, args...) = PEPSKit.edge(env.inner, args...)

peps = InfinitePEPS(ℂ^2, ℂ^1; unitcell = (1, 1))
env = WrappedEnv(CTMRGEnv(peps, ℂ^1))

# Ordinary dispatch sees the new hook and returns eight boundary factors.
@show length(PEPSKit.boundary_contraction_expr(typeof(env), 1:1, 1:1))

# The generated function still sees the original throwing fallback.
PEPSKit.contract_local_norm([CartesianIndex(1, 1)], peps, peps, env)
```

</details>

src/algorithms/contractions/absorb.jl
```diff
@@ -0,0 +1,65 @@
...
+    ) where {S, N}
+    return twistnondual(C, 2) * P
+end
+
+"""
+    absorb_right(
+        A::AbstractTensorMap{<:Any, S}, C::AbstractTensorMap{<:Any, S, 1, 1}
+    ) where {S}
+
+Absorb a matrix `C` into the right of a tensor map `A` by contracting the last leg of `A`
```
Yue-Zhengyuan, line 31: Well, it is not always the last leg of `A`, but `domainind(A)[1]`, unless `A` only has one leg on the domain side.

src/algorithms/expectation_value/reduced_densitymatrix.jl
```diff
@@ -0,0 +1,109 @@
...
+Alternatively, construct the reduced density matrix `ρ` of a mixed state specified by the 
+density matrix PEPO `state`. The reduced density matrix is contracted over the virtual
+indices surrounding the open physical indices at sites `inds` using the environment `env`,
+and is normalized such that `str(ρ) = 1`.
+
+See also [`str`](@ref).
+""" reduced_densitymatrix
+
+# PEPS case has fast-path specializations
+function reduced_densitymatrix(
```
Yue-Zhengyuan, line 20: The logic basically requires that when e.g. only 1-site is involved, all `env` types should have `reduced_densitymatrix1x1` defined, even if they have the general `_contract_densitymatrix`. The fallback mechanism can be more clever.

src/algorithms/expectation_value/patch_contractions.jl
```diff
@@ -0,0 +1,87 @@
+# Direct local patch contractions
+# -------------------------------
+
+"""
+    contract_local_operator(inds, O, ket::InfinitePEPS, bra::InfinitePEPS = ket, env)
```
Yue-Zhengyuan, line 5: Still, no bra = ket version actually exists.

Yue-Zhengyuan (commented):
Some code extensibility issues involving custom environment types discovered by Codex. Again, we need to explain why supertrace instead of trace occurs in expectation values.

I'm in favor of putting the low-level contractions beside the algorithm that uses it, instead of collecting them all together. But this can be deferred to a follow up.

src/algorithms/contractions/local_patch/patch_contractions.jl
leburgel: I never really thought of these as user-extendable entry points, my only goal was to make it easier for us to add support for new internal environments and operators. It would be nice if this also allowed users to extend support in an easy way, so this is a very good point!

I did a bit of digging, and unfortunately this doesn't seem straightforward at all. A `@generated` function's generator can only call methods that existed when the generated function was defined. The patch contractions are `@generated` and their generators dispatch on `boundary_contraction_expr` / `bulk_contraction_expr` / `operator_contraction_expr`, so any method a user adds after using PEPSKit is invisible to them — they silently keep resolving the throwing fallback, even though ordinary dispatch finds the new method fine.

A new hook method corresponding to a genuinely new environment can only be seen by a generator defined *after* the new hook method definitions. The only way out of this is to `@eval` a fresh generated method after the hooks exist.

We could provide a convenience macro that does this, which would look something like

```julia
function PEPSKit.boundary_contraction_expr(::Type{MyNewEnvType}, rows, cols)
    # implement contraction expression for custom environment types
end
PEPSKit.@patch_contraction_methods env = MyNewEnvType # refresh the generators so they actually find these new boundary contraction expressions
```

This would require some changes,

<details>
<summary>Required changes</summary>

```julia
# factor the generator bodies into plain functions
function _patch_operator_body(inds::Type, operator::Type, state::Type, env::Type)
    sites = _patch_inds(inds)
    rowrange, colrange = _patch_ranges(sites)
    multiplication_ex = Expr(
        :call, :*,
        boundary_contraction_expr(env, rowrange, colrange)...,
        bulk_contraction_expr(state, rowrange, colrange, sites)...,
        operator_contraction_expr(operator, length(sites))...,
    )
    return macroexpand(@__MODULE__, _tensor_expr(multiplication_ex))
end

function _patch_norm_body(inds::Type, state::Type, env::Type)
    sites = _patch_inds(inds)
    rowrange, colrange = _patch_ranges(sites)
    multiplication_ex = Expr(
        :call, :*,
        boundary_contraction_expr(env, rowrange, colrange)...,
        bulk_contraction_expr(state, rowrange, colrange, nothing)...,
    )
    return macroexpand(@__MODULE__, _tensor_expr(multiplication_ex))
end

# the three generated functions become one-liners over these bodies
@generated function _contract_local_operator(
        inds::NTuple{N, Val}, operator, state, env
    ) where {N}
    return _patch_operator_body(inds, operator, state, env)
end

@generated function _contract_local_norm(inds::NTuple{N, Val}, state, env) where {N}
    return _patch_norm_body(inds, state, env)
end

# use this to construct a "refresh-macro"
"""
    @patch_contraction_methods [env = EnvType] [state = StateType] [operator = OperatorType]

Regenerate the generated patch contractions in the calling module, narrowed to the given
argument types. Needed whenever a `boundary_contraction_expr`, `bulk_contraction_expr` or
`operator_contraction_expr` method is defined *after* PEPSKit is loaded: the contractions are
`@generated`, and a generator may only call methods that already existed when it was defined,
so PEPSKit's own definitions can never see such a method. Regenerating them afterwards gives
them a generator young enough to see it.
"""
macro patch_contraction_methods(args...)
    env = state = operator = :Any
    for arg in args
        if Meta.isexpr(arg, :(=), 2) && arg.args[1] in (:env, :state, :operator)
            key, value = arg.args
            key === :env && (env = value)
            key === :state && (state = value)
            key === :operator && (operator = value)
        elseif length(args) == 1
            env = arg   # a single bare type argument specifies the environment
        else
            throw(ArgumentError("Invalid argument to `@patch_contraction_methods`: $arg."))
        end
    end

    M = @__MODULE__
    return esc(
        quote
            Base.@generated function $M._contract_local_operator(
                    inds::NTuple{N, Val}, operator::$operator, state::$state, env::$env
                ) where {N}
                return $M._patch_operator_body(inds, operator, state, env)
            end
            Base.@generated function $M._contract_local_norm(
                    inds::NTuple{N, Val}, state::$state, env::$env
                ) where {N}
                return $M._patch_norm_body(inds, state, env)
            end
            Base.@generated function $M._contract_densitymatrix(
                    inds::NTuple{N, Val}, state::$state, env::$env
                ) where {N}
                return $M._patch_densitymatrix_body(inds, state, env)
            end
            nothing
        end
    )
end
```

</details>

This goes well beyond what I originally intended, and hinges very strongly on brand new internals that might have to be changed if we bump into issues when actually adding support for new environments and operators.

Do you think adding something like this would be worth it?

Yue-Zhengyuan: If these are never intended to be extended outside the package, then current design is fine.

src/utility/tensor_traces.jl
leburgel: I updated the docstring, and added an explicit reference to `str`.

src/PEPSKit.jl
leburgel: I moved the most basic label generators up into a dedicated `utility` file.

src/algorithms/expectation_value/patch_contractions.jl
leburgel: I removed the non-existing default from the docstring.

src/algorithms/expectation_value/reduced_densitymatrix.jl
leburgel: That's a great catch, big oversight on my part. Should be fixed now.

src/algorithms/contractions/absorb.jl
leburgel: I corrected the docstring to be correct in the general case.

src/algorithms/expectation_value.jl
leburgel: I just added an explicit reference to `str` to be more precise, since this is equivalent to calling `str` on the reduced density matrix of the same patch.

src/algorithms/expectation_value/patch_contractions.jl
```diff
@@ -0,0 +1,89 @@
...
+
+"""
+    contract_local_norm(inds, ket::InfinitePEPS, bra::InfinitePEPS, env)
+    contract_local_norm(inds, ket::InfinitePEPO, bra::InfinitePEPO, env)
+    contract_local_norm(inds, state::InfinitePEPO, env)
+
+Contract a local norm corresponding to the overlap `ket` and `bra` states, computing a patch
+of `⟨bra|ket⟩`, where `ket` and `bra` correspond to either a PEPS or a PEPO representing a
+PEPS with ancillary legs.
+Alternatively, contract a local norm patch of a density matrix PEPO `state`, computing a patch of `tr(state)`.
```
Yue-Zhengyuan, line 56: I changed this back to ordinary trace. The RDM is using `str` only because we manually added twist on the 2nd codomain index of each iPEPO tensor.

Yue-Zhengyuan (approved):
Ready for merge after #429 and updating to the new test format.

PR #425, opened 2026-09-02
[WIP] Add support for local expectation values of tensor product and MPO terms

leburgel:
Built on top of #423.

Adds support for alternative operator forms for situations that are very badly suited to our current dense-tensor-operator approach, while still being compatible with AD to use them in variational optimizations. For now this is mostly a demonstration of how #423 allows to easily add support for the evaluation of different operator types by only overloading `operator_contraction_expr`. The concrete design will be revisited later.


Yue-Zhengyuan:
I suppose this is *exact* contraction with CTMRG/BPEnv ?

leburgel:
> I suppose this is _exact_ contraction with CTMRG/BPEnv ?

Yes, precisely that. The idea was to add support for alternative operator forms for situations that are very badly suited to our current dense-tensor-operator approach, while still being compatible with AD to use them in variational optimizations. In particular, the motivating cases are multi-site interactions that factor into a plain tensor product across sites (e.g. plaquette terms in gauge theories), and short-range two-site interactions for systems with a large physical dimension. Both of these are extremely expensive for our current approach, while simply storing the tensor product factors separately or storing the operator as a 2-site MPO of low bond dimension greatly reduces the cost.

This is still very much a draft, but I already filed the PR to demonstrate how #423 allows to easily add support for exact evaluation of different operator types. In particular, the `MPOTerm` here is a straight duplicate of the `MPOObservable` introduced in #415. This PR can just stay in draft status until the design that is merged, and then I'll adopt the resulting design here.

src/operators/localoperator.jl
Yue-Zhengyuan, line 352: This will bypass the check in `add_terms!` and add two MPOs incorrectly. The issue can be detected by

```julia
Hₓ = LocalOperator(lattice, sites => gate_to_mpo(X ⊗ X))
Hᶻ = LocalOperator(lattice, sites => gate_to_mpo(Z ⊗ Z))
H = Hₓ + Hᶻ
@test dense_operator(only(values(H.terms))) ≈ X ⊗ X + Z ⊗ Z
```

where `dense_operator` is a function that converts the MPO back to a dense TensorMap.

PR #427, opened 2026-09-04
Collect twistdual utility functions

Yue-Zhengyuan:
I think there is now a large enough family of twistdual-like functions, and it's time to collect them in one place.

The non-mutating `twistdual` is no longer implemented with `twist! + copy`, but directly using `twist`.

### DIFF
--- a/src/utility/twistdual.jl
+++ b/src/utility/twistdual.jl
@@ -1,0 +1,78 @@
+"""
+$(SIGNATURES)
+
+Twist the i-th leg of a tensor `t` in place if it represents a dual space.
+"""
+function twistdual!(t::AbstractTensorMap, i::Int)
+    isdual(space(t, i)) || return t
+    return twist!(t, i)
+end
+function twistdual!(t::AbstractTensorMap, is)
+    is′ = filter(i -> isdual(space(t, i)), is)
+    return twist!(t, is′)
+end
+
+"""
+$(SIGNATURES)
+
+Twist the i-th leg of a tensor `t` if it represents a dual space.
+"""
+function twistdual(t::AbstractTensorMap, i::Int)
+    isdual(space(t, i)) || return t
+    return twist(t, i)
+end
+function twistdual(t::AbstractTensorMap, is)
+    is′ = filter(i -> isdual(space(t, i)), is)
+    return twist(t, is′)
+end
+function ChainRulesCore.rrule(
+        config::RuleConfig, ::typeof(twistdual), t::AbstractTensorMap, i
+    )
+    tout, twistdual_pullback = rrule_via_ad(config, twistdual, t, i)
+    return tout, twistdual_pullback
+end
+
+"""
+$(SIGNATURES)
+
+Twist the i-th leg of a tensor `t` in place if it represents a non-dual space.
+"""
+function twistnondual!(t::AbstractTensorMap, i::Int)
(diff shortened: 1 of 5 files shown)


src/utility/twistdual.jl
```diff
@@ -0,0 +1,66 @@
...
+    is′ = filter(i -> !isdual(space(t, i)), is)
+    return twist(t, is′)
+end
+
+"""
+$(SIGNATURES)
+
+Apply a twist to domain or codomain indices that correspond to dual spaces.
+"""
+function _linearmap_twist!(t::AbstractTensorMap)
```
leburgel: I think `_twist_linearmap` sounds slightly more natural.

src/utility/twistdual.jl
```diff
@@ -0,0 +1,66 @@
...
+Twist the i-th leg of a tensor `t` if it represents a dual space.
+"""
+function twistdual(t::AbstractTensorMap, i::Int)
+    isdual(space(t, i)) || return t
+    return twist(t, i)
+end
+function twistdual(t::AbstractTensorMap, is)
+    is′ = filter(i -> isdual(space(t, i)), is)
+    return twist(t, is′)
+end
```
leburgel: Since the implementations are really exactly the same up to the bang, it might be good to fold both into an `@eval` loop so the implementations really  share the same lines, in case we have to change things.


```suggestion
for (twistdual, twist, inplace) in zip((:twistdual!, :twistdual), (:twist!, :twist), (" in place ", " "))
    @eval begin
        """
        $(SIGNATURES)

        Twist the i-th leg of a tensor `t`$($(inplace))if it represents a dual space.
        """
        function $twistdual(t::AbstractTensorMap, i::Int)
            isdual(space(t, i)) || return t
            return $twist(t, i)
        end
        function $twistdual(t::AbstractTensorMap, is)
            is′ = filter(i -> isdual(space(t, i)), is)
            return $twist(t, is′)
        end
    end
end
```

src/utility/twistdual.jl
```diff
@@ -0,0 +1,66 @@
...
+Twist the i-th leg of a tensor `t` if it represents a non-dual space.
+"""
+function twistnondual(t::AbstractTensorMap, i::Int)
+    !isdual(space(t, i)) || return t
+    return twist(t, i)
+end
+function twistnondual(t::AbstractTensorMap, is)
+    is′ = filter(i -> !isdual(space(t, i)), is)
+    return twist(t, is′)
+end
```
leburgel: Same here, a truly shared implementation might be slightly more future-proof.
```suggestion
for (twistnondual, twist, inplace) in zip((:twistnondual!, :twistnondual), (:twist!, :twist), (" in place ", " "))
    @eval begin
        """
        $(SIGNATURES)

        Twist the i-th leg of a tensor `t`$($(inplace))if it represents a non-dual space.
        """
        function $twistnondual(t::AbstractTensorMap, i::Int)
            !isdual(space(t, i)) || return t
            return $twist(t, i)
        end
        function $twistnondual(t::AbstractTensorMap, is)
            is′ = filter(i -> !isdual(space(t, i)), is)
            return $twist(t, is′)
        end
    end
end
```

leburgel (approved):
Just some minor preference comments, but I think this is great. Especially happy to get rid of the AD hacks.

PR #428, opened 2026-09-05
Reorganize code for CTMRG projector algorithms

Yue-Zhengyuan:
This PR puts code for each projector algorithm into its own file, hopefully making the source tree cleaner and ready for addition of new projector algorithms (such as #263, #422).


leburgel:
Would it make sense to also split off the C4v projectors (`C4vEighProjector` and `C4vQRProjector`)?

src/algorithms/ctmrg/projectors/projectors.jl
```diff
@@ -0,0 +1,92 @@
...
+    # parse SVD forward & rrule algorithm
+
+    decomposition_algorithm = if alg in [:HalfInfiniteProjector, :FullInfiniteProjector]
+        _alg_or_nt(SVDAdjoint, decomposition_alg)
+    elseif alg in [:C4vEighProjector]
+        _alg_or_nt(EighAdjoint, decomposition_alg)
+    elseif alg in [:C4vQRProjector]
+        _alg_or_nt(QRAdjoint, decomposition_alg)
+    end # TODO: how do we solve this in a proper way?
+
```
leburgel, line 35: Not the point here but more a note for the future: it would be much cleaner to handle this via dispatch on the alg_type instead of a symbol-based switch. Same for the truncation handling.

I can do this in a follow-up.

src/algorithms/ctmrg/projectors/c4v_eigh.jl
leburgel, line 1: Why not add this to the same `projectors` folder, but name the file `c4veigh.jl`?

src/algorithms/ctmrg/projectors/c4v_qr.jl
leburgel, line 1: Why not add this to the same `projectors` folder, but name the file `c4vqr.jl`?

src/algorithms/ctmrg/projectors_c4v/qr.jl
```diff
@@ -0,0 +1,101 @@
...
+
+"""Contract two CTMRG edge tensors into a corner tensor."""
+@generated function contract_edges(
+        EL::CTMRGEdgeTensor{T, S, N}, ER::CTMRGEdgeTensor{T, S, N}
+    ) where {T, S, N}
+    C´_e = tensorexpr(:C´, -1, -2)
+    EL_e = tensorexpr(:EL, (-1, (2:N)...), 1)
+    ER_e = tensorexpr(:ER, 1:N, -2)
+    return macroexpand(@__MODULE__, :(return @tensor $C´_e := $EL_e * $ER_e))
+end
```
leburgel: It feels to me like these two belong more in the original `c4v.jl` file. Would that be okay, or do you prefer them here?

Yue-Zhengyuan: Moved back to `c4v.jl`, with some refactoring of `c4v_renormalize_corner` to make slightly upper level code look more symmetric on `eigh` and `qr` paths.

Yue-Zhengyuan:
The test errors are due to upstream Julia 1.13 / VectorInterface compatibility issue: https://github.com/QuantumKitHub/VectorInterface.jl/pull/54.

leburgel (approved):

PR #429, opened 2026-09-17
Move tests into testsuite

kshyatt:
The idea here is to end up with something similar to MatrixAlgebraKit, and allow us to support ✨ new array types ✨ without having to reduplicate a lot of the test logic -- but this can really be done separately, so I thought it would make sense to split it off into its own PR.


kshyatt:
The `ctmrg` failures are not related, it's a preexisting issue with VI

leburgel (approved):
I don't really have time to go through the whole thing right away, but this looks pretty good to me. Especially given it saves future copies.

I like that the method names shadow the file paths, this is something we should try to stick to, as it makes things quite deterministic. Adding tests takes a few extra steps now, but the workflow and instructions in `TestSuite` are clear, so I have no real objections.

kshyatt:
Thanks, @leburgel . I'll let this sit over the weekend so others can chime in if they have thoughts.

kshyatt:
Ah thanks @Yue-Zhengyuan ! This is exactly why I didn't insta merge this

leburgel:
Had a better look now and spotted a few more things.

Updates:
- Fixed a missing `AT` argument in the U1 contraction test — it aborted the entire "Specific U1 spaces" testset, silently dropping 3621 tests including the 4×4 unit cell case.
- Added `TensorKit.storagetype` for the five container types, mirroring the neighbouring `spacetype` definitions, plus a changelog entry — the PR's new assertions called it and hit 21 `MethodError`s.
- Moved `Random.seed!` from file top level into each test entry point — the seeds ran at include time rather than at draw time, so every declared seed was dead.
- Fixed a mangled comment reading "update `trunc` to `trunc`".
- Added the missing `($AT)` to one testset name.
- Removed a redundant bare `using TestExtras`.
- Collapsed a stray double blank line left behind by moving the seed out of the VUMPS test.

One remaining thing is that the `test/types/*.jl` files don't have a `is_buildkite` guard, so they now run on the GPU runners. I wasn't sure if this was intentional, so I wanted to check before changing anything.

kshyatt:
Nice, thanks for catching that.

For the types tests, there probably should be a buildkite guard, I didn't notice because they run so quickly. If you want to add it, go ahead, or I can do so later today.

kshyatt:
Cool, shall we hit the big merge button if CI passes, then?

leburgel:
Sounds good to me!

kshyatt:
OK the gradients and ctmrg failures are pre-existing, I'm going to merge this to unblock other things

PR #430, opened 2026-09-20
[WIP] Implicit gradient performance improvements

leburgel:
Performance improvements for the computation of implicit contraction gradients. Supporting benchmarks on the way.


src/algorithms/contractions/ctmrg/characteristic_equations.jl
```diff
@@ -95,15 +99,15 @@ function generate_symmetric_characteristic_equation(
         Cfp::CornerTensor,
         Efp::EdgeTensor, # unused
         Ufp::LeftProjector,
-        ULfp::LeftProjector,
+        ULfp::LeftProjector;
+        site = getindex,
```
pbrehmer: Maybe a different name would make it clearer that this is a getter function? Just `site` sounded like an index to me initially.
```suggestion
        getsite = getindex,
```

src/algorithms/contractions/ctmrg/characteristic_equations.jl
```diff
@@ -480,45 +513,70 @@ end
...
+function ``F(n, C, E, u, S, v)`` which characterizes the convergence of the CTMRG algorithm
+in terms of the characteristic equation ``F(n, C, E, u, S, v) = 0``. Here, ``n`` is the
+`InfiniteSquareNetwork` containing the state variable being optimized, and ``(C, E, u, S,
+v)`` represents a CTMRG contraction environment on a generic unit cell meaning that all
+tensors have a directional and unit cell index. ``C`` and ``E`` directly represent the
+corner and edge tensors, while ``u`` and ``v`` parametrize differentiable projectors ``U =
+U_{fp} + P_\\perp^U u`` and ``V = V_{fp} + v P_\\perp^V``, and ``S`` denotes the singular
+values of the decomposed environment. Here ``P_\\perp^U = 1 - U_{fp} U_{fp}^\\dagger`` and
+``P_\\perp^V = 1 - V_{fp}^\\dagger V_{fp}`` are projectors onto the null spaces of the
+fixed-point isometries. Thefore, while ``u`` and ``v`` formally live in the full projector
```
pbrehmer:
```suggestion
fixed-point isometries. Therefore, while ``u`` and ``v`` formally live in the full projector
```

pbrehmer (commented):
This is really neat and the benchmarks look great! I only have very minor comments. Given that there is a substantial speedup, I was wondering whether we should default to implicit differentiation now? I think for `C4vCTMRG` that would certainly make sense, for `SimultaneousCTMRG` it's maybe not as obvious? I guess it would be good to have a comparative benchmark of the sped-up implicit versus fixed-point differentiation; so in case that's just a quick Claude prompt away, maybe we could find out? :-)

src/algorithms/contractions/ctmrg/characteristic_equations.jl
leburgel: I picked `site` to try and be in line with the `corner` and `edge` getters we have now, but it does sound very much like an index variable. I'll change it to `getsite` throughout.

leburgel:
> Given that there is a substantial speedup, I was wondering whether we should default to implicit differentiation now?

I'd be surprised if the implicit approach could beat fixed-point differentiation with the full pullbacks using the un-truncated decompositions for the bond dimensions we're currently usually dealing with, but I'll try to set up some comparison to see how big the gap still is at this point.

pbrehmer:
> I'd be surprised if the implicit approach could beat fixed-point differentiation with the full pullbacks using the un-truncated decompositions for the bond dimensions we're currently usually dealing with, but I'll try to set up some comparison to see how big the gap still is at this point.

Of course you're right. I forgot that the default case is using untruncated decompositions that are hard to beat. Still it'd be interesting to see, thanks!

leburgel:
I ran into some unexpected issues while benchmarking so I don't have a full result, but so far the implicit approach consistently beats even the full pullback for the C4v case for all bond dimensions I've tried. So there we could switch the default already.

For the asymmetric case, the implicit approach is still about a factor 2 slower than using full pullbacks. For small bond dimensions, it's even slower than using the MatrixAlgebraKit.jl truncated pullbacks, and this only changes at larger bond dimensions. I'll keep working on getting a full set of benchmarks once I resolve my remaining issues, but I think we can keep a default-switch for a follow up. Also to make it more visible.

I'm quite happy with this by itself though, should be good to go as is as far as I'm concerned.

pbrehmer:
I guess the CTMRG failures on `v1.13` are unrelated to this PR, right? Also good to go for me in that case.

pbrehmer (approved):

